PR #1, opened 2019-08-31
Add CAIP-2 (Blockchain references)

ligi:
An initial draft. Currently only for Ethereum and Bitcoin. Perhaps @webmaster128 can chime in how to best do it for Cosmos and @soc1c for polkadot parachains. Input welcome!
Also wondering if IPFS could be a target. It is not a BlockChain - but e.g. for IN3 it gets assigned a chainId (0x7d0 https://github.com/slockit/in3#ipfs) which seems quite odd. Giving it a prefix in CAIP-2 could work around this a bit cc @CJentzsch @simon-jentzsch

### DIFF
--- a/CAIPs/caip-2.md
+++ b/CAIPs/caip-2.md
@@ -2,7 +2,7 @@
 caip: 2
 title: Blockchain references
 author: ligi <ligi@ligi.de>
-discussions-to: <URL>
+discussions-to: https://ethereum-magicians.org/t/caip-2-blockchain-references/3612
 status: Draft
 type: Standard
 created: 2019-08-31
--- a/CAIPs/caip-2.md
+++ b/CAIPs/caip-2.md
@@ -1,0 +1,38 @@
+---
+caip: 2
+title: Blockchain references
+author: ligi <ligi@ligi.de>
+discussions-to: <URL>
+status: Draft
+type: Standard
+created: 2019-08-31
+---
+
+## Simple Summary
+
+CAIP-2 defines a way to reference a blockchain. E.g. Ethereum mainnet, Görli, BTC, ..
+
+## Abstract
+
+Often you need to reference a blockchain. For example when you want to state where some asset or smart contract is located. In Ethereum the [EIP155] ChainId is used most of the time. But with an Ethereum chainId you cannot reference e.g. a Bitcoin or Cosmos chain.
+
+## Motivation
+
+The final trigger to create this CAIP (and the CAIP process itself) was a discussion around [EIP2256] at [Ethereum-Magicians](https://ethereum-magicians.org/t/eip-2256-add-wallet-getownedtokens-json-rpc-method/3600/14)
+
+## Specification
+
+The blockchain reference in CAIP-2 is a string. The string starts with a prefix. This prefix indicates what type of chain it is and how the blockchain reference is to be interptreted. For [EIP155] compatible chains the prefix is `eip155-` and is then followed by the chainId in decimal. This means `eip155-1` is the Ethereum mainnet and `eip155-5` is Görli. Bitcoin chains have the prefix `bitcoin-` and the prefix is followed by the genesis hash (like in [BIP122]) - this means `bitcoin-000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f` is the bitcoin mainnet. `bitcoin-000000000933ea01ad0ee984209779baaec3ced90fa3f408719526f8d77f4943` is the bitcoin testnet and `bitcoin-12a765e31ffd4059bada1e25190f6e98c99d9714d334efa41a195a7e7e04bfe2` is litecoin
+
+## Rationale
+
+Another approach would be to just use the genesis hash as in [BIP122] - but this would make it hard to determine what kind of chain it is. Also having the prefix allows to easily enhance current [EIP155] references by just adding the prefix.
+
+## links
+
+[EIP155]: https://eips.ethereum.org/EIPS/eip-155
+[EIP2256]: https://github.com/ethereum/EIPs/pull/2256
+[BIP122]: https://github.com/bitcoin/bips/blob/master/bip-0122.mediawiki
+
+## Copyright
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
--- a/CAIPs/caip-2.md
+++ b/CAIPs/caip-2.md
@@ -22,7 +22,17 @@ The final trigger to create this CAIP (and the CAIP process itself) was a discus
 
 ## Specification
 
-The blockchain reference in CAIP-2 is a string. The string starts with a prefix. This prefix indicates what type of chain it is and how the blockchain reference is to be interptreted. For [EIP155] compatible chains the prefix is `eip155-` and is then followed by the chainId in decimal. This means `eip155-1` is the Ethereum mainnet and `eip155-5` is Görli. Bitcoin chains have the prefix `bitcoin-` and the prefix is followed by the genesis hash (like in [BIP122]) - this means `bitcoin-000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f` is the bitcoin mainnet. `bitcoin-000000000933ea01ad0ee984209779baaec3ced90fa3f408719526f8d77f4943` is the bitcoin testnet and `bitcoin-12a765e31ffd4059bada1e25190f6e98c99d9714d334efa41a195a7e7e04bfe2` is litecoin
+### Syntax
+
+The blockchain reference in CAIP-2 is a string. This string consists of a prefix and a reference separated by a dash ("-"). Prefix and reference can only contain lower-case english letters (a..z) and numbers (0..9)
+
+    request                 = "prefix" "-" "reference"
+    prefix                  = STRING
+    reference               = STRING
+
+### Semantics
+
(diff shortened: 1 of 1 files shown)


SebastienGllmt:
SLIP44 allocates numbers to blockchains so it at least gets you part of the way there as an existing standard (https://github.com/satoshilabs/slips/blob/master/slip-0044.md). It doesn't help for differentiating networks inside a blockchain however. You can sometimes know this on project-specific levels for addresses because encodings like BECH32 encode the network inside the address

ligi:
Thanks for the feedback. Perhaps slip44 can be another prefix - I do not see it as an alternative though.

pedrouid:
I just reference this CAIP on Universal Chain Registry as we are also looking to standarize the "network id" for chains across blockchain interfaces.

We should create a second CAIP to define what is a blockchain interface. As we see it, currently there are `bitcoin`, `ethereum`, `cosmos`, `polkadot`, etc.

Each one of them have a standard interface for decentralized applications to interact with and multiple chains might share it even if they extend it beyond its standard interface.

As we currently see it we would like to namespace network ids with the interface first, example:

```
ethereum:eip155-123
cosmos:eip155-123
cosmos:cosmos-hub2
```
From the example above you can see the scenario where a chain with a cosmos interface shares the same network id with another ethereum-interface chain.

We would also need to restrict certain characters such as a colon (as it's used as a separator) from the valid characters for network ids.

Let me know what your feedback is as we would like to implement this as part of the UCR spec

ligi:
Sounds great to me - but instead of blacklisting the colon I would whitelist a-z and 0..9 - would this work for you?

pedrouid:
Yeah, that sounds good to me too. Let's whitelist a-z, 0-9 and dashes then.

ligi:
@pedrouid I added the suggested change - does this work for you?

pedrouid:
My idea would be for the prefix to indicate a blockchain interface which could be defined by another standard (CAIP-3 for example) and it would be separated by the blockchain reference using a colon (`:`) then the blockchain reference would only be valid if they used a-z, 0-9 and dashes (`-`).

Additionally given the blockchain interface they would determine how you could format the blockchain reference for each.

An example inspired by a [WCIP-12](https://github.com/WalletConnect/WCIPs/issues/12#issuecomment-523910084) and using the already defined references you used on CAIP-2 draft.
```
request                  = "interface" ":" "reference"
interface                = STRING
reference                = STRING

// example 1 - ethereum mainnet
evm:eip155-1

// example 2 - bitcoin mainnet
utxo:bitcoin-000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f

// example 3 - cosmos hub
cosmos:cosmos-hub2

// example 4 - binance chain
cosmos:binance-chain-tigris
```

In these examples we can see how a chain with EVM interface has a reference prefixed by eip155 followed by the chainId with a dash as a separator and a chain with UTXO interface has a reference prefixed by bitcoin followed by the genesis hash (like BIP122), finally a chain with COSMOS interface has a reference of the network id defined on the genesis file using the Cosmos SDK.

pedrouid:
Actually, in this example I defined the interface for bitcoin as UTXO but that might be too generic and perhaps it's best to call it bitcoin as well and define the blockchain reference to be prefixed by bip122

```
request                  = "interface" ":" "reference"
interface                = STRING
reference                = STRING

// example 1 - ethereum mainnet
evm:eip155-1

// example 2 - bitcoin mainnet
bitcoin:bip122-000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f

// example 3 - cosmos hub
cosmos:cosmos-hub2

// example 4 - binance chain
cosmos:binance-chain-tigris
```

webmaster128:
Amazing work. Can we brainstorm about potential `interface` and `reference` examples to set a min and max length along with the whitelisted characters? As people might need to store this in blockchains for certain use cases, we should not allow unnecessary long strings.

webmaster128:
In

```
// example 3 - cosmos hub
cosmos:cosmos-hub2
```

we assume that the the chain decides to use a chain ID that is in the allowed format. As far as I heard there are no restrictions to the chain ID string at Cosmos. So what do we do when a chain decides to call itself `wonderland🧝`?

ligi:
@pedrouid currently not yet sure about the rationale for this change you propose - can you elaborate a bit?
@webmaster128 currently only 0..9 and a..z are allowed - I think allowing all unicode chars is a problem - we should restrict it. But I am not really familiar with cosmos - can we perhaps hash the name? Is there no other ID?

webmaster128:
> But I am not really familiar with cosmos - can we perhaps hash the name? Is there no other ID?

Here is some information about the chain identifies in Cosmos/IBC: https://github.com/UCRegistry/registry/pull/13#discussion_r323779405. The pure chain ID is an arbitrary JSON string in Tendermint, the consensus engine that Cosmos SDK build upon. However, I think we should collect feedback from Cosmos experts on this one (which I am not).

ligi:
@webmaster128 thanks! Hope we find some cosmos experts.

also should we add SLIP44 here as a prefix? Advantage would be the coverage of chains would be higher. Disadvantage is that there is redundancy.

webmaster128:
We identified two independent issues whis Cosmos chain IDs

1. Arbitrary string format
2. For (at least) Cosmos Hub, the chain ID is changing

For 1. we disussed to have 2 separate interfaces, one limiting the chain ID format and one hashing the chain ID to make it fit into a character set. I think this is cool. I'd just make sure that a chain is in only one of those interfaces, either by forcing it into the plaintext interface if possible or by some manual decision making process. I prefer the first way.

For 2. I really like @pedrouid's idea that the realitiy is that new chains are spawned starting from the latest state of the previous chain and thus should have different chain IDs. E.g. `cosmoshub-2` -> `cosmoshub-3`. This is not ideal for a name service like IOV since for us those chains are different parts of the same chain (i.e. same accounts/addresses) but this is something that can also be handled by the application.
I think the best way to model that in a registry like UCRegistry is the concept of a parent chain. Using this approach provides a free solution for forks, where the parent chain might still continue to live when a child chain is forked out.

cc @RiccardoM

---------

> also should we add SLIP44 here as a prefix? Advantage would be the coverage of chains would be higher. Disadvantage is that there is redundancy.

I'd avoid mixing coins with chains. While it make help to create the picture in the beginning, it leads to all kind of issues. I was thinking of a SLIP44 -> chain ID mapping but this idea already breaks at 0x80000001 Testnet (all coins) which cannot be mapped to a blockchain. Same for 0x800000ea IOV which is a coin index that is used on IOV mainnet as well as different testnets. So I think the value of SLIP44 is primarily in brainstorming about what chains need to be considered.

willclarktech:
Hello, a few comments on the discussion so far:

### Format

> Let's whitelist a-z, 0-9 and dashes then.

It seems like an important detail that dashes aren't whitelisted, it's just that there's a dash separating the prefix and reference. Each of the prefix and reference is validated against the whitelist, as @ligi wrote in the document. @webmaster128's suggestion about min/max lengths also seems important. Here are a few edge cases which might help clarify the spec here:

- `-`
- `--`
- `---`
- `-somereference`
- `someprefix-`
- `someprefix-whatsthis-somereference`

### SLIP44

> I'd avoid mixing coins with chains. While it make help to create the picture in the beginning, it leads to all kind of issues.

I second this separation. Another reason to keep these things distinct is that some projects (eg Cosmos) talk about moving tokens from one chain to another. I'm not sure quite how literally these project intend this to be construed, but it seems at least conceptually possible that the same SLIP44 ID could be used to track tokens which exist on more than one chain (even if we ignore testnets).

### BIP122

From the [BIP122 definition](https://github.com/bitcoin/bips/blob/master/bip-0122.mediawiki#definition-of-chain-id):

> The chain ID of a chain is the block hash of the corresponding genesis block. For forked chains, it's the block hash of the first block after fork.

One problem with this approach is that it's often open to interpretation and highly political which counts as the original chain and which counts as the fork. At the very least it can take a while for a consensus to emerge about which chain counts as what.

You might be able to get round some aspects of the problem by saying any hard fork counts as a new chain with a new ID (as Cosmos apparently does), but you still don't cover all cases that way. Imagine a network is somehow bifurcated for a while so that two separate chains develop, and by the time the networks find each other again they've passed the point of automatic rollback. Both communities by this point are invested in their separate chain histories, so neither is willing to roll back and join the other one. In this case we have two distinct chains with the same genesis block hash and no hard fork distinguishing them. BIP122 gives them the same chain ID.

ligi:
closing in favor of #9

Mirage1348 (commented):
#87

gaowhire (commented):
1

wils81ouwehand:
Spreekt hier niemand nederlands

alokeutpal (approved):

alokeutpal (approved):
Yes

thakhin12346 (commented):
> caip-2

ALANthemoor:
> Hello, a few comments on the discussion so far:
> 
> ### Format
> > Let's whitelist a-z, 0-9 and dashes then.
> 
> It seems like an important detail that dashes aren't whitelisted, it's just that there's a dash separating the prefix and reference. Each of the prefix and reference is validated against the whitelist, as @ligi wrote in the document. @webmaster128's suggestion about min/max lengths also seems important. Here are a few edge cases which might help clarify the spec here:
> 
> * `-`
> * `--`
> * `---`
> * `-somereference`
> * `someprefix-`
> * `someprefix-whatsthis-somereference`
> 
> I assume all of these should be invalid?
> 
> ### SLIP44
> > I'd avoid mixing coins with chains. While it make help to create the picture in the beginning, it leads to all kind of issues.
> 
> I second this separation. Another reason to keep these things di// example 3 - cosmos hub
cosmos:cosmos-hub2stinct is that some projects (eg Cosmos) talk about moving tokens from one chain to another. I'm not sure quite how literally these project intend this to be construed, but it seems at least conceptually possible that the same SLIP44 ID could be used to track tokens which exist on more than one chain (even if we ignore testnets).
> 
> ### BIP122
> From the [BIP122 definition](https://github.com/bitcoin/bips/blob/master/bip-0122.mediawiki#definition-of-chain-id):
> 
> > The chain ID of a chain is the block hash of the corresponding genesis block. For forked chains, it's the block hash of the first block after fork.t
> 
> One problem with this approach is that it's often open to interpretation and highly political which counts as the original chain and which counts as the fork. At the very least it can take a while for a consensus to emerge about which chain counts as what.
> 
> You might be able to get round some aspects of the problem by saying any hard fork counts as a new chain with a new ID (as Cosmos apparently does), but you still don't cover all cases that way. Imagine a network is somehow bifurcated for a while so that two separate chains develop, and by the time the networks find each other again they've passed the point of automatic rollback. Both communities by this point are invested in their separate chain histories, so neither is willing to roll back and join the other one. In this case we have two distinct chains with the same genesis block hash and no hard fork distinguishing them. BIP122 gives them the same chain ID, or at least no decision procedure for designating one as the original and the other as the fork.

Hardikgosai1 (commented):
Tx

jk3691:
Main

PR #9, opened 2019-12-05
CAIP-2/3/4/5/6/7: Blockchain IDs

### LINKED ISSUE #2: CAIP-2 (Chain ID): General format definition
This is a refinement of #1 and especially https://github.com/ChainAgnostic/CAIPs/pull/1#issuecomment-534692931 and kudos go to everyone envolved.

The goals of the general chain ID format is:
1. Uniqueness within the entire blockchain ecosystem
2. To some degree human readable and helps for basic debugging
3. Restricted in a way that it can be stored on chain
4. Character set basic enough to display in hardware wallets as part of a transaction content
5. Maybe: Can be used unescaped in URL paths
6. Maybe: Can be used as filename in a case-sensitive UNIX file system (Linux/git).
7. ~Maybe: Can be used as filename in a case-insensitive UNIX file system (macOS).~
8. ~Maybe: Can be used as filename in a Windows file system ([that one is going to be fun](https://stackoverflow.com/questions/1976007/what-characters-are-forbidden-in-windows-and-linux-directory-names))~.

5.-8. are open questions to me and I'd love to hear feedback on those. Especially the requirement for a case-insensitive format seems seems to limit the implementation of `reference`.

```
chain_id     = a case-sensitive string in the form: interface + ":" + reference
interface    = a case-sensitive string in the form: [-a-z]{3,16}
reference    = a case-sensitive string in the form: [-a-zA-Z0-9]{3,47}
```

where `interface` identifies a document describing how `reference` is composed.

The `interface` typically corresponds to a class of blockchains and allows delegation of
the `reference` format to an ecosystem-specific standard.
Example interfaces are "ethereum" for all the blockchains that can be identified by EIP-155,
"bitcoin" for Bitcoin and closely related projects as Litecoin, or "cosmos" for projects that
build on top of Tendermint.

`reference` should strive for uniqieness within the interface. For user-choosen chain IDs
(e.g. EIP-155 or Tendermint chain ID), uniqueness cannot be guaranteed and it is reponsibility
of registries and communities to resolve collisions.

## Examples

```
# Ethereum mainnet
ethereum:eip155-1

# Bitcoin mainnet (see https://github.com/bitcoin/bips/blob/master/bip-0122.mediawiki#definition-of-chain-id)
bitcoin:bip122-000000000019d6689c085ae165831e93

# Litecoin
bitcoin:bip122-12a765e31ffd4059bada1e25190f6e98

# Feathercoin (Litecoin fork)
bitcoin:bip122-fdbe99b90c90bae7505796461471d89a

# Cosmos Hub (Tendermint + Cosmos SDK)
cosmos:cosmoshub-2
cosmos:cosmoshub-3

# Binance chain (Tendermint + Cosmos SDK; see https://dataseed5.defibit.io/genesis)
cosmos:Binance-Chain-Tigris

# IOV Mainnet (Tendermint + weave)
cosmos:iov-mainnet

# Lisk Mainnet (LIP-0009; see https://github.com/LiskHQ/lips/blob/master/proposals/lip-0009.md)
lisk:lip9-9ee11e9df416b18b

# Random max length (16+1+47 = 64 chars/bytes)
blockchain1-hash:xip3343-8c3444cf8970a9e41a706fab93e7a6c4-xxxyyy
```

Open questions

- [x] Is there a very good reason for a case-insensitive format?
- [x] Is 100 ASCII chars a resonable max length?
- [ ] What about whitelisting `_` as an alternative separator in `reference`?

Thoughts?

### LINKED ISSUE #3: CAIP-2 (Chain ID): The Ethereum interface
This is ticket is about the details of the Ethereum interface for #2.

-------

## Interface name

The name of this interface should either be "ethereum", referring to the wider Ethereum ecosystem.

## Reference definition

The definition is delegated to [EIP155](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md). The reference format is `eip155-%d`, where `%d` is an unsigned integer in decimal represenation and corresponds to `CHAIN_ID` of EIP155.

Note: due to length restrictions of the `reference` field, the largest supported `CHAIN_ID` is 9999999999999999999999999999999999999999.

## Examples

```
# Ethereum mainnet
ethereum:eip155-1

# Görli
ethereum:eip155-5

# Auxilium Network Mainnet
ethereum:eip155-28945486
```

Open questions:

- [x] ~Do we want interface name `ethereum` or `evm`?~ `ethereum`, see https://github.com/ChainAgnostic/CAIPs/issues/3#issuecomment-559554777
- [x] ~Should we drop the "eip155-" prefix?~ No, see https://github.com/ChainAgnostic/CAIPs/issues/3#issuecomment-559864996

### LINKED ISSUE #4: CAIP-2 (Chain ID): The Bitcoin interface
This is ticket is about the details of the Bitcoin interface for #2.

-------

## Interface name

The name of this interface should be "bitcoin", referring to the wider Bitcoin ecosystem, including multiple forks of the chain and the code.

## Reference definition

The definition is delegated to [BIP122](https://github.com/bitcoin/bips/blob/master/bip-0122.mediawiki#definition-of-chain-id). The reference format is `bip122-%s`, where `%s` is a 32 character prefix of the block hash from BIP122 (lower case hex).

## Blockchains in this interface

* Bitcoin + chain forks
* Litecoin + chain forks (e.g. Feathercoin)

## Blockchains not in this interface

* ZCash: BIP122 excluded explicitely by https://github.com/zcash/zips/issues/87#issuecomment-250988155. Further research may be necessary to understand why and if there is an alternative blockchain ID specification.

## Examples

```
# Bitcoin mainnet (see https://github.com/bitcoin/bips/blob/master/bip-0122.mediawiki#definition-of-chain-id)
bitcoin:bip122-000000000019d6689c085ae165831e93

# Litecoin
bitcoin:bip122-12a765e31ffd4059bada1e25190f6e98

# Feathercoin (Litecoin fork)
bitcoin:bip122-fdbe99b90c90bae7505796461471d89a
```

Open questions:

- [ ] Should we drop the "bip122-" prefix?
- [ ] How to find a list of important blockchains that can be covered by this interface? What about e.g. Bitcoin Cash, Monero, ~Zcash~, …

### LINKED ISSUE #5: CAIP-2 (Chain ID): The Cosmos interfaces
This is ticket is about the details of the Cosmos interface for #2.

-------

## Interface name

The name of this interface is "cosmos", referring to the wider Cosmos ecosystem.

## Reference definition

The `reference` uses the Tendermint `chain_id` from the genesis file directly (a JSON-compatible unicode string), assuming it matches the case-sensitive pattern `[-a-zA-Z0-9]{3,47}`. Otherwise the [Cosmos Hash interface](https://github.com/ChainAgnostic/CAIPs/issues/6) must be used.

## Blockchains in this interface

* Cosmos SDK blockchains with a chain ID matching `[-a-zA-Z0-9]{3,47}`: Cosmoshub, Binance, [Cosmos Testnets](https://github.com/cosmos/testnets), …
* IOV blockchains (Tendermint + weave)

### Changing chain IDs

During the development of this chain ID definition, we came across changing chain IDs for Cosmos Hub (`cosmoshub-1`, `cosmoshub-2`, `cosmoshub-3`). A new chain ID is assigned every time Cosmos Hub dumps the current blockchain state and creates a new genesis from the old state. Technicalls this leads to different blockchains and can (and maybe should) treated as such. For this specification, we treat them as different blockchains. It is responsibility of a higher level application to interprete some chains as sequels of each other or create equality sets.

## Examples

```
# Cosmos Hub (Tendermint + Cosmos SDK)
cosmos:cosmoshub-2
cosmos:cosmoshub-3

# Binance chain (Tendermint + Cosmos SDK; see https://dataseed5.defibit.io/genesis)
cosmos:Binance-Chain-Tigris

# IOV Mainnet (Tendermint + weave)
cosmos:iov-mainnet
```

Open questions:

- [ ] Should characters other than `[-a-zA-Z0-9]` be whitelisted (see also #2)

### LINKED ISSUE #6: CAIP-2 (Chain ID): The Cosmos Hash interfaces
This is ticket is about the details of the Cosmos Hash interface for #2.

-------

## Interface name

The name of this interface is "cosmos-hash", where "cosmos" refers to the wider Cosmos ecosystem and "hash" to the fact that native chain IDs are hashed in order to fit in the [general format](https://github.com/ChainAgnostic/CAIPs/issues/2).

## Reference definition

The `reference` uses the format `hex(sha256(utf8(chain_id)))`, with

* the Tendermint `chain_id` from the genesis file (a JSON-compatible unicode string)
* `utf8` being the UTF-8 encoding
* `sha256` being the the SHA256 hash function
* `hex` being a lowecase hex encoder

## Blockchains in this interface

* Tendermint-based blockchains with a chain ID matching not matching `[-a-zA-Z0-9]{3,47}`: no real world example known yet

## Blockchains not in this interface

* If the `chain_id` is compatible to the [Cosmos interface](https://github.com/ChainAgnostic/CAIPs/issues/5), that one must be used.

## Examples

```
# Example chain with chain_id "wonderland🧝"
cosmos-hash:843d2fc87f40eeb9198e0c2416be28a6c31a01eb9cc32499b030723d662c006f
```

Open questions:

- [ ] Do we need to define unicode normalization?

### LINKED ISSUE #7: CAIP-2 (Chain ID): The Lisk interface
This is ticket is about the details of the Lisk interface for #2.

-------

## Interface name

The name of this interface should be "lisk", referring to the wider Lisk ecosystem.

## Reference definition

The definition is delegated to [LIP9](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0009.md). The reference format is `lip9-%s`, where `%s` is a 16 character prefix of the hash from LIP9 (lower case hex).

## Blockchains in this interface

* Lisk Mainnet and Testnet
* Forks
* Side chains

## Examples

```
# Lisk Mainnet (https://github.com/LiskHQ/lips/blob/master/proposals/lip-0009.md#appendix-example)
lisk:lip9-9ee11e9df416b18b

# Lisk Testnet (echo -n "da3ed6a45429278bac2666961289ca17ad86595d33b31037615d4b8e8f158bbaLisk" | sha256sum | head -c 16)
lisk:lip9-e48feb88db5b5cf5
```

Open questions:

- [ ] Should we drop the "lip9-" prefix?

webmaster128:
This wraps up the results of the discussions in #2, #3, #4, #5, #6, #7 and supersedes the first draft from #1. I think I covered all the discussion. IOV is ready to build a name service on top of this specification.

Closes #1
Closes #2
Closes #3
Closes #4
Closes #5
Closes #6
Closes #7


CAIPs/caip-3.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Reference definition
+
+The definition is delegated to [EIP155](https://eips.ethereum.org/EIPS/eip-155). The format is `eip155-%d`, where `%d` is an unsigned integer in decimal represenation and corresponds to `CHAIN_ID` of EIP155.
+
+Note: due to length restrictions of the reference field (47 characters), the largest supported `CHAIN_ID` is 9999999999999999999999999999999999999999.
+
+## Rationale
+
+The chain ID defined in EIP155 is the most widely used chain identifier in the Ethereum ecosystem known to the authors. It strives for uniqueness and the fact that the standrad is used for reply protection ensure that creators of a new Ethereum network have an incentive to use an ID that is not used elsewhere.
```
willclarktech: Spelling: "the standard is used for replay protection"

CAIPs/caip-4.md
```diff
@@ -0,0 +1,66 @@
...
+
+### Reference definition
+
+The definition is delegated to [BIP122](https://github.com/bitcoin/bips/blob/master/bip-0122.mediawiki#definition-of-chain-id). The format is `bip122-%s`, where `%s` is a 32 character prefix of the block hash from BIP122 (lower case hex).
+
+## Rationale
+
+We delegate the identification of Bitcoin-like chains to BIP122, as this is the best Bitcoin chain identification standard known to the authors of this CAIP.
+
+In order to prepare for other network identification standards within the Ethereum interface, we keep the "eip155-" prefix, which seems unneeded at first glance.
```
willclarktech: Copypasta: "Bitcoin" and "bip122"

CAIPs/caip-4.md
```diff
@@ -0,0 +1,66 @@
...
+## Backwards Compatibility
+
+Not applicable
+
+## Test Cases
+
+This is a list of manually composed examples
+
+```
+# Ethereum mainnet
```
willclarktech: Also need to be updated.

CAIPs/caip-5.md
```diff
@@ -0,0 +1,71 @@
+---
+caip: 5
+title: The Cosmos interface for CAIP-2
+author: Simon Warta (@webmaster128)
+discussions-to: https://github.com/ChainAgnostic/CAIPs/issues/5, https://github.com/ChainAgnostic/CAIPs/pull/1
+status: Draft
+type: Standard
+created: 2019-12-05
+requires: 2, 6
```
willclarktech: Are circular dependencies allowed? Are they strict requirements in both directions here?

CAIPs/caip-5.md
```diff
@@ -0,0 +1,71 @@
...
+
+The name of this interface is "cosmos", referring to the wider Cosmos ecosystem.
+
+### Reference definition
+
+The reference uses the Tendermint `chain_id` from the genesis file directly (a JSON-compatible unicode string), assuming it matches the case-sensitive pattern `[-a-zA-Z0-9]{3,47}`. Otherwise the Cosmos Hash interface (CAIP-6) must be used.
+
+## Rationale
+
+Blockchains in this interface are [Cosmos SDK](https://github.com/cosmos/cosmos-sdk) blockchain (e.g. Cosmoshub, Binance, Cosmos Testnets) and [Weave](https://github.com/iov-one/weave) based blockchains (e.g. IOV) with a chain ID matching `[-a-zA-Z0-9]{3,47}`.
```
willclarktech: Typo: "Cosmos SDK blockchains"

CAIPs/caip-5.md
```diff
@@ -0,0 +1,71 @@
...
+
+While there is no enforced restriction on chain_id, the author of this document did not find a
+non-conforming chain ID in the wild. There is [a discussion about documenting a best practice chain ID pattern](https://github.com/cosmos/cosmos-sdk/issues/5363).
+
+During the development of this chain ID definition, we came across changing chain IDs for Cosmos Hub (`cosmoshub-1`, `cosmoshub-2`, `cosmoshub-3`). A new chain ID is assigned every time Cosmos Hub dumps the current blockchain state and creates a new genesis from the old state. Technically this leads to different blockchains and can (and maybe should) treated as such. For this specification, we treat them as different blockchains. It is responsibility of a higher level application to interprete some chains as sequels of each other or create equality sets.
```
willclarktech: Spelling/grammar: "It is the responsibility of a higher level application to interpret"

CAIPs/caip-5.md
```diff
@@ -0,0 +1,71 @@
...
+# Binance chain (Tendermint + Cosmos SDK; see https://dataseed5.defibit.io/genesis)
+cosmos:Binance-Chain-Tigris
+
+# IOV Mainnet (Tendermint + Weave)
+cosmos:iov-mainnet
+```
+
+## Links
+
+- [EIP155](https://eips.ethereum.org/EIPS/eip-155)
```
willclarktech: Links need to be updated.

CAIPs/caip-6.md
```diff
@@ -0,0 +1,72 @@
+---
+caip: 6
+title: The Ethereum interface for CAIP-2
```
willclarktech: Needs to be updated.

CAIPs/caip-6.md
```diff
@@ -0,0 +1,72 @@
...
+requires: 2, 5
+---
+
+## Simple Summary
+
+This document is about the details of the Cosmos Hash interface for CAIP-2.
+
+## Abstract
+
+In CAIP-2 a general blockchain identification scheme is defined. This is the
```
willclarktech: Typo: "due to"

CAIPs/caip-6.md
```diff
@@ -0,0 +1,72 @@
...
+The name of this interface is "cosmos-hash", where "cosmos" refers to the wider Cosmos ecosystem and "hash" to the fact that native chain IDs are hashed in order to fit in the general format of CAIP-2.
+
+### Reference definition
+
+The `reference` is defined as `first_16_chars(hex(sha256(utf8(chain_id))))`, with
+
+- the Tendermint `chain_id` from the genesis file (a JSON-compatible unicode string)
+- `utf8` being the UTF-8 encoding
+- `sha256` being the SHA256 hash function
+- `hex` being a lowecase hex encoder
```
willclarktech: Typo: "lowercase"

CAIPs/caip-6.md
```diff
@@ -0,0 +1,72 @@
...
+# chain_id "123456789012345678901234567890123456789012345678" (too long for CAIP-5)
+cosmos-hash:0204c92a0388779d
+
+# chain_id "wonderland🧝" (invalid character for CAIP-5)
+cosmos-hash:843d2fc87f40eeb9
+```
+
+## Links
+
+- [EIP155](https://eips.ethereum.org/EIPS/eip-155)
```
willclarktech: Need to be updated.

CAIPs/caip-7.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Test Cases
+
+This is a list of manually composed examples
+
+```
+# Lisk Mainnet (https://github.com/LiskHQ/lips/blob/master/proposals/lip-0009.md#appendix-example)
+lisk:lip9-9ee11e9df416b18b
+
+# Lisk Testnet (echo -n "da3ed6a45429278bac2666961289ca17ad86595d33b31037615d4b8e8f158bbaLisk" | sha256sum | head -c 16)
```
willclarktech: Is "Lisk" really supposed to be on the end of the hash?

webmaster128: Yes, it is a concatenation of nethash and "community identifier", see https://github.com/LiskHQ/lips/blob/master/proposals/lip-0009.md#specification

CAIPs/caip-5.md
webmaster128: I think circular dependencies don't harm in this context. It better models the problem than pretensing one interface does not know of the other.

webmaster128:
Thanks @willclarktech, all comments implemented

pedrouid:
I would just update the titles for these to be clearer:

* CAIP-2 describes a standard for chain_id (which consist of interface + reference)
* CAIP-3 to CAIP-7 describe standards for reference of a specific interface (which all of these require CAIP-2)

So my suggestion for the titles would be: 

* CAIP-2: chain_id format
* CAIP-[3-7]: reference format for <INTERFACE> chain

CAIPs/caip-2.md
```diff
@@ -0,0 +1,110 @@
+---
+caip: 2
+title: Blockchain ID
```
pedrouid: Blockchain chain_id format

CAIPs/caip-3.md
```diff
@@ -0,0 +1,69 @@
+---
+caip: 3
+title: The Ethereum interface for CAIP-2
```
pedrouid: Blockchain reference for Ethereum interface

CAIPs/caip-4.md
```diff
@@ -0,0 +1,66 @@
+---
+caip: 4
+title: The Bitcoin interface for CAIP-2
```
pedrouid: Blockchain reference for Bitcoin interface

CAIPs/caip-5.md
```diff
@@ -0,0 +1,70 @@
+---
+caip: 5
+title: The Cosmos interface for CAIP-2
```
pedrouid: Blockchain reference for Cosmos interface

CAIPs/caip-6.md
```diff
@@ -0,0 +1,71 @@
+---
+caip: 6
+title: The Cosmos Hash interface for CAIP-2
```
pedrouid: Blockchain reference hash for Cosmos interface

CAIPs/caip-7.md
```diff
@@ -0,0 +1,61 @@
+---
+caip: 7
+title: The Lisk interface for CAIP-2
```
pedrouid: Blockchain reference for Lisk interface

pedrouid (changes requested):

CAIPs/caip-2.md
webmaster128: I'd try to keep the technical variable name `chain_id` out of a title. Is "format" the right word here? What about "Blockchain ID specification"?

CAIPs/caip-3.md
webmaster128: Should we rally drop the link to CAIP-2?

pedrouid:
I just noticed the CAIP-6 about the reference hash for Cosmos chain and it doesn't follow the CAIP-2 if the interface is defined as `cosmos-hash`. The interface is still `cosmos` so to indicate that it's a hash instead it should be a prefix to the hash

*BEFORE*
```
# chain_id "x" (too short for CAIP-5)
cosmos-hash:2d711642b726b044
# chain_id "123456789012345678901234567890123456789012345678" (too long for CAIP-5)
cosmos-hash:0204c92a0388779d
# chain_id "wonderland🧝" (invalid character for CAIP-5)
cosmos-hash:843d2fc87f40eeb9
```

*AFTER*
```
# chain_id "x" (too short for CAIP-5)
cosmos:hash-2d711642b726b044
# chain_id "123456789012345678901234567890123456789012345678" (too long for CAIP-5)
cosmos:hash-0204c92a0388779d
# chain_id "wonderland🧝" (invalid character for CAIP-5)
cosmos:hash-843d2fc87f40eeb9
```

Additionally, I wouldn't add the requirement for CAIP-5 on CAIP-6 but instead just describe this as a standard to normalize Cosmos chain_id's which CAIP-5 will require

CAIPs/caip-5.md
pedrouid: I think in this case, it's not really required to make them circular because the use of CAIP-6 to normalize chain_id format is independent IMO of the use-case. In fact, I would add that CAIP-6 could be used to normalize any chain_id regardless of interface

webmaster128:
> and it doesn't follow the CAIP-2 if the interface is defined as `cosmos-hash`

Why not?

> cosmos:hash-2d711642b726b044

The issue with this approach is that you cannot differentiate between a chain called "hash-2d711642b726b044" and a chain ID hased to "2d711642b726b044".

CAIPs/caip-2.md
pedrouid: Agreed, specification is a better name than format. Let's call it "Blockchain ID Specification"

CAIPs/caip-3.md
pedrouid: Yes, because the title shouldn't describe the requirements. We just refer it on `require: 2`. Additionally standards are modular because they can be used individually. I could just make use of CAIP-3 for something other chan CAIP-2 for example. I think @ligi would be able to provide more clarity on this than me

ligi (commented):
Thanks for making the PR and the split into multiple CAIP's - I like this approach as it makes it easy to add new ones without modifying a CAIP and hence lead to more immutability.
Looks good to me so far - just dislike the introduced redundance for Ethereum and Bitcoin - IMHO they do not need the added prefixes - so instead of:

`ethereum:eip155-5`

IMHO we can leave it as:

`eip155:5`

 and bitcon:

`bitcoin:bip122-000000000019d6689c085ae165831e93`

can IMHO be left like this:

`bip122:000000000019d6689c085ae165831e93`

CAIPs/caip-5.md
ligi: Yea - let's please avoid circular dependencies

webmaster128:
At this point we probably need to get some clarity on what the purpose of the `xychain:` part of the ID is.
- I hear the desire to merge cosmos/cosmos-hash into one thing
- Dedup bitcoin/bip122 and ethereum/eip155 (which @pedrouid disagrees with)

Maybe the term "interface" carries too much expectations and calling it "namespace" for the purpose of this chain ID specification is more appropriate. Then we can define the namespaces in ecosystem specific documents, like

- CAIP-3 Ethereum: namespace `eip155:`
- CAIP-4 Bitcoin: namespace `bip122:`
- CAIP-5 Cosmos: namespace `cosmos:` and namespace `cosmos-hash:`
- CAIP-7 Lisk: namespace `lip9:`

Any non-eip155 namespace for Ethereum could be an extension of CAIP-3.

pedrouid:
> Maybe the term "interface" carries too much expectations and calling it "namespace" for the purpose of this chain ID specification is more appropriate. Then we can define the namespaces in ecosystem specific documents, like
> 
> * CAIP-3 Ethereum: namespace `eip155:`
> * CAIP-4 Bitcoin: namespace `bip122:`
> * CAIP-5 Cosmos: namespace `cosmos:` and namespace `cosmos-hash:`
> * CAIP-7 Lisk: namespace `lip9:`

Well in that case we could have all interfaces to follow a standard 

* CAIP-3 Ethereum: namespace `eip155:`
* CAIP-4 Bitcoin: namespace `bip122:`
* CAIP-5 Cosmos: namespace `caip6:` 
* CAIP-7 Lisk: namespace `lip9:`

Any reason for not simplifying it like this? Perhaps it shouldn't be called `interface` but just `specification` standard. So we would update CAIP-2 as follows:

```
chain_id:    specification + ":" + reference
specification:   [-a-z]{3,16}
reference:   [-a-zA-Z0-9]{3,47}
```

webmaster128:
I'll stick with the term "namespace" for now to avoid introducing more terms. So I'm thinking about

```
chain_id:        namespace + ":" + reference
namespace:       [-a-z]{3,16}
reference:       [-a-zA-Z0-9]{3,47}
```

> Any reason for not simplifying it like this?

I don't think a namespace needs to to be calld as any standard, just because it works for Ethereum, Bitcoin and Lisk. I also don't think we need a 1:1 mapping between CAIPs and namespaces. It would be very easy to define both the `cosmos:` as well as the `cosmos-hash:` namespace as part of a single document.

A shared namespace for `cosmos:`/`cosmos-hash:` created collisions, as described here: https://github.com/ChainAgnostic/CAIPs/pull/9#issuecomment-563308085. By having two different namespaces in the same document, we get a nice trade off: only one CAIP, no circular dependencies between CAIPs, collision resistence.

> This would also remove the need for all other CAIPs in this PR

Which other CAIPs? I think we agree that having one CAIP per ecosystem is desired and the only question is if we can merge CAIP-6 and CAIP-7.

CAIPs/caip-2.md
webmaster128: Implemented in https://github.com/ChainAgnostic/CAIPs/pull/9/commits/84431bc7f77c2c5bb4a8496c65d94557a9223fd8. I assume we use the headline style that capitalize (almost) every word then. Resolving this for now, feel free to re-open.

CAIPs/caip-3.md
webmaster128: Implemented in https://github.com/ChainAgnostic/CAIPs/pull/9/commits/c803d602f33596b7660fba63b91c5dd7bebea8d5. Please bear in mind that the goal is to merge the two Cosmos documents into one document one way or another, so "Blockchain Reference for Cosmos (Hashed Variant)" should be considered a temporary working title. Resolving this.

CAIPs/caip-5.md
webmaster128: In https://github.com/ChainAgnostic/CAIPs/pull/9/commits/c51894d7e951314d8459803ca105b78dd42b1b94 the two namespaces "cosmos:" and "cosmos-hash:" got merged into one document, avoiding the need for circular dependencies.

webmaster128 (commented):
With 619a1ed67ca60393849978553a3a642fdf84a492 and c51894d7e951314d8459803ca105b78dd42b1b94 we achieve

- Decupling of interface and namespaces
- Multiple namespaces per document possible
- Merge of CAIP-5/CAIP-6, avoiding circular dependencies between them

This leads us to the structure

```
CAPI-3 Blockchain Reference for Ethereum
  - namespace "ethereum"
    - prefix "eip155"
    - other prefixes possible

CAPI-4 Blockchain Reference for Bitcoin
  - namespace "bitcoin"
    - prefix "bip122"
    - other prefixes possible

CAPI-5 Blockchain Reference for Cosmos
  - namespace "cosmos"
  - namespace "cosmos-hash"

CAPI-7 Blockchain Reference for Lisk
  - namespace "lisk"
    - prefix "lip9"
    - other prefixes possible
```

Regarding the open questions:
1. I'd keep ecosystem specific CAIPs that explain _how_ standards like BIP122 are used. This is especially required for truncating. It also helps to cover the discussion around the usage of an external standard in our context.
2. I'd like to keep namespaces "cosmos"/"cosmos-hash" separate to avoid collisions and to keep the happy path simple. An alternative would be to use `cosmos:direct-native-id-from-chain42` and `cosmos:hash-716248761abab` which pushes the differentiation down to prefix level at the cost of adding the "direct-" prefix for all real worlds IDs.
3. I'm not convinced we should use the eips/bips/lips as the top-level identifier. One reason is that "bip122" / "eip155" are not as intuitive for newcomers. The other reason is that I see this as two level hierarchy: you have the bitcoin ecosystem and within that, you are using BIP122.

webmaster128:
There is another possible strategy to avoid the collisions in Cosmos: we can blacklist `/^hash-/` for the direct version, such that chain IDs starting with "hash-" need to be hased. This way we know that every ID starting with "hash-" was hashed.

antoineherzog:
I personally prefer the other strategy to blacklist /^hash-/ for the direct version, such that native chain IDs starting with "hash-" need to be hashed. @pedrouid are you good with that?

ligi:
Sorry wanting to look into this a bit earlier but the 2020 migration and 36c3 caused a deleay.
I am still thinking we-should collapse namespace and prefix. As I fail to see the concrete advantage of it and so do not think the added redundancy and complexity should be there.

Means - this proposal suggests `ethereum:eip155-1` and I suggest to stick to `eip155:1` as suggested in the original PR #1 

But this is not a very strong opinion (no over my dead body) so if this one has rough consensus I am also OK with going this route. But I would prefer the other and like to hear arguments what *concrete* advantages the separated namespace brings or will bring.

antoineherzog:
Ok I write here the summary of the call with Ligi:

CAPI-3 Blockchain Reference for Ethereum following eip155
  - namespace "eip155"
  - reference is an unsigned integer in decimal represenation and corresponds to CHAIN_ID of EIP155.

as example **eip155:1** for ethereum mainnet

CAPI-4 Blockchain Reference for Bitcoin following bip122
  - namespace "bip122"
  - reference is a 32 character prefix of the block hash from BIP122 (lower case hex).

as example **bip122:000000000019d6689c085ae165831e93** for bitcoin mainnet

CAPI-5 Blockchain Reference for Cosmos
  - namespace "cosmos"
  - reference uses the Tendermint chain_id from the genesis file directly (a JSON-compatible unicode string), assuming it matches the case-sensitive pattern [-a-zA-Z0-9]{3,47}.

as example **cosmos:iov-mainnet** for iov mainnet

CAPI-7 Blockchain Reference for Lisk following lip9
  - namespace "lip9"
  - reference format is a 16 character prefix of the hash from LIP9 (lower case hex).

webmaster128:
Okay, good. I also don't have any killer argument. I just think that "eip155:57564" is not welcoming to newcomers as it is very cryptic and gives no hint that this has something to do with Ethereum. But yeah, luckly I know what it means and am fine with changing it.

Moving the standard one level up can lead to a structure like

```
CAPI-3: Blockchain Reference for Ethereum (following EIP155)
CAPI-4: Blockchain Reference for Bitcoin (following BIP122)
CAPI-2356: Blockchain Reference for Bitcoin (following BIP1265)
CAPI-2357: Blockchain Reference for Bitcoin (following BIP1266)
CAPI-3213: Blockchain Reference for Ethereum (following EIP2155)
```

which should make the CAIPs pretty much immutable. So yeah, as Python Zen says: "Flat is better than nested.".

Other things I plan to change:
- Rename CAIP-7 to CAIP-6
- Blacklist prefix "hash-" for unhashed cosmos chain IDs

ligi:
>  I just think that "eip155:57564" is not welcoming to newcomers

this should not be end-user facing anyway. Normal "users"/newcomers should not really see it 

Moving one layer up is actually a great idea! This solves some problems of a mapping - would support the change

webmaster128:
0dacaa303fd1a63f7ed3a687b54ad08b6a4f5c88 contains the change from Ethereum to EIP155. @ligi could you check this? If that is what is desired, I change Bitcoin and Lisk accordingly

ligi:
@webmaster128 Thanks - LGTM! Would just also change Namespace -> Prefix

webmaster128:
@ligi 👍 
It is called namespace consistently now. What's the purpose of such a renaming?

ligi:
OK leave it at namespace then

webmaster128:
I think all discussed changes are included now. At https://github.com/webmaster128/CAIPs/compare/c51894d7e951314d8459803ca105b78dd42b1b94...chain-id you see a diff of all changes done post-36c3

CAIPs/caip-2.md
```diff
@@ -0,0 +1,111 @@
...
+The blockchain ID (short "chain ID") is a string designed to uniquely identify blockchains in a developer-friendly fashion.
+
+### Syntax
+
+The `chain_id` is a case-sensitive string in the form
+
+```
+chain_id:    namespace + ":" + reference
+namespace:   [-a-z]{3,16}
+reference:   [-a-zA-Z0-9]{3,47}
```
antoineherzog: It should change to 1 because for ethereum main this is: eip155:1

webmaster128: Nice catch! In this case, we should also open up the cosmos battern to allow chain "x" to be encoded as "cosmos:x"

antoineherzog: Yes I agree! If you could add my name on the spec in author or reviewer that's cool also :)

webmaster128: Done in 69f92ed

CAIPs/caip-2.md
```diff
@@ -0,0 +1,111 @@
...
+cosmos:Binance-Chain-Tigris
+
+# IOV Mainnet (Tendermint + weave)
+cosmos:iov-mainnet
+
+# Lisk Mainnet (LIP-0009; see https://github.com/LiskHQ/lips/blob/master/proposals/lip-0009.md)
+lip9:9ee11e9df416b18b
+
+# Random max length (16+1+47 = 64 chars/bytes)
+blockchain1-hash:xip3343-8c3444cf8970a9e41a706fab93e7a6c4-xxxyyy
```
pedrouid: Typo `blockchain1-hash` should be `blockchain1:hashed-`

pedrouid:
Great work guys! 💪 I really like where these standards ended up

CAIPs/caip-2.md
webmaster128: This is a dummy example of a max length chain ID. As this does not exist anyways, it can be any namespace. The important thing is length 17.

pedrouid: Gotcha let's just update the example's namespace to remove ambiguity with hashed references

CAIPs/caip-2.md
```diff
@@ -0,0 +1,111 @@
...
+
+The blockchain ID (short "chain ID") is a string designed to uniquely identify blockchains in a developer-friendly fashion.
+
+### Syntax
+
+The `chain_id` is a case-sensitive string in the form
+
+```
+chain_id:    namespace + ":" + reference
+namespace:   [-a-z]{3,16}
```
pedrouid: namespace is missing whole numbers

CAIPs/caip-2.md
pedrouid: I tried a few iterations, maybe `examplenamespace`

pedrouid: or even better `max-namespace-16`

CAIPs/caip-2.md
pedrouid: it should be 
```
namespace:   [-a-z0-9]{3,16}
```

webmaster128: Done in 3513aa26710c4d3ed028aa7dba1bde3a69b239af

CAIPs/caip-2.md
webmaster128: Done in 74ce2ba95872c1eee65fa977961b40f935daa77c

pedrouid (approved):
Awesome 🎉 Let's get the approval of @antoineherzog and @ligi too

romeo4934 (approved):

CAIPs/caip-3.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Specification
+
+### EIP155 Namespace
+
+The namespace is called "eip155" as in [EIP155](https://eips.ethereum.org/EIPS/eip-155).
+
+#### Reference Definition
+
+The definition is delegated to EIP155. The format is an unsigned integer in decimal represenation and corresponds to `CHAIN_ID` of EIP155.
```
ligi: typo
```suggestion
The definition is delegated to EIP155. The format is an unsigned integer in decimal representation and corresponds to `CHAIN_ID` of EIP155.
```

ligi (changes requested):
Only this one typo - afterwards I would merge it

CAIPs/caip-3.md
webmaster128: Hmm, why can't I accept this change?

ligi: hm - I have the button - but I cannot press it:

Selection_326

webmaster128: Done in 5740ce43f028367d34b885209e5a4d7a5917f669

webmaster128: Done in 5740ce43f028367d34b885209e5a4d7a5917f669

ligi (approved):

ligi:
PR was merged

webmaster128:
Why did you close this PR manually? Github usually understands PR merges via the web UI merge button and command line merge+push. Now all the autoclose commands from the description have not been triggered.

ligi:
Yea - first time this happened to me - maybe some github bug. I merged via the UI - but then it told me "there is a conflict" - need to merge manually - but everything was merged already ..

webmaster128:
I see. Yeah, Github is a bit slow and strange at the moment. Anyways, great to have this in 🥂

CAIPs/caip-5.md
```diff
@@ -0,0 +1,71 @@
...
+## Test Cases
+
+This is a list of manually composed examples
+
+```
+# Cosmos Hub (Tendermint + Cosmos SDK)
+cosmos:cosmoshub-2
+cosmos:cosmoshub-3
+
+# Binance chain (Tendermint + Cosmos SDK; see https://dataseed5.defibit.io/genesis)
```
sumit12108, line 78: To be added

Dustin4444 (commented):
#403

Issue #17, opened 2020-04-20
Define a CAIP standard for cross-chain asset ids


michaelsena:
This would allow on-chain assets to be referenced in a chain-agnostic way! 🚀

ligi:
@michaelsena can this be closed now after merging #19 ?

oed:
@ligi yes it can be closed!

PR #18, opened 2020-04-23
Add CAIP for NFT tokens

oed:
Partially addresses #17

### DIFF
--- a/CAIPs/caip-18.md
+++ b/CAIPs/caip-18.md
@@ -1,0 +1,69 @@
+---
+caip: 18
+title: NFT ID Specification
+author: Joel Thorstensson (@oed)
+discussions-to: https://github.com/ChainAgnostic/CAIPs/pull/18
+status: Draft
+type: Standard
+created: 2020-04-23
+updated: 2020-04-23
+requires: 2
+---
+
+## Simple Summary
+
+CAIP-18 defines a way to identify a specific NFT token on any blockchain specified by CAIP-2 blockchain id.
+
+## Abstract
+
+This proposal aims to facilitate specifying NFT tokens on any blockchain extending CAIP-2 blockchain id specification. This is useful for both decentralized applications and wallets to specify NFT tokens on multiple chains using string identifiers specific to each chain. Currently wallets are usually designed for each chain and multi-chain wallets use proprietray data structures to differentiate NFTs. This proposal aims to standardize these identifiers for NFTs to allow interoperability.
+
+## Motivation
+
+The motivation for proposal stem from designing a chain-agnostic protocol for communication between dapps and wallets that was independent of any blockchain but provide the flexibility to be backwards compatible with existing applications.
+
+## Specification
+
+The NFT id specification will be prefixed with the NFT token ID and delimited with an at sign (`@`) followed by a CAIP-2 blockchain ID.
+
+### Syntax
+
+The `account_id` is a case-sensitive string in the form
+
+```
+account_id:        nft_id + "@" + chain_id
+nft_id:            contract_address + ":" + token_id
+contract_address:  [a-zA-Z0-9]{1,63}
+token_id:          [0-9]{1,78}
+chain_id:          [:-a-zA-Z0-9]{5,64}
+```
+
+### Semantics
+
+The `contract_address` is a case sensitive string which its format is specific to the blockchain that is referred to by the `chain_id`. It represents the address of the token contract of the given NFT type. This is very similar to `account_address` in [CAIP-10](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-10.md)
+The `token_id` represents an integer (`uint256` on ethereum) that specifies the specific NFT token in the token contract.
+The `chain_id` is specified by the [CAIP-2](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-2.md) which describes the blockchain id.
+
+## Rationale
+
+The goals of the general account ID format is:
+ - Uniqueness between chains regardless if they are mainnet or testnet
+ - Readibility using `@` to easily identify the NFT identifier
+ - Restricted to constrained set of characters and length for parsing
+ 
+## Test Cases
+
+This is a list of manually composed examples
+
+```
+# Ethereum mainnet
+0xfac7bea255a6990f749363002136af6556b31e04:43628454634268938339309640558483935353102684380863414629699234195440168867178@eip155:1
+```
+
+## Links
+
+n/a
+
+## Copyright
+
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
--- a/CAIPs/caip-18.md
+++ b/CAIPs/caip-18.md
@@ -7,7 +7,7 @@ status: Draft
 type: Standard
 created: 2020-04-23
 updated: 2020-04-23
-requires: 2
+requires: 10
 ---
 
 ## Simple Summary
@@ -16,7 +16,7 @@ CAIP-18 defines a way to identify a specific NFT token on any blockchain specifi
 
 ## Abstract
 
-This proposal aims to facilitate specifying NFT tokens on any blockchain extending CAIP-2 blockchain id specification. This is useful for both decentralized applications and wallets to specify NFT tokens on multiple chains using string identifiers specific to each chain. Currently wallets are usually designed for each chain and multi-chain wallets use proprietray data structures to differentiate NFTs. This proposal aims to standardize these identifiers for NFTs to allow interoperability.
+This proposal aims to facilitate specifying NFT tokens on any blockchain extending CAIP-2 blockchain id specification. This is useful for both decentralized applications and wallets to specify NFT tokens on multiple chains using string identifiers specific to each chain. Currently wallets are usually designed for each chain and multi-chain wallets use proprietary data structures to differentiate NFTs. This proposal aims to standardize these identifiers for NFTs to allow interoperability.
 
 ## Motivation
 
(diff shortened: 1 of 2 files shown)


CAIPs/caip-18.md
```diff
@@ -0,0 +1,69 @@
...
+
+The `account_id` is a case-sensitive string in the form
+
+```
+account_id:        nft_id + "@" + chain_id
+nft_id:            contract_address + ":" + token_id
+contract_address:  [a-zA-Z0-9]{1,63}
+token_id:          [0-9]{1,78}
+chain_id:          [:-a-zA-Z0-9]{5,64}
+```
```
pedrouid, line 39: Let's not use it as account_id since it's already used for CAIP-10, I would change this to be called `nft_id` instead. Additionally this is very similar to `account_id` from CAIP-10 so I would change the order of `contract_address` and `token_id` to make use of the existing standard. 

```
nft_id:          token_id + ":" + account_id
account_id:      [@:-a-zA-Z0-9]{6,128}
token_id:        [0-9]{1,78}
```

CAIPs/caip-18.md
```diff
@@ -0,0 +1,69 @@
+---
+caip: 18
+title: NFT ID Specification
+author: Joel Thorstensson (@oed)
+discussions-to: https://github.com/ChainAgnostic/CAIPs/pull/18
+status: Draft
+type: Standard
+created: 2020-04-23
+updated: 2020-04-23
+requires: 2
```
pedrouid, line 10: requires: 10

pedrouid:
Any comments? cc @ligi @webmaster128

pedrouid:
I would love your opinion on this @okwme since you've worked on NFT's on both Ethereum and Cosmos

CAIPs/caip-18.md
oed: Cool, yeah that makes sense to me 👍

README.md
```diff
@@ -29,3 +29,5 @@ If your CAIP requires images, the image files should be included in a subdirecto
 * **CAIP-6** - [Blockchain Reference for the LIP9 Namespace](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-6.md)
 * **CAIP-10** - [Account ID Specification](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-10.md)
 * **CAIP-13** - [Blockchain Reference for the Polkadot Namespace](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-13.md)
+* **CAIP-18** - [NFT ID Specification](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-18.md)
+
```
ligi, line 33: please remove the new line at the end

CAIPs/caip-18.md
```diff
@@ -0,0 +1,69 @@
...
+
+The `contract_address` is a case sensitive string which its format is specific to the blockchain that is referred to by the `chain_id`. It represents the address of the token contract of the given NFT type. This is very similar to `account_address` in [CAIP-10](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-10.md)
+The `token_id` represents an integer (`uint256` on ethereum) that specifies the specific NFT token in the token contract.
+The `chain_id` is specified by the [CAIP-2](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-2.md) which describes the blockchain id.
+
+## Rationale
+
+The goals of the general account ID format is:
+ - Uniqueness between chains regardless if they are mainnet or testnet
+ - Readibility using `@` to easily identify the NFT identifier
```
ligi, line 51:
```suggestion
 - Readability using `@` to easily identify the NFT identifier
```

CAIPs/caip-18.md
```diff
@@ -0,0 +1,69 @@
...
+requires: 2
+---
+
+## Simple Summary
+
+CAIP-18 defines a way to identify a specific NFT token on any blockchain specified by CAIP-2 blockchain id.
+
+## Abstract
+
+This proposal aims to facilitate specifying NFT tokens on any blockchain extending CAIP-2 blockchain id specification. This is useful for both decentralized applications and wallets to specify NFT tokens on multiple chains using string identifiers specific to each chain. Currently wallets are usually designed for each chain and multi-chain wallets use proprietray data structures to differentiate NFTs. This proposal aims to standardize these identifiers for NFTs to allow interoperability.
```
ligi, line 19:
```suggestion
This proposal aims to facilitate specifying NFT tokens on any blockchain extending CAIP-2 blockchain id specification. This is useful for both decentralized applications and wallets to specify NFT tokens on multiple chains using string identifiers specific to each chain. Currently wallets are usually designed for each chain and multi-chain wallets use proprietary data structures to differentiate NFTs. This proposal aims to standardize these identifiers for NFTs to allow interoperability.
```

oed:
Thanks @pedrouid and @ligi. Updated according to your suggestions!

CAIPs/caip-18.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Motivation
+
+The motivation for proposal stem from designing a chain-agnostic protocol for communication between dapps and wallets that was independent of any blockchain but provide the flexibility to be backwards compatible with existing applications.
+
+## Specification
+
+The NFT id specification will be prefixed with the NFT token ID and the NFT contract address and delimited with an at sign (`@`) followed by a CAIP-2 blockchain ID. Since [CAIP-10](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-10.md) can be used to reference a specific account (which includes contract addresses) we can represent the NFT id as the token specific id concatinated with the CAIP-10 account id.
```
antoineherzog, line 27: For a Cosmos NFT, I think there is no NFT contract address ( because no smart contract, but only natives modules https://github.com/cosmos/modules/tree/master/incubator/nft and just transactions types https://github.com/cosmos/modules/blob/master/incubator/nft/docs/spec/03_messages.md).

oed: Thanks for providing insight @ntoineherzog! So what do you need to uniquely identify an NFT on cosmos? Just the ID? Or do you also need the Denom?

antoineherzog: One more thing, this spec is still draft I think and I believe things will get clarifies after Games of Zones when we actually have NFT moving from chain A to chain B. To give some context, at IOV, we don't implement yet the NFT spec because we are waiting for to see how things will mature. So this spec is very early stage.

I think you are mixing the concept of address and smart contract. In the ethereum world, an address can be smart contract so it makes sense to mix the concept in your current CAIP for ethereum but it makes no sense for native chains like cosmos.

Also I would propose something  more like an URL because you can think it like a path

So you should start with the CAIP2
so something like with the chain_ID
eip155:1/type_of_asset

type_of_asset could be ERC2O token or NFT, for ethereum it is going to be a smart contract address but for cosmos it is going to be the TICKER or the name of the NFT collection

then for NFT, you can add later the ID of the NFT so something like that at the end:

eip155:1/type_of_asset/ID_of the NFT

example for DAI

eip155:1/0x6b175474e89094c44da98b954eedeac495271d0f

example for ATOM

cosmos:comoshub-3/ATOM

example for crypto-kitties NFT with ID 1

eip155:1/0x06012c8cf97BEaD5deAe237070F9587f8E7A266d/1

I hope it helps!

antoineherzog: what do you think @pedrouid ?

okwme: i have no preference for the order of ids but agree with @antoineherzog that `denom` should be the stand in for contract address on cosmos-like chains. `denom` does not change per asset, although there is currently no enforced character or size limit on the string which makes up denom. This is the best draft of an Interchain Standard for NFT transfers to date: https://github.com/cosmos/ics/pull/294/files. It seems to follow the spec implemented for fungible token transfers here: https://github.com/cosmos/ics/tree/master/spec/ics-020-fungible-token-transfer. Neither seem to have restrictions on denom strings.

pedrouid:
Thanks a lot for your input 🙌 @antoineherzog @okwme 

It seems to me that an asset_id standard is better suited for this where we can define namespaces for each asset type

Similarly to how chainId used different namespaces for different blockchain references

We could perhaps close this PR and kickstart a new one where we could start a new CAIP where it defines

asset_id = chain_id + “/“ + asset_namespace + asset_reference

Consequently an NFT on Ethereum would be defined with a different namespace than an NFT on Cosmos

oed:
`asset_id = {chain_id}/{asset_namespace}/{asset_reference}` makes sense to me. My only question is if it will be possible to detect if the asset is an NFT from the CAIP encoding. This is critical for our use case.

pedrouid:
well technically you would parse the string for the asset_namespace

we could define it as for example

```
eip155:1/erc721/0x06012c8cf97BEaD5deAe237070F9587f8E7A266d:1
```

okwme:
That seems like a good solution @pedrouid although it also brings up the question about what kind of distinctions should be made between assets? just fungible and non-fungible? what about semi-fungible? is that a non-problem or are there enough cases of other asset types that it should have careful planning? or should that type of organization take place a level higher? like one registry just for fungible an done for non-fungible, so that within that context you already know some of the attributes of the asset?

pedrouid:
Well that would be defined in a different namespace technically because a fungible token on ethereum would be erc20 namespace then a non-fungible token on ethereum would be erc721 namespace and a semi-fungible token would be a new standard with its own namespace

antoineherzog:
To be honest, I would not do any distinction with the type of asset it is in the URL. We could have many Type of assets. You have erc20, erc721, then tomorrow more typesetc.. The client will have to look the collection_id and then do a query somewhere or on the smart-contract ifself.

I have to know first the collection_id URL
collection_id = {chain_id}/{asset_namespace}

If you want to know what kind of asset is it, then you will do another query somewhere to know with the collection_id and you will query a database to give you the type
Also the type could be encoded into the collection_id or smart_contract itself.

ligi:
closing via #19 - thanks for your work here and there @oed !

PR #19, opened 2020-06-09
add CAIP For Asset Type and Asset ID

romeo4934:


pedrouid:
I think this CAIP relates well with CAIP #18

The need for an Asset ID schema is clear and we could capture both use-cases with a more generalized format IMO

If specified a asset namescape we could include SLIP44 and NFT assets on Ethereum

Example
```
// Ethereum NFT
eip155:1/erc721/0x06012c8cf97BEaD5deAe237070F9587f8E7A266d:1

// Ethereum ERC20 
eip155:1/erc20/0x6b175474e89094c44da98b954eedeac495271d0f

# Ether of Ethereum mainnet
eip155:1/slip44/60
# Bitcoin of Bitcoin mainnet
bip122:000000000019d6689c085ae165831e93/slip44/0
# ATOM of Cosmos Hub
cosmos:cosmoshub-3/slip44/118
```

RFC @ligi @oed @okwme @webmaster128

webmaster128:
I also don't think that the SLIP-0044 coin index is sufficient to capture the tokens needed.

As Pedro mentioned, a good commonly known example could be DAI that you typically hold with an ETH keypair. I wonder if contract addresses are what we want here. Maybe yes, but very technical an tokens move to addresses sometimes.

oed:
**Contract interface**, **Contract Address**, and **parameters** is really what we need here. Seems like the first two options in @pedrouid proposal covers this. However I think using both contract interface and a registry in the same example doens't make sense @pedrouid (interface e.g. *erc20*, and registry *slip44*).

antoineherzog:
Looks good for me.

If it is fine for you, I would break this CAIP in 3 distincts CAIP like we did before with generic CAIP-2 specification and namespace specific spec for CAIP-3, CAIP-4, CAIP-5, CAIP-6, and CAIP-13.

1) I would create a CAIP-19 which define the general framework to specify an asset ID which is:

asset_id: chain_id + "/" + asset_namespace + "/" + asset_reference
 
2) I would create a CAIP-20 which define the namespace slip44. This namespace is chain-agnostic

3) I would create a CAIP-21 which define the namespace erc20, erc721 which exist only on the chain-id eip155

What do you think?

pedrouid:
Agreed. Defining namespaces for assets is flexible for us to have a consistent assetId schema and allows us to define more detailed schemas for each namespace

This was the pattern adopted for CAIP-2 and it worked pretty well. I would like to have a little more eyes on this but I like the strategy suggested for developing the following CAIPs

oed:
I'm quite concerned with having a **namespace** that includes both **slip44** as well as **ercXXX**. These are two very different things.

* **slip44** is a registry of assets
* **ercXXX** defines contract interfaces

Overloading the **namespace** to refer to two very different things is very confusing and will likely cause issues in the future.

oed:
Fair enough. So each namespace would in this case have to define what `asset_reference` is? For erc20 that's just an address, for erc721 it's an address + token id.

Also for ethereum addresses we need to decide if they are all lowercase or if they need to include the checksum.

pedrouid:
We can expand on the ethereum addresses for the specific namespace CAIP but IMO it should inherit the same standards for Ethereum (requiring to be either checksum or lowercase)

antoineherzog:
Ok I updated CAIP19 and CAIP20 accordingly! @pedrouid @oed @webmaster128 
Also I am going to work on CAIP21 (erc20 and erc721 asset namespace this week) However regarding for asset namespace ERC20 and ERC721, are we sure we want to go this route?

I just want to hightlight the fact that these namespace (ERC20 and ERC721) are not chain-agnostics. Also ERC20 is not the only token standard, there is ERC777 and maybe others standards will come.

For smart-contract chains, I would prefer to use a more chain-agnostic approach: We could define a namespace: CA (for contract account) that could be used by any smart contract chains to reference an asset (fungible or not fungible)like: Ethereum, EOS, CosmWasm etc..

like

for DAI
eip155:1/ca/0x6b175474e89094c44da98b954eedeac495271d0f

What do you think?

oed:
@antoineherzog I'd prefer to have the standard specific namespaces. This will give the wallet more metadata about the asset directly from the identifier without having to look things up on-chain. It also makes the identifier human readable.

rekmarks:
As a wallet developer, I second @oed’s point about asset namespaces. Keeping the standard in the identifier itself is a significant improvement for wallets. The less network calls we have to make, the better.

pedrouid:
Let's go with the standard specific namespace.. it worked well for ChainID and I think it fits perfectly for AssetID too

For example... ERC20 and BEP2 tokens would have different Token interfaces so it would make sense to differentiate the two even if they are on different chain namespaces. Just like ERC20 and ERC721 would have different interfaces but they share the same chain namespace

antoineherzog:
Got it! Will work on it tomorow morning :)

antoineherzog:
@pedrouid @oed @ligi Ok I add erc20 and erc721, please fell free to give your feedback :)
Super excited about this spec!

pedrouid:
This is great, I would also add @oed as an author `Joel Thorstensson (@oed)` and finally add the tokenId to the ERC721 spec

oed:
Looks great, but need to add tokenId for ERC721 as @pedrouid mentioned. Would also be cool if there was an example of a non-fungible token on cosmos!

pedrouid:
> Would also be cool if there was an example of a non-fungible token on cosmos!
cc @okwme

okwme:
The current NFT module is at https://github.com/cosmos/modules but i’m working on finding a sustainable way to get it maintained. There’s a demo app at https://github.com/okwme/cosmos-nft using the current implementation if useful

antoineherzog:
@pedrouid @oed Please have a look, I updated the PR following your feedbacks

CAIPs/caip-22.md
```diff
@@ -0,0 +1,62 @@
...
+
+Not applicable
+
+## Test Cases
+
+This is a list of manually composed examples
+
+```
+# CryptoKitties Collectible
+eip155:1/erc721:0x06012c8cf97BEaD5deAe237070F9587f8E7A266d
```
oed, line 54: What's the purpose of having a way to refer to an ERC721 contract?

oed (commented):
Looks good overall, have a question on the erc721 caip.

CAIPs/caip-22.md
antoineherzog: Exchanges and Wallet usually needs to give a human readable name or abreviation  for different type of assets.

 It can be a ticker for fungible like
XBT for bitcoin -> bip122:000000000019d6689c085ae165831e93/slip44:0 

or giving a name for a type of collectible. It can be quite usefull for signed contract between 2 parties for example.

cryptokitties -> eip155:1/erc721:0x06012c8cf97BEaD5deAe237070F9587f8E7A266d

We could imagine a universal asset registry which would register asset type name.  Then no need for each exchange, each wallet to redo this painfull work all the time again.

oed: Makes sense, ty!

antoineherzog:
@pedrouid @ligi Any other feedback?

pedrouid:
Just needs updating the CAIP Index on the README.md and then it's good to merge 👍 cc @ligi

antoineherzog:
@pedrouid @ligi Readme fixed!

CAIPs/caip-19.md
```diff
@@ -0,0 +1,117 @@
...
+
+Often you need to reference the asset type or the asset ID. For example when you want to do an atomic swap between a fungible asset and a non fungible asset.
+
+## Motivation
+
+Currently, each wallet or each exchange needs to create their own registry of type of assets and their associated metadata for example like [Trust-Wallet](https://github.com/trustwallet/assets/tree/master/blockchains) or [CoinMarketCap](https://coinmarketcap.com/). Providing a unique type of Asset and an Asset ID for each asset for developers can reduce the risk of confusion betweens different assets. 
+
+## Specification of Asset Type
+
+The Asset Type is a string designed to uniquely identify type of assets in a developer-friendly fashion.
```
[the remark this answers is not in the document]
ligi: also possible "types of assets"

CAIPs/caip-19.md
```diff
@@ -0,0 +1,117 @@
...
+```
+asset_type:    chain_id + "/" + asset_namespace + ":" + asset_reference
+chain_id:          Blockchain ID Specification cf. CAIP2 
+asset_namespace:   [-a-z0-9]{3,16}
+asset_reference:   [-a-zA-Z0-9]{1,47}
+```
+
+## Specification of Asset ID
+
+The Asset ID is a string designed to uniquely identify a non fungible asset in a developer-friendly fashion.
```
ligi, line 42: Do we really want to limit it to "non fungible"? Also if ERC-20 is an asset we want to cover - isn't ERC-20 fungible?

CAIPs/caip-20.md
```diff
@@ -0,0 +1,72 @@
+---
+caip: 20
+title: Asset Reference for the SLIP44 Asset Namespace
+author: Antoine Herzog (@antoineherzog), Pedro Gomes (@pedrouid), Joel Thorstensson (@oed)
+discussions-to: https://github.com/ChainAgnostic/CAIPs/pull/19
+status: Draft
+type: Standard
+created: 2020-06-23
+updated: 2020-06-23
+requires: 10
```
ligi, line 10: +requires 19

CAIPs/caip-21.md
```diff
@@ -0,0 +1,58 @@
+---
+caip: 21
+title: Asset Reference for the ERC20 Asset Namespace
+author: Antoine Herzog (@antoineherzog), Pedro Gomes (@pedrouid), Joel Thorstensson (@oed)
+discussions-to: https://github.com/ChainAgnostic/CAIPs/pull/19
+status: Draft
+type: Standard
+created: 2020-06-23
+updated: 2020-06-23
+requires: 10
```
ligi, line 10: +19

CAIPs/caip-21.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Motivation
+
+See CAIP-19.
+
+## Specification
+
+### ERC20 Asset Namespace
+
+The asset namespace is called "erc20" as in [ERC20](https://eips.ethereum.org/EIPS/eip-20). It reference erc20 tokens in the eip155 namespace (cf. CAIP3).
```
ligi, line 30: I would rather use "see" instead of "cf."

CAIPs/caip-22.md
```diff
@@ -0,0 +1,62 @@
+---
+caip: 22
+title: Asset Reference for the ERC721 Asset Namespace
+author: Antoine Herzog (@antoineherzog), Pedro Gomes (@pedrouid), Joel Thorstensson (@oed)
+discussions-to: https://github.com/ChainAgnostic/CAIPs/pull/19
+status: Draft
+type: Standard
+created: 2020-06-23
+updated: 2020-06-23
+requires: 10
```
ligi, line 10: +19

ligi (changes requested):
Good overall - just some nits

CAIPs/caip-19.md
antoineherzog: We can define an Asset Type for non fungible token and also fungible token: for example, DAI or CRYPTOKITTIES
When it comes to Asset ID, I believe we can give an ID only to non fungible token because for fungible token, they are fungible token, so we cannot attribute a specific ID for each coin?

antoineherzog:
@ligi
Hi Ligi,
I addresses all your feedback in my last commit except for your question and answer your question. 
Please let me know what do you think?

pedrouid (approved):

ligi (approved):

ligi:
Great work @antoineherzog @pedrouid @oed !

PR #23, opened 2020-08-27
CAIP-23: Blockchain Reference for Filecoin Namespace

ukstv:
This pull request adds CAIP-2 chain id for Filecoin.

### DIFF
--- a/CAIPs/caip-draft_filecoin_namespace.md
+++ b/CAIPs/caip-draft_filecoin_namespace.md
@@ -1,0 +1,62 @@
+---
+caip: <to be assigned>
+title: Blockchain Reference for the LIP9 Namespace
+author: Sergey Ukustov (@ukstv)
+discussions-to: TBD
+status: Draft
+type: Standard
+created: 2020-08-26
+requires: 2
+---
+
+## Simple Summary
+
+This document is about the details of the Filecoin network namespace and reference for CAIP-2.
+
+## Abstract
+In CAIP-2 a general blockchain identification scheme is defined. This is the
+implementation of CAIP-2 for Filecoin network.
+
+## Motivation
+See CAIP-2.
+
+## Specification
+
+### Filecoin Namespace
+
+The namespace "fil" refers to the wider Filecoin ecosystem.
+
+#### Reference Definition
+
+The reference relies on Filecoin's current designation of addresses belonging to test or main networks by prefixing them
+with `t` or `f` correspondingly.
+
+Reference could only be populated with `f` or `t` symbols.
+
+## Rationale
+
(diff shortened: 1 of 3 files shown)


oed (approved):

CAIPs/caip-draft_filecoin_namespace.md
```diff
@@ -0,0 +1,62 @@
+---
+caip: <to be assigned>
```
ligi, line 2: please set to 23 - also the filename

ligi (changes requested):
DIP number is 23 - please set header and filename accordingly

ligi:
Thanks for the PR! Does Filecoin have a concept similar to chainID's like in EIP155? Just having `t` and `f` feels lacking upgrade paths a bit. Would love some other opinions before merging.

vasa-develop:
cc @arajasek

oed:
Some context from the filecoin slack:

---

**Aayush Rajasekaran Aug 28th at 4:09 AM**
Hey @Joel! I may be missing some context, but this looks broadly good to me (testnet addresses will start t, and mainnet ones will start with f).

**Joel  3 days ago**
Thanks @Aayush Rajasekaran Context: CAIP-2 is a standard to represent any blockchain system by having a namespace and a chainId. There are currently CAIP-2 specs for a number of popular blockchains.So if I understand correctly filecoin adds a chain identifier to each address on the blockchain which tell users which chain the address is associated with. Currently you have f and t, so that seems appropriate here as you said. Is there any plan for how to deal with potential forks or new networks based on filecoin in the future?

**Aayush Rajasekaran  3 days ago**
No plan that I’m aware of, but maybe good to think about now…I’ll poke the team. For now, though, yeah f for mainnet addresses sounds good to me.Do you typically include testnet stuff in the spec? I don’t think there’s anything wrong with that, just wondering whether we want to specify the t addresses at all.An additional note is that you could include more info in this spec if you chose. Filecoin has (loosely speaking) 4 kinds of addresses:

    ID addresses, which start f0
    Secp addresses, which start f1
    Multisig addresses, which start f2
    BLS addresses, which start f3

I’m not sure whether you wanna include that detail (it’s a little more vulnerable to change than just the f prefix).

oed (changes requested):
Can you also add CAIP-23 to the list in README.md?

ukstv:
@oed There. Links to master branch. The link would be all right after merge.

ligi (approved):

oed (approved):

oed:
Ping @ligi, anything blocking a merge here?
Also, welcome back to Berlin :)

ligi:
Sorry for the late merge - what a wild year ..

PR #25, opened 2020-10-14
Chain Agnostic Provider Handshake

pedrouid:
As a result of the discussion in #24 I've created a new proposal focused on chain agnostic provider authentication

### DIFF
--- a/CAIPs/caip-25.md
+++ b/CAIPs/caip-25.md
@@ -1,0 +1,111 @@
+---
+caip: 25
+title: Chain Agnostic Provider Authentication
+author: Pedro Gomes (@pedrouid)
+discussions-to: https://github.com/ChainAgnostic/CAIPs/pull/25
+status: Draft
+type: Standard
+created: 2020-10-14
+updated: 2020-10-14
+requires: 2, 10
+---
+
+## Simple Summary
+
+CAIP-25 defines an authentication procedure for a chain agnostic provider to authenticate a session
+
+## Abstract
+
+This proposal has the goal to define a standard procedure for decentralized application to interface with cryptocurrency wallets with multiple chains and defining a set of rules to be followed during a session managed by a provider.
+
+## Motivation
+
+The motivation comes from the lack of standardization across blockchains to expose accounts and define the expected JSON-RPC methods to be used by an application through a provider connecting to a wallet.
+
+## Specification
+
+The provider is defined within a session once the authentication procedure is successfully approved by a wallet and is considered to be over once the wallet chooses to terminate the session by sending a corresponding event labelled as `disconnect`
+
+### Request
+
+The application would interface with a provider to initiate a session by calling the following JSON-RPC request
+
+```
+{
+    "id": 1,
+    "jsonrpc": "2.0",
+    "method": "caip_authenticate",
+    "params": {
+        "chains": ["eip155:1"],
+        "methods": ["eth_sendTransaction", "eth_signTransaction", "eth_sign", "personal_sign"]
+    }
+}
+```
+
+The JSON-RPC method is labelled as `caip_authenticate` and expects two parameters:
+
+* chains - array of CAIP-2 complaint chainId's
+* methods - array of JSON-RPC methods expected to be used during the session
+
+### Response
+
+The wallet can respond to this method with either a success result or an error message.
+
+#### Success
+
+The response MUST be a success result when the user approved accounts matching the requested chains to be exposed and the requested methods to be used.
+
+An example of a successful response should match the following format:
+
+```jsonc
+{
+    "id": 1,
+    "jsonrpc": "2.0",
+    "result": {
+        "accounts": ["0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb@eip155:1"],
+    }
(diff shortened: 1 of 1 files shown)


ligi (approved):
<editor_mode>
Thanks for the PR. Formally it is good - so I approve it and am happy to merge.
</editor_mode>
<user_mode>
Not really happy about the wording "authenticate". Wikipedia says about authentication:

> is the act of proving an assertion, such as the identity of a computer system user.

and I do not see any proving of assertion here. Maybe something like handshake could be a better name here?

Also maybe we should not reduce the answer to yes or no. An alternative I imagine could be that the client requests a set of chains and functions and the response could be a subset of functions/chains and accounts. (just thinking out loud here as I think there could be cases where functions or chains could be optional)

</user_mode>

pedrouid:
> Not really happy about the wording "authenticate".

Agreed... I wasn't also very happy with the wording "authenticate" since there isn't an actual signature being returned either to verify account ownership or to authenticate the set of methods requested.

Suggested replacements: andshake, disclosure, connection, agreement, initialization

> Also maybe we should not reduce the answer to yes or no. An alternative I imagine could be that the client requests a set of chains and functions and the response could be a subset of functions/chains and accounts. 

Technically there are multiple error messages to describe the reason for rejecting this method. However my reasoning for not allowing a partial approval is to provide a clear intent of the dapp requirements

In theory a dapp would strictly need these methods to provide complete functionality of its application.

However we could provide a partial approval with the potential of requesting methods later on with a second method. Im not a huge fan of this approach but Im open for discussing it in more detail

ligi:
> Suggested replacements: andshake, disclosure, connection, agreement, initialization

I think we should go with handshake as stated in the first comment

> In theory a dapp would strictly need these methods to provide complete functionality of its application.

> However we could provide a partial approval with the potential of requesting methods later on with a second method. Im not a huge fan of this approach but Im open for discussing it in more detail

ack - let's do this in a subsequent CAIP to keep things simple in the beginning - should also cover most paths

let's just say all chains and methods here are mandatory. Later on we can e.g. add the parameters optional_chains and optional_methods so dApps can handshake on and probe for optional chains and methods

pedrouid:
Alright I will rename all "authentication" mentions on this standard proposal to "handshake"

ligi (approved):

dominicletz:
Awesome! This is great and totally relates to https://github.com/MetaMask/metamask-extension/issues/1820 -- Didn't even realize this exists when I posted my `wallet_requestChainId` [proposal](https://github.com/MetaMask/metamask-extension/issues/1820#issuecomment-712898431) over there.

Question: Should others APIs that are not specified but supported by the chain still be callable, blocked, or should this decision be up to the wallet implementation?

E.g. when they specifying `["eth_sendTransaction", "eth_signTransaction", "eth_sign", "personal_sign"]` would that block `eth_getBalance`? 

What do you think of making this choice explicit e.g.
`["eth_sendTransaction", "*"]` would mean `eth_sendTransaction` is required but other calls should still be possible

and then

`["eth_sendTransaction"]` would mean `eth_sendTransaction` is the only call made available by the wallet.

@pedrouid

pedrouid:
Yeah exactly. I want the provider to be explicit about the methods it wants to use. One could add `eth_getBalance` as well but from my experience with WalletConnect I saw providers being developed with a split responsibility around read methods and signing methods. This is also present with ethers v5 design where you have Provider and Signer classes.

That being said I'm not sure if wildcards should be valid unless they are prefixed, eg: `"*"` is invalid but `"eth_*"` is valid. Yet I would like to see methods being very explicit because one of main motivations for this proposal was the lack of support for `eth_signTypedData` so using a wildcard would essentially loose the purpose here

LoesungErde-dotcom:
Top

Dhy2928 (commented):
Good

eatray (approved):

Foxrid5988 (approved):

Uthen242 (approved):

Uthen242:
8ae39e435aeb35802ce3f8db0f25f9fe497fa6ec

Uthen242 (approved):

noemiplatform (approved):

eatray (approved):

eatray (approved):

eatray (approved):

eatray (approved):

eatray (approved):

glaser-blip (approved):

jhonharrisricauerta2-ops (approved):

accesgranet-sketch (approved):
Just check the conditions

0xsarawut (approved):
https://github.com/ChainAgnostic/CAIPs/pull/25

0xsarawut (approved):

eatray (commented):
Hopefully this helps

eatray (commented):
Hopefully it can be enjoyed

0xsarawut (approved):
@0xsarawut CAIPs/caip-25.md

accesgranet-sketch (approved):

blaze27444 (commented):
Approve

PR #29, opened 2020-11-12
add a wildcard case for CAIP 25

romeo4934:


ligi:
just wondering if a wildcard like this could make sense "eip155:*" 
as far as I understand with this change this would not be possible

antoineherzog:
Yes Pedro mention this usecase in the issue. However, I was just thinking that it could be a little bit confusing because it means the character "*" could break regex rules for CAIP:2

antoineherzog:
I can add it no problem

ligi:
I think the * must not match the CAIP:2 rules. It is not part of CAIP-2 in this case. Just think "accepting all eth chains" could be a nice use case. But no strong opinion on this. Waiting for @pedrouid 's input.

pedrouid:
I'm still all debating this personally because on one hand the wildcard will introduce complexity to the CAIP-2 specification but on the other I don't see many better alternatives.

I would love to actually schedule a call to work on something that I think would be extremely beneficial for many blockchain projects. A working group for building a specification for a CAIP provider that could include both CAIP-24, CAIP-25 and other potential proposals to develop something analogous to EIP-1193 spec

ligi:
I am not sure the wildcard should be part of CAIP-2. Maybe part of this CAIP or an CAIP that extends CAIP-2 with wildcard options (then blochain IDs instead of blockchain ID)

LMK when the call will be - ideally for me wednesday would work

antoineherzog:
I am happy to do a call as well! For starname, we need to be able to query wallets and know which chains they support so we can get all their accounts. So we can add a wildcard or we could add another type of query to get the list of chains a wallet support.

antoineherzog:
Should we abandon this idea and create another method to get the list of chain-id that a wallet provider support? @pedrouid @ligi

pedrouid:
Actually I think that's a better idea which is already in line with #24

PR #33, opened 2020-11-30
Implementation of CAIP-2 for EOSIO blockchains

sebastianmontero:
Initial draft of the CAIP-2 for EOSIO blockchains.

### DIFF
--- a/CAIPs/caip-7.md
+++ b/CAIPs/caip-7.md
@@ -32,7 +32,7 @@ The namespace "eosio" refers to the EOSIO open-source blockchain platform.
 #### Reference Definition
 
 The definition is delegated to the [EOSIO Transactions Protocol Documentation](https://developers.eos.io/welcome/v2.0/protocol/transactions_protocol/#32-sign-transaction), the [Chain API Plugin Documentation of the EOSIO Developers Manual](https://developers.eos.io/manuals/eos/latest/nodeos/plugins/chain_api_plugin/api-reference/index?query=chain%20id&page=1#operation/get_info) and the pull request that implemented its generation [Chain ID generation implementation pull request](https://github.com/EOSIO/eos/pull/3425). 
-The Chain ID is the SHA256 hash of the genesis state of the chain, represented as lower case hexadecimal number of 64 digits.
+The Chain ID, as defined by EOSIO, is the SHA256 hash of the genesis state of the chain, represented as lower case hexadecimal number of 64 digits. In order to fit the reference restrictions, the format is a 32 character prefix of the Chain ID.
 
 ## Rationale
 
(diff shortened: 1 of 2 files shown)


ligi (changes requested):
Please also add your CAIP to the list in the README

ligi (approved):
LGTM

ligi:
LGTM - but would love some more eye-ballz on it before the merge  @pedrouid @webmaster128 @rekmarks

CAIPs/caip-7.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Motivation
+
+See CAIP-2.
+
+## Specification
+
+### EOSIO Namespace
+
+The namespace "EOSIO" refers to the EOSIO open-source blockchain platform.
```
webmaster128, line 30:
```suggestion
The namespace "eosio" refers to the EOSIO open-source blockchain platform.
```

CAIPs/caip-7.md
```diff
@@ -0,0 +1,71 @@
...
+
+#### Reference Definition
+
+The definition is delegated to the [EOSIO Transactions Protocol Documentation](https://developers.eos.io/welcome/v2.0/protocol/transactions_protocol/#32-sign-transaction), the [Chain API Plugin Documentation of the EOSIO Developers Manual](https://developers.eos.io/manuals/eos/latest/nodeos/plugins/chain_api_plugin/api-reference/index?query=chain%20id&page=1#operation/get_info) and the pull request that implemented its generation [Chain ID generation implementation pull request](https://github.com/EOSIO/eos/pull/3425). 
+The Chain ID is the SHA256 hash of the genesis state of the chain, and its represented as hexadecimal number of 64 digits
```
webmaster128, line 35:
```suggestion
The Chain ID is the SHA256 hash of the genesis state of the chain, represented as lower case hexadecimal number of 64 digits.
```

webmaster128 (changes requested):
Oh, and this conflicts with https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-2.md#syntax. The reference part must not exceed 47 characters, which is why we use prefixes in https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-4.md

webmaster128:
The alternative to prefixes is to use [a different encoding](https://en.wikipedia.org/wiki/Binary-to-text_encoding#Encoding_standards) for the 32 bytes. CAIP-2 is case-sensitive and you need 32/47 = 68 % efficiency. So candidates are [Base58](https://de.wikipedia.org/wiki/Base58) or [Base62](https://en.wikipedia.org/wiki/Base62) (no idea how this encoding works but promising in theory).

However, I think it is easier for human consumption to use a prefix of the default encoding than to re-encode it.

sebastianmontero:
I've updated the reference definition, to use a 32 character prefix of the Chain ID

CAIPs/caip-7.md
```diff
@@ -32,7 +32,7 @@ The namespace "eosio" refers to the EOSIO open-source blockchain platform.
 #### Reference Definition
 
 The definition is delegated to the [EOSIO Transactions Protocol Documentation](https://developers.eos.io/welcome/v2.0/protocol/transactions_protocol/#32-sign-transaction), the [Chain API Plugin Documentation of the EOSIO Developers Manual](https://developers.eos.io/manuals/eos/latest/nodeos/plugins/chain_api_plugin/api-reference/index?query=chain%20id&page=1#operation/get_info) and the pull request that implemented its generation [Chain ID generation implementation pull request](https://github.com/EOSIO/eos/pull/3425). 
-The Chain ID is the SHA256 hash of the genesis state of the chain, represented as lower case hexadecimal number of 64 digits.
+The Chain ID, as defined by EOSIO, is the SHA256 hash of the genesis state of the chain, represented as lower case hexadecimal number of 64 digits. In order to fit the reference restrictions, the format is a 32 character prefix of the Chain ID.
```
webmaster128, line 35:
```suggestion
The Chain ID, as defined by EOSIO, is the SHA256 hash of the genesis state of the chain, represented as lower case hexadecimal number of 64 digits. In order to fit the CAIP-2 reference format, a 32 character prefix of the Chain ID is used.
```

webmaster128 (commented):
Some wording 💅 but otherwise 👍

webmaster128 (approved):
Cool and welcome

sebastianmontero:
Great! Thanks!

webmaster128:
@ligi can we merge this? I don't have permissions to do so.

PR #39, opened 2020-12-15
Add updated field to template

webmaster128:
The `updated:` field was set in all documents except CAIP-1 and the template

### DIFF
--- a/CAIPs/caip-1.md
+++ b/CAIPs/caip-1.md
@@ -5,6 +5,7 @@ status: Active
 type: Meta
 author: ligi <ligi@ligi.de>
 created: 2019-08-31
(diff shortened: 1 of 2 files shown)


ligi (approved):

webmaster128:
Ah, we should adapt CAIP as well. Should `updated` really remain optional? I think it is very easy to forget changing this if does not exist in the original document.

webmaster128:
Is this field supposed to contain a list of dates or the last updated date only? IIRC we only used the last updated here so far, but in CAIP-1 it says

> ` * updated:` \<comma separated list of dates\>

ligi (approved):

PR #40, opened 2020-12-15
Make CAIP-25 front matter correct YAML

webmaster128:


caip-template.md
```diff
@@ -1,4 +1,6 @@
 ---
+# Every document starts with a front matter in YAML enclosed by trible dashes.
```
ligi, line 2:
```suggestion
# Every document starts with a front matter in YAML enclosed by triple dashes.
```

webmaster128: Thanks

caip-template.md
```diff
@@ -1,4 +1,6 @@
 ---
+# Every document starts with a front matter in YAML enclosed by triple dashes.
```
ligi, line 2: is this how you comment in the front-matter? Can you point me to some documentation about it?

webmaster128: I'm not 100% sure about this. It is based on
- makes IDE happy
- is valid YAML

I did not find any better docs than the jekyll link I put there. [This npm package](https://www.npmjs.com/package/front-matter) says

> This concept that was originally introduced to me through the jekyll blogging system and is pretty useful where you want to be able to easily add meta-data to content without the need for a database. YAML is extracted from the the top of a file between matching separators of "---" or "= yaml =". It will also extract YAML between a separator and "...".

ligi (approved):

PR #47, opened 2021-03-25
Update error codes for CAIP-25

pedrouid:


ligi (approved):
LGTM - although some context to the change would be nic

pedrouid:
Good point. Here is the reasoning behind this PR:

I was chatting with @rekmarks about JSON-RPC error codes regarding EIP-1193 and I've realized that I had misunderstanding about the requirements of the range it was required to be JSON-RPC 2.0 compatible.

This prompted me to change the codes to be positive integers and it also taking into consideration existing error codes on Ethereum: https://github.com/MetaMask/eth-rpc-errors/

Finally I've refactored WalletConnect 2.0 to include these as well: https://github.com/WalletConnect-Labs/walletconnect-v2-monorepo/blob/master/packages/utils/src/error.ts

PR #49, opened 2021-06-10
Maximum Lengths for CAIPs

pedrouid:
This proposal updates maximum lengths for CAIP-2, CAIP-10 and CAIP-19

Previous maximum lengths favored the utf-8 string length within standard sizes like 64 and 128 characters however these included the separators. 

This would then create the situation where namespaces, references and addresses had uncommon maximum lengths (47 and 63)  to accommodate for the separators.

Additionally they were unnecessarily long and since now we have quite a few namespace standards specified under CAIPs we can observe that chain namespaces are no longer than 8 characters and chain references are no longer than 32 characters.

Finally these new proposed maximum lengths allow for cleaner byte encoding with fixed lengths without the need of separators

### DIFF
--- a/CAIPs/caip-10.md
+++ b/CAIPs/caip-10.md
@@ -32,8 +32,8 @@ The `account_id` is a case-sensitive string in the form
 
 ```
 account_id:        account_address + "@" + chain_id
-account_address:   [a-zA-Z0-9]{1,63}
-chain_id:          [:-a-zA-Z0-9]{5,64}
+account_address:   [a-zA-Z0-9]{1,64}
+chain_id:          [:-a-zA-Z0-9]{5,41}
 ```
 
 ### Semantics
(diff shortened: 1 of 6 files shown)


webmaster128:
> Finally these new proposed maximum lengths allow for cleaner byte encoding with fixed lengths without the need of separators

How do you do that. They are still variable length, so you cannot simply use e.g. 8+32 = 40 bytes to encode a (namespace, reference) pair.

webmaster128 (commented):
No strong opinion on this change in general. But https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-5.md needs to be carefully reviewed after this change. There we hash everything longer than 47.

pedrouid:
Thanks for the feedback @webmaster128 💪 I've updated CAIP-3, CAIP-5 and CAIP-20 to match the new max length restrictions

webmaster128 (approved):
Looks reasonable to me. But I don't know who is enforcing those rules today and if that breaks people's code.

Would be good to hear from actual users. Carefully approving.

pedrouid:
I would like to have your feedback 👌  @antoineherzog @ligi

ligi:
I have no strong opinion on this one - happy either way.

antoineherzog:
No strong opinion as well. All good for me either way.

pedrouid:
Great 🙌  Looks like we have 4 approvals already which includes authors for all CAIPs. Could we get this merged @ligi ?

reneromeroverdugo1982:
Verdugo

Issue #51, opened 2021-06-23
CAIP-10 Proposed Changes


pedrouid:
CAIP-10 has received a lot of feedback within the last year including some criticisms over its structure. Many Twitter and Discord threads have been written about what would constitute an ideal multi-chain account identifier from which it has also generated new CAIP proposals (#50).

A lot of feedback has also been generated recently in an Ethereum Magicians thread: https://ethereum-magicians.org/t/chain-specific-addresses/6449

Additionally some polls were also created on Twitter:
Poll 1 - https://twitter.com/SchorLukas/status/1404831686714613769
Poll 2 - https://twitter.com/pedrouid/status/1404869512512606218

From my perspective there is essentially a division between 3 main identifiers:
```
A - "0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb@eip155:1" (current CAIP-10)
B - "evm:1:0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb" (updated CAIP-10)
C - "zUJWDxUnc5gJJSWFzxyKkmLp1dgyxiPYT3BrT4" (proposed CAIP-50)
```

From both polls and also from the Ethereum Magicians thread, the majority supports the identifier with format B.

This would mean that it would require the following changes:
1. change of endianness of CAIP-10 and using the same separator as CAIP-2 
```
specific@generic:middle → generic:middle:specific
```
2. update CAIP-2 namespaces to shorter and more recognizable names 
```
bip122 → btc
eip155 → evm
cosmos → cosm
polkadot → dot
```

I propose that we make these changes to respective CAIP standards which are in DRAFT status with two separate PRs for each change and follow up with discussions for each.

lukasschor:
I support these changes and we would be happy to adopt the updated CAIP-10 address format with Gnosis Safe.

ligi:
I am also in favor of B (although I think we should keep eip155  and not evm as evm is to broad and eip155 is where the chainIDs are coming from) - but not really sure we should change CAIP-10 - this might be a bad precedence. I think it is better to create a new one and just let CAIP-10 die.

pedrouid:
My motivation for changing CAIP-10 is mainly because we didn't finalize its status yet besides I wouldn't want to create another CAIP standard that doesn't offer much changes

While CAIP-50 is a more compact and efficient format than CAIP-10 by encoding all accounts in the same base and even introduces a checksum

[...]

PR #56, opened 2021-07-31
Rename namespaces

pedrouid:
Related to #51

### DIFF
--- a/CAIPs/CAIP-29.md
+++ b/CAIPs/CAIP-29.md
@@ -28,7 +28,7 @@ See CAIP-19.
 
 ### ERC1155 Asset Namespace
 
-The asset namespace is called "erc1155" as in [ERC1155](https://eips.ethereum.org/EIPS/eip-1155). It references ERC1155 assets in the eip155 namespace (cf. CAIP-3).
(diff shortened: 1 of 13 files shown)


ligi:
I am lightly against this change - IMHO before it was more specific and I also see no good reason for this breaking change. So I would signal to not do it. Not an "over my dead body" tough - just a light preference. So if the majority wants this change we can merge it - maybe we should do a vote in the CASA call.

bohendo:
FWIW, I'm going to start using this new multi-chain address format in my projects whether or not this PR actually gets merged.

The new format is dope:
 - doesn't leak irrelevant details re Ethereum's governance process
 - uses recognizable namespaces that give devs useful context & won't intimidate users.
 - new endianess means you get more useful info w a quick glance at the first dozen chars of an address
 - new endianess also groups namespaces together when you sort them

pedrouid:
My main motivation for making this change as aggressively as this is:
a) these specs are in DRAFT status
b) starting to see examples in the wild as @bohendo mentioned that use these new namespaces 

But 100% agree @ligi these PRs are in preparation of the CASA meeting for next week

Once we get the voting for these PRs then we should merge them

pedrouid:
Replaced by #59

PR #76, opened 2021-11-01
Hedera Account ID

shemnon:
Describe the Account ID format for Hedera

Signed-off-by: Danno Ferrin <danno.ferrin@hedera.com>


ligi (approved):
:+1:

PR #75, opened 2021-11-01
Hedera Blockchain Reference

shemnon:
Describe the blockchain ID for Hedera

Signed-off-by: Danno Ferrin <danno.ferrin@hedera.com>

PR #77, opened 2021-11-01
Hedera Token Asset Definition

shemnon:
Define the `token` asset references in `hedera` namespace.

Signed-off-by: Danno Ferrin <danno.ferrin@hedera.com>

### DIFF
--- a/CAIPs/caip-77.md
+++ b/CAIPs/caip-77.md
@@ -1,8 +1,8 @@
 ---
-caip: <to be assigned>
+caip: 77
 title: Asset Reference for the Hedera token namespace
 author: Danno Ferrin (@shemnon)
-discussions-to: <URL>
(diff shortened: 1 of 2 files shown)


CAIPs/caip-77.md
```diff
@@ -0,0 +1,69 @@
...
+## Motivation
+
+See CAIP-19.
+
+## Specification
+
+### Token Asset Namespace
+
+The asset namespace is called "token" hosted in the Hedera Token Service (HTS).
+It references HTS tokens in the `hedera` namespace (see CAIP-xx).
```
ligi, line 34: please replace xx with the number - should be 76 right?

ligi (changes requested):

CAIPs/caip-77.md
shemnon: #75 actually, still needs a review.

shemnon: done

PR #80, opened 2021-11-10
Update caip-10.md

daoauth:
update cosmos address use case 

https://github.com/decentralized-identity/did-jwt/pull/205


ligi (commented):
Can you elaborate on the change - bit contrair to what I read in CAIP-5 - also summoning @webmaster128 - maybe you can shed light here

webmaster128:
This change contradicts the overall structure of the document by removing the chain ID as defined in CIP-2.

Also splitting a bech32 address that way is big hackery.

ligi:
@webmaster128 - thanks!

PR #82, opened 2021-11-11
Replace forward slash with semicolon for hierarchical separation

### LINKED ISSUE #81: CAIP-19 cannot be unescaped in URL paths
Referring to: https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-19.md

The following claim is made:

> The following secondary goals can easily be achieved:
>  - Can be used unescaped in URL paths

with e.g. an identifier like

```
eip155:1/slip44:60
```

However, I doubt this claim. If e.g. I was to design a RESTful HTTP API that'd allow users to query for chain-agnostic ERC721 tokens, I'd probably design the endpoint as

```
/api/v1/nfts/:caip-22id:
```

and a user would then fill in the `:caip-22id` portion themselves with e.g. `eip155:1/erc721:0x06012c8cf97BEaD5deAe237070F9587f8E7A266d`, to together it'd look like this:

```
/api/v1/nfts/eip155:1/erc721:0x06012c8cf97BEaD5deAe237070F9587f8E7A266d
```

However, going back to the original claim that CAIP-19 ids can be used unescaped in URL paths, I doubt this is the case as between `/eip155:1` and `erc721:`, a slash interferes with the routing scheme on my API.
I'm not entirely sure how I'd have to escape it, but I'd guess that I'd had to call something like JavaScript's `[encodeURI](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI)` to escape the `/` character.

TimDaub:
Fixes #81


obstropolos:
Hey @TimDaub  - we're going through the backlog here, is this still active / relevant for review?

pedrouid:
Hey @TimDaub, could you join our Discord for CASA? I would love to discuss more in detail about this PR

https://discord.gg/sykcEGpN

CAIPs/caip-19.md
```diff
@@ -31,7 +31,7 @@ The Asset Type is a string designed to uniquely identify the types of assets in
 The `asset_type` is a case-sensitive string in the form
 
 ```
-asset_type:    chain_id + "/" + asset_namespace + ":" + asset_reference
+asset_type:    chain_id + ";" + asset_namespace + ":" + asset_reference
```
oed, line 34:
```suggestion
asset_type:    chain_id + "-" + asset_namespace + ":" + asset_reference
```

Would porefer if we use `-` instead because semicolon might break DIDs in the future:
https://www.w3.org/TR/did-core/#did-url-syntax

silverpill: Both CAIP-2 identifiers and asset namespaces may also contain `-`, so with `-` separators CAIP-19 identifiers will be ambiguous. Example:

CAIP-2: `namespace:chain-1`
Asset namespace: `asset-namespace`
Asset reference: `123`

The resulting CAIP-19 asset type would be `namespace:chain-1-token-namespace:123`. It is not clear how it should be parsed.

octavio12345300 (approved):

PR #94, opened 2022-02-03
Hive Blockchain Reference

stoodkev:


obstropolos:
Doing some cleanup here, we've since moved all namespace requests to the namespaces repo. If this discussion is to continue, please reopen this in the namespaces repo: https://github.com/chainagnostic/namespaces - thanks!

PR #104, opened 2022-03-27
to add Namespaces CAIP (-104) and templates to match

bumblefudge:
based on my experiences overhauling all the namespace-CAIPs.  Note that any changes requested will have to be made in [all 11 of the namespaces already specified](https://github.com/ChainAgnostic/namespaces/pulls) 🙈


oed (approved):
lgtm, should the namespace template live in the namespace repo though?

CAIPs/caip-104.md
```diff
@@ -0,0 +1,229 @@
...
+multi-dimensional than can be efficiently governed over time in the current CAIP
+single-document structure; instead, a Namespace Reference is proposed that
+splits out the application of each cross-chain CAIP to a given namespace as a
+distinct document with its own ratification/supersession lifecycle.  This way,
+long-term specifications for low-level primitives like Chain IDs and addresses
+can be defined once and not be superseded by the addition or modification of
+higher-level specifications for RPC interfaces or multiple asset types.
+
+In practical terms, that means each blockchain ecosystem's namespace is defined
+by a folder containing one document per applicable CAIP.  At a high-level, the
```
ligi, line 24: when reading this I would ask myself "why not just a document if it is *one* document per folder.
Having attended the last call I know there will be more documents in this folder - maybe we should reflect this here to avoid WTF moments in humans reading this.

CAIPs/caip-104.md
```diff
@@ -0,0 +1,229 @@
...
+important facts about a namespace, its governance, and its nomenclature can be
+defined/introduced in a `README.md` file, and each CAIP can be defined in a
+`caip{X}.md` file, where X is the number of the ratified CAIP.
```
ligi, line 27: this already sounds like there will be multiple documents in this folder and contradicts the *one* in the beginning

CAIPs/caip-104.md
```diff
@@ -0,0 +1,229 @@
...
+of a cross-namespace/cross-ecosystem interaction deeply to avoid serious
+security, UX, and design problems. The primary function of defining a namespace
+reference is to map the cross-chain CAIPs against the specifics of a given
+ecosystem and namespace. For this reason, it is important to consider a reader
+with very little context and tease out any potential risk of unfounded
+assumptions about:
+- what a blockchain or DAG is,
+- how transactions get validated against targeted chains or states, 
+- how addressing and routing work across different systems,
+- how a node's state at any given time relates to finality and the authoritative chain,
```
ligi, line 54: I would remove this section - feels like it calls for extra prosa. IMHO readers of this document should know what a blockchain or DAG is. IMHO these documents should really be as concise as possible and not explain such things.

CAIPs/caip-104.md
```diff
@@ -0,0 +1,229 @@
...
+rendered obsolete by a later document; the value is the number of the CAIP that
+replaces the current document. The newer CAIP must have a `replaces` header
+containing the number of the CAIP that it rendered obsolete.
+
+## Auxiliary Files
+
+CAIPs may include auxiliary files such as diagrams. Such files must be named
+CAIP-XXXX-Y.ext, where “XXXX” is the CAIP number, “Y” is a serial number
+(starting at 1), and “ext” is replaced by the actual file extension (e.g.
+“png”).
```
ligi, line 195: this IMHO contradicts the relative assets before. I actually like this one more - maybe we should remove the relative mention before

CAIPs/caip-104.md
bumblefudge: one document **per caip**, meaning many documents.  I reworded it to be clearer, thanks!

CAIPs/caip-104.md
bumblefudge: good catch! done.

PR #111, opened 2022-05-06
added a caip24 proposal for cosmos caip19 asset namespace to make cos…

theoboldfrazier:
This PR would extend the CAIP19 standard for cosmos native and ibc assets. Currently, the CAIP19 is insufficient for dealing with non-native, or "non-base", cosmos assets.

Example: Osmosis caip 19 would be `cosmos:osmosis-1/slip44:118` but for the native ion asset on the osmosis chain, there is no good caip19 for the current standard.

This PR would label the ion asset on the osmosis chain `cosmos:osmosis-1/native:uion`.

Furthermore, it would expand caip19 to handle ibc assets on cosmos network chains. For example, ATOM on the osmosis chain would have a caip19 of `cosmos:osmosis-1/ibc:27394FB092D2ECCD56123C74F36E4C1F926001CEADA9CA97EA622B25F41E5EB2`

### DIFF
--- a/CAIPs/caip-24.md
+++ b/CAIPs/caip-24.md
@@ -1,0 +1,72 @@
+---
+caip: 24
+title: Asset Reference for the Cosmos Asset Namespace
+author: Theobold Frazier
+discussions-to: https://github.com/shapeshift/CAIPs/issues/1
+status: Draft
+type: Standard
+created: 2020-02-14
+updated: 2020-02-14
+requires: 19
+---
+
+## Simple Summary
+
+This document is about the details of the Cosmos network asset namespace and reference for CAIP-19.
+
+## Abstract
+
+In CAIP-19 a general asset identification scheme is defined. This is the
+implementation of CAIP-19 for Cosmos chain native assets and ibc assets.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


cjthompson:
This proposal is sponsored and supported by the [ShapeShift DAO[(https://github.com/shapeshift). We are currently using this implementation in our [library packages](https://github.com/shapeshift/lib).

ligi (changes requested):
Thanks for the PR! We just move all namespace related stuff here: https://github.com/ChainAgnostic/namespaces
Would be great if you could submit it there. Sorry for getting caught in the middle of it - but the move should be easy. That said we should finish the move soon so this does not happen anymore.

cc @bumblefudge

ligi:
> This proposal is sponsored and supported by the [ShapeShift DAO](https://github.com/shapeshift). We are currently using this implementation in our [library packages](https://github.com/shapeshift/lib).

nice! Think ShapeShift DAO should be added here then: https://github.com/ChainAgnostic/CASA#members can you add a PR for that?

bumblefudge:
super cool, guys! could I ask you to check out namespaces/template and [caip104](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-104.md) and rework your PR to fit the model of:
namespaces/cosmos/caip19.md
?  Ping me if you have any questions!

PR #125, opened 2022-07-14
CAIP-19: Add block number tag in query

qizhou:
See discussions in https://github.com/ChainAgnostic/CAIPs/pull/119

### DIFF
--- a/CAIPs/caip-19.md
+++ b/CAIPs/caip-19.md
@@ -31,10 +31,12 @@ The Asset Type is a string designed to uniquely identify the types of assets in 
 The `asset_type` is a case-sensitive string in the form
 
 ```
-asset_type:    chain_id + "/" + asset_namespace + ":" + asset_reference
+asset_type:        chain_id + "/" + asset_namespace + ":" + asset_reference + "?" + query
 chain_id:          Blockchain ID Specification cf. CAIP2
 asset_namespace:   [-a-z0-9]{3,8}
 asset_reference:   [-a-zA-Z0-9]{1,64}
+query:             ("block_number" | "n") "=" block_number
+block_number:      [0-9]{1,78}|"latest"
 ```
 
 ## Specification of Asset ID
@@ -66,6 +68,7 @@ The goals of the general asset type and asset ID format is:
 - To some degree human-readable and helps for basic debugging
 - Restricted in a way that it can be stored on-chain
 - Character set basic enough to display in hardware wallets as part of a transaction content
+- Idempotency and ability to cache the requests to the resource resolvers
 
 The following secondary goals can easily be achieved:
 
@@ -111,6 +114,9 @@ eip155:1/erc721:0x06012c8cf97BEaD5deAe237070F9587f8E7A266d
 
 # CryptoKitties Collectible ID
 eip155:1/erc721:0x06012c8cf97BEaD5deAe237070F9587f8E7A266d/771769
+
+# CryptoKitties Collectible ID at block number 15051263
+eip155:1/erc721:0x06012c8cf97BEaD5deAe237070F9587f8E7A266d/771769?block_number=15051263
 ```
 
 ## Copyright


ntn-x2:
I don't think it's correct to mix an identifier with a querying language. CAIP-19 identifiers should be used to identify assets. Adding a query part to it can be quite confusing.

I think it would make more sense to define a new CAIP which describes the querying language to use. Also because also a chain ID, or an account ID, might want to support this feature, to retrieve historical data.

TimDaub:
> I don't think it's correct to mix an identifier with a querying language.

I agree that it looks confusing syntactically and I wish we had other options. Is there something we could make this less looking like an non-deterministic query @qizhou?

> CAIP-19 identifiers should be used to identify assets.

I think @qizhou and I agree with you. But our problem is that we believe CAIP-19 indeed doesn't identify assets at this point. It is a Universal Resource Location as the identity of the asset can change over time. Hence, to truly create an URI that is permanently accessible, we somehow need to add the block number. Now, I personally have little competence or understanding what should be the right symbol in the case and previously I had suggested the `#` symbol, which as @qizhou correctly pointed out is unfortunately stripped by the web browser (and hence often not sent to a backend). 

> I think it would make more sense to define a new CAIP which describes the querying language to use.

We're not trying this define a query language. This is an extremely important distinction.

qizhou:
Perhaps we could use other symbols other than `#` if the query is not preferred?  E.g., `@`?

ntn-x2:
Not to feather our own's nest, but we have opened an issue in this very same space to integrate CAIP-19 into a new DID method. What that would do, is that it would enhance CAIP-19 with DID features, such as an already standardised query language. In this way, it would be trivial to define a `block_number` query parameter in the DID method to do exactly what @qizhou was proposing here. Maybe you guys want to take a look at that and have the discussion going there? 
The issue in this repo is [here](https://github.com/ChainAgnostic/CAIPs/issues/126), while the new DID method draft is [here](https://github.com/KILTprotocol/specifications/pull/3).

ntn-x2:
Not to feather our own's nest, but we have opened an issue in this very same space to integrate CAIP-19 into a new DID method. What that would do, is that it would enhance CAIP-19 with DID features, such as an already standardised query language. In this way, it would be trivial to define a `block_number` query parameter in the DID method to do exactly what @qizhou was proposing here. Maybe you guys want to take a look at that and have the discussion going there? 
The issue in this repo is [here](https://github.com/ChainAgnostic/CAIPs/issues/126), while the new DID method draft is [here](https://github.com/KILTprotocol/specifications/pull/3).

TimDaub:
> block_number query parameter

block_number is not a query parameter, it's a fragment of an identifier. Anyone that is running CAIP-19 without a block_number is querying as there can be many results to one URL but only one result to an URI.

ntn-x2:
@TimDaub what I am saying (and re-iterating) is that I do not think a block number should be part of the identifier itself. I think it should be some additional detail that might be required in some use cases.
For instance, if we assume verifiable credentials are issued to an CAIP-19 identified asset, the asset would not be identifier by the block number, as otherwise at each block number the identifier would change, which sounds a bit counter-intuitive for me.

Different is the case where the verifier of an asset's credential might want to retrieve the status of the asset at some past block, and it would do that with a query parameter, which would be defined in the DID spec linked above and taken care of by the DID resolver.

I think having the block number as part of the identifier would only introduce problems.

bumblefudge:
I am struggling with the backwards-compatibility part.  Are queries beginning with ? inherently/syntactically optional?  I guess my question is:

Do today's "block-height-free" CAIP-19s default to "block_number=latest"? Is
`:example-namespace:example-network:example-asset` still valid after the PR, and equivalent to
:example-namespace:example-network:example-asset?block_number=latest`
?  

My one concern here is that If there are systems that, for whatever reason, strip the `?...` somehow along the way, they are effectively changing any `?block_number=123` into `?block_number=latest`, which could conceivably introduce problems, or at least a security consideration... systems that support these query suffixes and systems that don't might interact unpredictably or insecurely, non?

TimDaub:
> Are queries beginning with ? inherently/syntactically optional? I guess my question is:

I'm not suggesting backward compatibility in the syntactical sense. CAIP-19 is in status: DRAFT, and so breaking assumptions should be allowed, right? What I mean is that the current implementation of CAIP-19 that doesn't include a block_number would then later be block_number=latest because that's what it is today (albeit implicitly).

> Do today's "block-height-free" CAIP-19s default to "block_number=latest"?

Yes, this is my interpretation.

> My one concern here is that If there are systems that, for whatever reason, strip the `?...` somehow along the way, they are effectively changing any `?block_number=123` into `?block_number=latest`, which could conceivably introduce problems, or at least a security consideration

Fair. 

> Perhaps we could use other symbols other than `#` if the query is not preferred? E.g., `@`?

I liked @qizhou's proposal of using e.g. "@".

ntn-x2:
>Imagine an NFT with a cute kitty and the name "Cute kitty PFP," and then we have this new type of protocol that allows "liking" the NFT kitty using a Verifiable Credential where the issuer sign(CAIP-19WithoutBlock, "I like it") which results in a signature. Now. In this case, since CAIP-19WithoutBlock doesn't contain a block number, the user has signed whatever the latest state was at that point of the chain.

I think this should be a concern of the verifiable credential layer OR of a querying language on top of identifiers, for which historic state can be queried. For instance, either the VC could have a block number as a claim, or an issuance date from which a given block number can be retrieved.

Different is if we want to assume that two CAIP-19 identifiers that only differ in block number are to be considered completely independent on each other: in that case adding a block number to the identifier would make sense.

bumblefudge:
Discussed on today's call; will discuss further at inperson gathering in sept

PR #145, opened 2022-09-30
Add final status

FrederikBolding:
If we want to use a "final" status, which we do for CAIP-2 currently even though it isn't defined in CAIP-1, we must add that status to this yml file for it to show up in Jekyll.


bumblefudge:
will be covered by #158 , thanks for noticing!

PR #152, opened 2022-10-18
Adds new CAIP as superset of CAIP-10

bumblefudge:


oed:
Why can't this be an update to caip10?

oed:
If these changes are backwards compatible we should just update caip10 since it's still in draft stage.

bumblefudge:
> If these changes are backwards compatible

I thought Ceramic's implementation failed validation when `.` or `%` was present? I'm not sure what legacy compatibility issues it would create for other implementers (WC, Spruce, etc.), but this "supersession" approach was the concensus when we talked about it 2 or 3 editorial meetings ago... let's discuss Thursday at the next one!

oed:
We use the js-caip library. I'm sure it can be updated to support this
change?

*Joel Thorstensson *
Tech co-founder,
[image: 3Box Labs] <https://3boxlabs.com>

PR #154, opened 2022-10-19
First draft of CAIP for restricting Web3 provider injection

kdenhartog:
Initial pass at creating a chain agnostic secure context based provider object. This is heavily derived from [EIP-5593](https://github.com/ethereum/EIPs/pull/5593) so it may make sense to make this more CAIP focused. The primary purpose of this is to take the concepts that were initially Ethereum focused from the EIP and abstract them so they're more useful for a multichain approach such as wallets that support Solana or other web3 platform APIs.


CAIPs/caip-tbd.md
```diff
@@ -0,0 +1,102 @@
...
+
+- Top level `http://a.com` -> blocked (insecure/top level)
+- Top level `https://a.com` -> allowed
+- Top level `https://a.com` with `<iframe src="http://a.com/">` -> blocked (insecure first party iframe)
+- Top level `http://a.com` with `<iframe src="https://a.com/">` -> blocked (insecure top level window)
+- Top level `https://a.com` with `<iframe src="https://a.com">` -> allowed
+- Top level `https://a.com` with `<iframe src="https://b.com">` -> blocked (3p iframe without sufficient privileges)
+- Top level `https://b.com` with `<iframe src="http://a.com/">` with `<iframe src="https://b.com">` -> blocked (insecure iframe)
+- Top level `https://b.com` with `<iframe src="https://a.com">` with `<iframe src="https://b.com">` -> blocked (3p iframe without sufficient privileges)
+- Top level `https://a.com` with `<iframe src="https://sub.a.com">` -> allowed
```
kdenhartog, line 59: TODO @kdenhartog: Double check this is actually correct. See https://github.com/ethereum/EIPs/pull/5593#discussion_r997175988

CAIPs/caip-tbd.md
```diff
@@ -0,0 +1,102 @@
...
+- Top level `data://foo with <iframe src="data://bar">` -> blocked (insecure top level scheme)
+- Top level `file://foo with <iframe src="file://bar">` -> blocked (3p iframe)
+- Top level `https://a.com` with `<iframe src="https://b.com" sandbox="allow-same-origin allow-scripts">` -> blocked (3p iframe without sufficient privileges)
+
+### Optional Test Cases
+
+- Top level `https://a.com with <iframe src="https://b.com" allow="ethereum">` -> Ethereum allowed, Solana blocked (3P iframe because "allow" attribute in use for ethereum)
+- Top level `https://a.com with <iframe src="https://b.com" allow="solana">` -> Solana allowed, Ethereum blocked (3P iframe because "allow" attribute in use for solana)
+- Top level `https://a.com with <iframe src="https://b.com" allow="ethereum solana">` -> Solana allowed, Ethereum allowed (3P iframe because "allow" attribute in use for solana and ethereum)
```
kdenhartog, line 70: I'm wondering if it makes sense to make this text more chain agnostic here? E.g. "web3" might make more sense here, but as discussed in https://github.com/w3c/webextensions/issues/271 we may run the risk of extensions colliding on registering this permission in a way that could produce an odd behavior. We'll need to consider the tradeoffs here if we're going to aim for a non unique `allow` attribute.

CAIPs/caip-tbd.md
coryhardman: I really think it would be better for this to not be treated as "allowed". So many sites have different security policies in their subdomains, it would be easy to find situations where one could own sub.a.com, then abuse this policy to get wallet interactions on a.com.

CAIPs/caip-tbd.md
coryhardman: +1 to making it more chain agnostic. 

OR lean into it being chain specific and use the chain ids themselves?

kdenhartog: Yeah that's the part I'm not certain about right now. Brave has taken the chain specific approach right now, but I'd be interested to see what others take is on this. I venture to guess that from a dApps perspective in a truly multichain world this would be a hindrance, but today it would be a nuance of just building something.

CAIPs/caip-tbd.md
kdenhartog: This one has been properly updated now

CAIPs/caip-tbd.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Motivation
+
+Wallets are oftentimes maintaining security and safety of users' funds that can be equivalent to large portions of money. For this reason, it's a good idea to restrict access to the Web3 Provider APIs to align it with other powerful APIs on the web platform. This will assist in reducing the surface area that attacks can be conducted to access users funds or data. Additionally, by adding in restrictions we're reducing the surface area that malicious web pages could fingerprint the user via the Web3 Provider APIs providing some additional privacy benefits. An example of a specific attack that's avoided by this is one where a malicious advertisement is loaded on a legitimate dApp that attempts to interact with a users wallet to maliciously request the user to access funds. With this CAIP implemented the advertisement frame would be considered a 3P iframe and therefore would not have the Web3 Provider API injected into it's sub frame because it's not a secure context.
```
bumblefudge, line 18:
```suggestion
Wallets are oftentimes maintaining security and safety of users' funds that can be equivalent to large portions of money. For this reason, it's a good idea to restrict access to the Web3 Provider APIs to align it with other powerful APIs on the web platform. This will assist in reducing the surface area that attacks can be conducted to access users funds or data. Additionally, by adding in restrictions we're reducing the surface area that malicious web pages could fingerprint the user via the Web3 Provider APIs providing some additional privacy benefits. An example of a specific attack that's avoided by this is one where a malicious advertisement is loaded on a legitimate dApp that attempts to interact with a users wallet to maliciously request the user to access funds. With this CAIP implemented the advertisement frame would be considered a third party ("3P") iframe and therefore would not have the Web3 Provider API injected into it's sub frame because it's not a secure context.
```

CAIPs/caip-tbd.md
```diff
@@ -0,0 +1,102 @@
...
+    - Secure contexts include sites that are served from HTTPS but also HTTP `localhost`. 
+    - The browser implementation may optionally also support configured [potentially trustworthy origins] that would normally not be considered trustworthy if the user configures their browser to do so. See the [Development Environments section of Secure Contexts][w3c-secure-context-dev-env] for additional details.
+- By default the Web3 Provider APIs MUST NOT be exposed to 3p iframes.
+- Web3 provider APIs MUST be `undefined` in an iframe where `window.isSecureContext` returns `false` in that iframe.
+- If the iframe is a third party to the top-level secure origin, it SHOULD be blocked. It MAY be unblocked if the iframe uses the `allow="ethereum"` or `allow="solana"` attribute. In order to achieve this implementers are REQUIRED to extend the Permissions API.
+- If the iframe is first-party to the top-level origin AND the `sandbox attribute is set on the iframe, the provider object MUST be blocked. If the sandbox attribute is set to `sandbox="allow-same-origin"` it MUST be injected for a first party frame.
```
bumblefudge, line 35:
```suggestion
- If the iframe is first-party to the top-level origin AND the `sandbox` attribute is set on the iframe, the provider object MUST be blocked. If the sandbox attribute is set to `sandbox="allow-same-origin"` it MUST be injected for a first party frame.
```

CAIPs/caip-tbd.md
```diff
@@ -0,0 +1,102 @@
...
+
+
+## Backwards Compatibility
+
+Wallet extensions SHOULD consider adding a "developer mode" toggle via a UX so that dApp developers have the capability to disable the insecure context (http) check for the http://localhost:<any-port> origin only in the event that [localhost][w3c-secure-context-localhost] does not return `true` for secure context</a>. This will allow dApp developers to be able to continue to host dApps on the localhost origin if a browser environment has chosen to not already consider localhost a secure context. Most major browser providers do consider localhost a secure context already. This toggle MUST be set to disabled by default.
```
bumblefudge, line 45:
```suggestion
Wallet extensions SHOULD consider adding a "developer mode" toggle via a UX so that dApp developers have the capability to disable the insecure context (http) check for the http://localhost:<any-port> origin only in the event that [localhost][w3c-secure-context-localhost] does not return `true` for secure context</a>. This will allow dApp developers to be able to continue to host dApps on the localhost origin if a browser environment has chosen to not already consider localhost a secure context. Most major browser providers do consider localhost a secure context already. If such a toggle is made available, it MUST be set to disabled by default.
```

CAIPs/caip-tbd.md
bumblefudge: not sure i understand what you mean by "chain agnostic"-- here at CASA we generally distinguish between "namespaces" (like `ethereum` or `solana`) which are VMs+RPC featuresets and "chains" (like ethereum main-net, i.e. `eip155:1`, or solana devnet, i.e. `solana: 8E9rvCKLFQia2Y35HXjjpWzj8weVo44K`) which are specific networks within them.

i guess i'm confused about the namespace of the allow string. the two options are currently `ethereum` and `solana`, because those are the two `window.x` options in today's web3js.  those map to CASA namespaces, perhaps sheerly by dumb luck-- what happens when `polkadot` and `cosmos` show up and their wallets start jockeying for "first permissions" on that namespace? things get complicated when someone wants to use `statemint` and `statemine` or `kusama` as the scope of allowance...

unless you're offering for the CASA namespaces repo to be the definitely registry of namespacing, in which case I'm all for it! 😉

CAIPs/caip-tbd.md
```diff
@@ -0,0 +1,102 @@
...
+
+### Restrictions for providers
+
+The provider objects, e.g. `window.ethereum` or a [Solana wallet adapter][solana-wallet-adapters], are expected to only inject the Web3 provider APIs in secure context when conforming with this standard. The following restrictions are REQUIRED for conformant wallets:
+
+- Provider objects MAY be accessible in private (incognito) windows.
+- The origin MUST be a [potentially trustworthy origin][w3c-secure-context-trustworthy-origin] to have access to the Web3 provider APIs. This can be checked using `window.isSecureContext`, including inside iframes.
+    - Secure contexts include sites that are served from HTTPS but also HTTP `localhost`. 
+    - The browser implementation may optionally also support configured [potentially trustworthy origins] that would normally not be considered trustworthy if the user configures their browser to do so. See the [Development Environments section of Secure Contexts][w3c-secure-context-dev-env] for additional details.
+- By default the Web3 Provider APIs MUST NOT be exposed to 3p iframes.
```
gobengo, line 32:
```suggestion
- By default the Web3 Provider APIs MUST NOT be exposed to third-party iframes.
```

Is this semantically equivalent? Regardless, what does '3p' or even 'third-party' mean exactly? (i.e. different origin? different apex domain?)

darobin: It's probably a good idea to align this with other cross-origin restrictions.

CAIPs/caip-tbd.md
darobin: If this risks expanding into a relatively wide namespace, it might be a good idea to consider prefixing names (eg. `web3-ethereum`). My concern is the potential for conflict with other `allow` values. Feature policies already has about 30 keywords and new ones are added regularly.

CAIPs/caip-tbd.md
kdenhartog: yup those are equivalent @gobengo.

> It's probably a good idea to align this with other cross-origin restrictions.
@darobin yeah that's my thinking here is that this is really focused on when we're dealing with different origins. I was under the impression that third party iframes would imply that meant different origins, but the fact that it's implied makes for imprecise language typically. Do you have suggestions about how to better word this?

CAIPs/caip-tbd.md
kdenhartog: > If this risks expanding into a relatively wide namespace

My hope would be that it doesn't in order to produce better web compatibility between the different wallets, but I'm not certain how to handle the conflict in a way that wouldn't introduce security issues or conflicts in terms of the wallet that's being injected. E.g. it wouldn't be good if the user has two extension based wallets and wallet A injected (via content scripts) first into the main window and wallet B injected first into the iframe.

kdenhartog: I just got a chance to talk with others on our team familiar with the feature and it seems like we'd be good to drop support of this from this spec and if we're able to find a way to get this working with extensions in the future we can extend this spec like @coryhardman suggested or bring it back in if we've not reached final call for it yet. I'll make the updates here and in the EIP

CAIPs/caip-tbd.md
```diff
@@ -0,0 +1,102 @@
...
+- Web3 provider APIs MUST be `undefined` in an iframe where `window.isSecureContext` returns `false` in that iframe.
+- If the iframe is a third party to the top-level secure origin, it SHOULD be blocked. It MAY be unblocked if the iframe uses the `allow="ethereum"` or `allow="solana"` attribute. In order to achieve this implementers are REQUIRED to extend the Permissions API.
```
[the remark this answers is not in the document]
kdenhartog: I'm just realizing I never got this statement removed from this one but I did from the EIP. Will get that updated tomorrow.

CAIPs/caip-tbd.md
```diff
@@ -0,0 +1,96 @@
+---
+caip: TBD
```
kdenhartog, line 2:
```suggestion
caip: 154
```

kdenhartog:
Looks like notes in Editorial meeting in November was good to merge this as draft for now. Can we merge @bumblefudge or @ligi ?

PR #162, opened 2022-10-25
DRAFT: DagJWS CACAO

zachferland:
Create and verify IPLD based JWS payloads using chain-agnostic Object Capabilities and blockchain accounts.

### DIFF
--- a/CAIPs/caip-dagjws_cacao.md
+++ b/CAIPs/caip-dagjws_cacao.md
@@ -28,19 +28,19 @@ Initial practical use cases include the ability to make multiple writes to an au
 
 ### JWS with CACAO Construction
 
-JWS CACAO support includes adding a ‘cap’ parameter to the JWS Protected Header and specifying the correct ‘kid’ parameter string. 
+JWS CACAO support includes adding a `cap` parameter to the JWS Protected Header and specifying the correct `kid` parameter string. 
 
 **“cap” Header Parameter**
 
-The “cap” parameter maps to a URI string. ln the scope here this is expected to be an IPLD CID resolvable to to CACAO object. 
+The `cap` parameter maps to a URI string. ln the scope here this is expected to be an IPLD CID resolvable to a CACAO object. 
 
 **“kid” Header Parameter**
 
-The ‘kid’ parameter references the key used to secure the JWS.  In the scope here this is expected to be a DID with reference to any key in the DID verification methods.  The ‘parameter’ MUST match the ‘aud’ target of the CACAO object for both the CACAO and corresponding signature to be valid together.
+The `kid` parameter references the key used to secure the JWS.  In the scope here this is expected to be a DID with reference to any key in the DID verification methods.  The parameter MUST match the `aud` target of the CACAO object for both the CACAO and corresponding signature to be valid together.
 
 **Protected Header**
 
-With ‘cap’ currently not being a registered header parameter name in the IANA "JSON Web Signature and Encryption Header Parameters" registry, we treat this as a “Private Header Parameter Name” for now with additional meaning provided by the CACAO for implementations that choose to use this specification.  
+With `cap` currently not being a registered header parameter name in the IANA "JSON Web Signature and Encryption Header Parameters" registry, we treat this as a “Private Header Parameter Name” for now with additional meaning provided by the CACAO for implementations that choose to use this specification.  
 
 Example 1: Protected JWS header with CACAO
 ```tsx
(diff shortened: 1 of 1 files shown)


CAIPs/caip-dagjws_cacao.md
```diff
@@ -0,0 +1,84 @@
...
+
+With the building blocks of [did:pkh](https://github.com/w3c-ccg/did-pkh/blob/main/did-pkh-method-draft.md), [IPLD DAG-JOSE](https://ipld.io/specs/codecs/dag-jose/spec/), [CAIP-74](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-74.md), and [CAIP-122](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-122.md) we can describe an additional building block towards building a rich capability-based authorization system on an authenticated DAG based data structure.
+
+Initial practical use cases include the ability to make multiple writes to an authenticated DAG based data structure (IPLD with dag-jose) using a blockchain account by way of a temporary key authorized with a CACAO. Allowing a user to only sign once with blockchain based account and/or wallet and then continue to sign many payloads for an authorized amount of time (session).
+
+## Specification
+
+### JWS with CACAO Construction
+
+JWS CACAO support includes adding a ‘cap’ parameter to the JWS Protected Header and specifying the correct ‘kid’ parameter string.
```
oed, line 31:
```suggestion
JWS CACAO support includes adding a `cap` parameter to the JWS Protected Header and specifying the correct `kid`parameter string. 
```

oed (commented):
Looks good overall.

Minor comments:
- Replace `‘` with ` throughout
- Maybe mention that a DagJWS is really like any JWS but the payload is a base64url encoded CID
- Maybe include an example complete DagJWS with a CACAO?

zachferland:
Fixed formatting, added dagjws payload note, and added full example

CAIPs/caip-dagjws_cacao.md
```diff
@@ -51,23 +51,55 @@ Example 1: Protected JWS header with CACAO
...
+      signatures: [
+        {
+          protected: "eyJhbGciOiJFZERTQSIsImNh...GU2djZEpLTmhYSDl4Rm9rdEFKaXlIQiJ9"
+          signature: "6usTYvu5KN0LFTQsWE9U-tqx...h60EgfvjL_rlAW7_tnQUl84sQyogpkLAQ"
+        }
+      ]
+    }
+  }
+}
+```
```
oed, line 92: Can use the full example with non-abbreviated strings?
So that you can test an implementation against it.

bumblefudge: would it make sense to 
1.) publish a sample/reference IPLD block and pin it for posterity, then
2.) link to it from the spec as an https://{cidv1}.ipfs.dweb.link/ link?
and/or
3.) put the entire IPLD block into an /assets/ folder in this repo and link to that from the spec as well to help people make sure they're fetching/deserializing/etc properly ?

i might be taking this whole `test vector` thing too far but i'm sure the folks at PL outercore would appreciate my asking for this to be extremely n00b-friendly to help with ipfs outreach :D

zachferland: test vectors are always nice, but not sure if makes sense for this spec here

This spec mostly describes how to use a cacao with a JWS, ie adding cacao in header and how to verify, it then also happens to be encoded in dag-jose, i think most of serialization/pathing/etc is entirely covered by dag-jose and would be covered by any tests vectors there and some verification is already covered by cacao spec, thanks for commenting!

CAIPs/caip-dagjws_cacao.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Specification
+
+### JWS with CACAO Construction
+
+JWS CACAO support includes adding a `cap` parameter to the JWS Protected Header and specifying the correct `kid` parameter string. 
+
+**“cap” Header Parameter**
+
+The `cap` parameter maps to a URI string. ln the scope here this is expected to be an IPLD CID resolvable to a CACAO object.
```
bumblefudge, line 35: paragraph me?

chunningham:
Is it possible to adopt more of a "should" and "recommended" approach for the CACAO object? By that I mean that use of a CACAO could be recommended but any authorisation-delegating token is possible. I'm thinking of UCANs, but in principle I think any "appropriate" alternative may be allowed.

Language in this spec which references particular features of CACAO (e.g. comparing the `aud` to the `kid`) can perhaps instead depend on a definition for "appropriate" which requires said features. Alternatively, I understand if CACAO is meant here as that given definition and can expand to include any reasonable candidate for a format.

Overall I think this fills a good niche in both UCAN and CACAO for a common ipld-native invocation token format, which is unspecified so far.

oed:
> Overall I think this fills a good niche in both UCAN and CACAO for a common ipld-native invocation token format, which is unspecified so far.

This is exactly what's intended here :) 
CACAO here is meant as an IPLD representation of any OCAP. There is ongoing work to eventually make CACAO and ucan-ipld equivalent. So eventually a "CACAO" can be a SIWE+ReCap or a UCAN.

The `kid` property has nothing to do with the CACAO and can be used for a DagJWS in general to indicate which DID signed the JWS.

zachferland:
Going to close soon instead of move forward, in favor of a future specification that builds on https://github.com/ChainAgnostic/CAIPs/pull/196 and supports both cacao/ucan and other changes there

mozz30-tech (approved):

PR #167, opened 2022-10-31
CAIP: StreamSync

oed:
A libp2p protocol for synchronizing event streams.

### DIFF
--- a/CAIPs/caip-stream-sync.md
+++ b/CAIPs/caip-stream-sync.md
@@ -1,0 +1,194 @@
+---
+# Every document starts with a front matter in YAML enclosed by triple dashes.
+# See https://jekyllrb.com/docs/front-matter/ to learn more about this concept.
+caip: <to be assigned>
+title: StreamSync
+author: Joel Thorstensson <joel@3box.io>
+discussions-to: <URL>
+status: Draft
+type: Standard
+created: 2022-10-31
+updated: 2022-10-31
+requires: 162
+---
+
+<!--You can leave these HTML comments in your merged EIP and delete the visible duplicate text guides, they will not appear and may be helpful to refer to if you edit it again. This is the suggested template for new EIPs. Note that an EIP number will be assigned by an editor. When opening a pull request to submit your EIP, please use an abbreviated title in the filename, `eip-draft_title_abbrev.md`. The title should be 44 characters or less.-->
+
+## Simple Summary
+<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the CAIP.-->
+A libp2p protocol for synchronizing hash linked event streams.
+
+## Abstract
+<!--A short (~200 word) description of the technical issue being addressed.-->
+This CAIP describes a protocol that enables peers in a network to synchronize self-certifying data structures. This is achived though creating a libp2p protocol which allows peers to query and gossip updates and synchronize event data. Furthermore the libp2p kademlia DHT is used to establish the network topology.
+
+## Motivation
+<!--The motivation is critical for CAIP. It should clearly explain why the state of the art is inadequate to address the problem that the CAIP solves. CAIP submissions without sufficient motivation may be rejected outright.-->
+[Bitswap](https://docs.ipfs.tech/concepts/bitswap/) as defined by and used by [IPFS](https://ipfs.io) is a decent way to synchronize data if you know the [CID](https://github.com/multiformats/cid) of this content. [CAR Mirror](https://github.com/fission-codes/spec/blob/main/car-pool/car-mirror/SPEC.md) improves upon bitswap and enables even more optimal synchronization of an IPLD DAG. These protocols are great when you want to synchronize a known data set. However, when you want to synchronize an ever changing data set like an event stream these protocols lack some features. This specification is focused on describing a protocol for retrieving the latest events of an event stream as represented by an IPLD DAG. The retrieval of these events are ensured to be in "verification order". In a self-certifying event stream this simply implies that the events are synchornized in the order in which they were created.
+
+## Specification
+<!--The technical specification should describe the standard in detail. The specification should be detailed enough to allow competing, interoperable implementations. -->
+
+### Protocol overview
+
+The stream sync protocol utilizes three main mechanisms in order to synchronize event streams between nodes. First of all the libp2p DHT is used to look up other nodes that provide the relevant streams, second a protocol for querying and gossiping tips ([CID](https://github.com/multiformats/cid) of latest event), and third a protocol for synchronizing events in verification order.
+
+### Protocol messages
+
+The stream sync protocol has four main message types that nodes send to each other. It’s implemented as a libp2p sub-protocol, and is inspired by [ipfs bitswap](https://docs.ipfs.tech/concepts/bitswap/).
+
+**Libp2p protocol id:**
+
+```
+/caip/streamsync/1.0.0
+```
+
+#### Synchronizing tips
+
+The following two protocol messages are used to query and gossip tips.
+
+**SubscribeSet**
+
+This message is used to subscribe to a set of streams. 
+
+```ipldsch
+type SubscribeSet struct {
+  separatorKey nullable String
+  separatorValue nullable Bytes
+  streamFilter nullable Bytes
+  controllerFilter nullable Bytes
+} representation tuple
+```
+
+*Properties:*
+
+* `separatorKey` - a string key that has to be set in the *genesis event header* of a matching stream
+* `separatorValue` - the value set for the `separatorKey`
+* `streamFilter` - a byte array which streamids in this set must end with, e.g. if this property contains 3 bytes all streamids that ends with these three bytes are part of the set
+* `controllerFilter` - A bloom filter populated with the DIDs that the set includes
+
+*Invariants:*
+
+* If `separatorKey` and `separatorValue` are null, `streamFilter` must be a full streamId
+* If `separatorKey` and `separatorValue` are set, `shardFilter` and `ctrlFilter` are optional
+
+**HaveTips**
+
+A response to a *SubscribeSet* message or an update sent to known subscribed peers containing the latest known tips of the given stream set.
+
+```ipldsch
+type HaveTips struct {
+  tips [Tips]
+  separatorKey nullable String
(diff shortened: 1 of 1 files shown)


CAIPs/caip-stream-sync.md
```diff
@@ -0,0 +1,193 @@
...
+
+<!--You can leave these HTML comments in your merged EIP and delete the visible duplicate text guides, they will not appear and may be helpful to refer to if you edit it again. This is the suggested template for new EIPs. Note that an EIP number will be assigned by an editor. When opening a pull request to submit your EIP, please use an abbreviated title in the filename, `eip-draft_title_abbrev.md`. The title should be 44 characters or less.-->
+
+## Simple Summary
+<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the CAIP.-->
+A libp2p protocol for synchronizing hash linked event streams.
+
+## Abstract
+<!--A short (~200 word) description of the technical issue being addressed.-->
+This CAIP describes a protocol that enables peers in a network to synchronize self-certifying data structures. This is achived though creating a libp2p protocol which allows peers to query and gossip updates and synchronize event data. Furthermore the libp2p kademlia DHT is used to establish the network topology.
```
avivash, line 23: small typo with `achieved` 👀

zachferland:
whats an example of separatorKey(/value) and streamFilter? 

is there maybe strategies to partly still build on gossipsub? ie use the robustness of the underlying unstructured p2p overlay, control mesh, peer scoring, etc and then build separatorKey(/value) and streamFilter based overlays similar to what gossipsub already does for topics. Also similar to episub as well where you combine the robustness of an unstructured p2p overlay with efficient (use case specific) structured overlays (trees) on top, and episub implemented as a router (not sure if this is still possible or any progress on that)

oed:
@zachferland As an example, in Ceramic the `separatorKey` would be `"model"` and the value the streamid bytes of the model set for that stream.

Not sure I follow exactly how you mean about partly using gossipsub. Could you elaborate a bit?

CAIPs/caip-stream-sync.md
bumblefudge, line 23:
```suggestion
This CAIP describes a protocol that enables peers in a network to synchronize self-certifying data structures. This is achieved though creating a libp2p protocol which allows peers to query and gossip updates and synchronize event data. Furthermore the libp2p kademlia DHT is used to establish the network topology.
```

bumblefudge:
i, for one, would greatly benefit from hearing the two of you talk through these multiple pub/sub architectures in person... any chance you can both make it to at least half of the authZ biweekly tomorrow to flesh this stuff out? my main concern about this CAIP is that i'm not entirely seeing how useful it would be to people NOT using ceramic streams, but maybe a non-normative section about how it could work in different pub/sub architectures would get us halfway there and help tease a non-ceramic usecase out of the either from activityStreams or some other network...

oed:
@bumblefudge if we push the meeting by 30 min both Zach and I will be able to make it :)

CAIPs/caip-stream-sync.md
```diff
@@ -0,0 +1,193 @@
...
+
+* `separatorKey` - a string key that has to be set in the *genesis event header* of a matching stream
+* `separatorValue` - the value set for the `separatorKey`
+* `streamFilter` - a byte array which streamids in this set must end with, e.g. if this property contains 3 bytes all streamids that ends with these three bytes are part of the set
+* `controllerFilter` - A bloom filter populated with the DIDs that the set includes
+
+*Invariants:*
+
+* If `separatorKey` and `separatorValue` are null, `streamFilter` must be a full streamId
+* If `separatorKey` and `separatorValue` are set, `shardFilter` and `ctrlFilter` are optional
```
stbrody, line 73: is this supposed to say `streamFilter` and `controllerFilter`?

CAIPs/caip-stream-sync.md
```diff
@@ -0,0 +1,193 @@
...
+
+**HaveTips**
+
+A response to a *SubscribeSet* message or an update sent to known subscribed peers containing the latest known tips of the given stream set.
+
+```ipldsch
+type HaveTips struct {
```
stbrody, line 80: `HaveStreams`?

CAIPs/caip-stream-sync.md
```diff
@@ -0,0 +1,193 @@
...
+  tips [Tips]
+  separatorKey nullable String
+  separatorValue nullable Bytes
+  streamFilter nullable Bytes
+  ctrlFilter nullable Bytes
+} representation tuple
+
+type Tips struct {
```
stbrody, line 88: StreamTips?

stbrody: trying to figure out a way for the struct to not have the same name as the field within the struct

CAIPs/caip-stream-sync.md
```diff
@@ -0,0 +1,193 @@
...
+The following two protocol messages are used to synchronize event data between peers.
+
+**WantStream**
+
+Request to stream events from another node given a streamId.
+
+```ipldsch
+type WantStream struct {
+  streamid Bytes
+  knownTips [CID] // the CID(s) of the event to start the streaming from
```
stbrody, line 113: the stream is going to start from the oldest event right?  So this isn't really where we'll "start the streaming from".  But it's also not where the streaming will end, since I assume if new events come in during the sync process we'll still send them and not just end the stream when the originally known tip is reached.  So then what is this knownTips actually used for?  Do we even need it?

CAIPs/caip-stream-sync.md
```diff
@@ -0,0 +1,193 @@
...
+  offset nullable Integer // number of bytes already received
+} representation tuple
+```
+
+*Properties:*
+
+* `streamid` - the streamId being requested
+* `knownTips` - the CIDs of the tips that the requester already has locally
+* `offset` - number of bytes already recieved on top of the known tip
```
stbrody, line 122: "on top of the known tip", or since the beginning of the stream from genesis event?

CAIPs/caip-stream-sync.md
```diff
@@ -0,0 +1,193 @@
...
+3. `A` also announces to the DHT that it is subscribed to `s`
+4. `A` sends a `SubscribeSet` message for `s` to `B`
+5. `B` registers that `A` is subscribed to `s` with a ttl of 12h
+6. `B` sends a `HaveTips` message with the tips it has validated as a response to `A`
+7. `A` registers that `B` is subscribed to `s'` (as defined in the `HaveTips` message) with a ttl of 12h
+
+#### Publishing updates
+
+1. A new event is created in a stream of `s`, the event creator adds this event to peer `A`
+2. `A` sends a `Have` message to all peers in its registry which subscribe to `s`
```
stbrody, line 166: HaveTips?  There's no general "Have" message right?

CAIPs/caip-stream-sync.md
```diff
@@ -0,0 +1,193 @@
...
+
+1. Peer `A` wants to subscribe to stream set `s`
+2. `A` looks up other peers that subscribe using the libp2p DHT, it finds peer `B`
+3. `A` also announces to the DHT that it is subscribed to `s`
+4. `A` sends a `SubscribeSet` message for `s` to `B`
+5. `B` registers that `A` is subscribed to `s` with a ttl of 12h
+6. `B` sends a `HaveTips` message with the tips it has validated as a response to `A`
+7. `A` registers that `B` is subscribed to `s'` (as defined in the `HaveTips` message) with a ttl of 12h
+
+#### Publishing updates
```
stbrody, line 163: so it seems like with this strategy, if a new event in a set is published, let's say to node A, then A first tells the subscribed peers that it has a new event, then the subscribes have to come back and request the actual event data, and then A can send the new events.  Should there be a way to skip the first two steps and just have A immediately begin publishing new events to known subscribers as soon as new events come in?

CAIPs/caip-stream-sync.md
```diff
@@ -0,0 +1,193 @@
...
+3. `A` repeats step 1 until it has caught up with the desired tip(s)
+
+#### Maintaining connectivity
+
+1. Every 6h nodes will send new `SubscribeSet` messages for all of its streams to all connected peers
+1. Recieving peers will respond as described in *Tip Subscription* from step 5
+
+## Rationale
+<!--The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages. The rationale may also provide evidence of consensus within the community, and should discuss important objections or concerns raised during discussion.-->
+At a first glance the `SubscribeSet` and `HaveTips` messages and protocol one might consider why not to use libp2p gossipsub instead. While gossipsub is a great protocol that has been battle tested there are a few reasons why it is not optimal for this use case.
```
stbrody, line 181:
```suggestion
At a first glance at the `SubscribeSet` and `HaveTips` messages and protocol one might consider why not to use libp2p gossipsub instead. While gossipsub is a great protocol that has been battle tested there are a few reasons why it is not optimal for this use case. 
```

CAIPs/caip-stream-sync.md
```diff
@@ -0,0 +1,193 @@
...
+
+Another consideration is that in gossipsub messages are propagated though many peers. However, in order to synchronize the events of a stream peers need to be directly connected to the originating peer anyway. In the future peers might be disincentivized to forward `HaveTips` messages since future versions of this protocol could introduce payments for event retrieval.
+
+The purpose of this protocol is to synchronize hash linked and signed event logs. This is the reason that event streams need to be synchronized in verification order, e.g. `event0` -> `event1` -> `event2` and so on. This is to prevent DoS attacks where if the events is sent from tip to first event an attacker can sent a lot of invalid events and the victim node would need to synchronize all data before seeing that these events are invalid.
```
stbrody, line 187:
```suggestion
The purpose of this protocol is to synchronize hash linked and signed event logs. This is the reason that event streams need to be synchronized in verification order, e.g. `event0` -> `event1` -> `event2` and so on. This is to prevent DoS attacks where if the events are sent from tip to first event an attacker can send a lot of invalid events and the victim node would need to synchronize all data before seeing that these events are invalid.
```

stbrody (commented):
super excited about this

oed:
Closing this as we are no longer pursuing this approach.

PR #168, opened 2022-11-07
Draft: IPLD Timestamp Proof

zachferland:
Create and verify IPLD based blockchain timestamp proofs.


bumblefudge:
surely there's a codec for filecoin?  should i ask CASA's fil rep or in the filecoin open slack?

bumblefudge:
i also wonder if there's a link worth adding to the references section at the end pointing to info on how FVM means filecoin nodes can connect to a testnet/quorum/etc "out of the box"...

oed:
@bumblefudge why is filecoin's codec and FVM relevant in particular here? This is a general way of representing a timestamp from any blockchain.

bumblefudge:
Neither required nor especially relevant, just wondering if a third example would be nice. Since some ipfs gateways are also fil nodes, might be an appealing option for some users since it's one-stop shopping in that case!

oed:
I'm not familiar enough with Filecoin to say how easy it would be to add an example from there.

CAIPs/caip-ipld_anchor_event.md
```diff
@@ -0,0 +1,129 @@
...
+}
+```
+
+Blockchain transaction resolution, parsing, and verification can later be defined by each blockchain CAIP-2 namespace if needed. Implementations will vary and may even vary for the same chain. Optional parameters can be included here to indicate the methods required. For now transaction resolution, parsing, and verification is out of scope for this specification. 
+
+### IPLD Based Merkle Proofs
+
+A Merkle proof allows verification that a given piece of data or “leaf” was included in the set of data for the given Merkle Tree. A  typical proof includes a Merkle root of the tree and a subset list of tree node hashes that allow verification by reconstructing the Merkle root starting from the data or "leaf" you are interested in.
+
+In IPLD a Merkle root (CID) and a “path” are provided that allows you to traverse the tree from the root to the piece of data your interested in. Paths are strings and refer to [Pathing in IPLD](https://ipld.io/docs/data-model/pathing/). Following an IPLD path, and ultimately resolving the data or "leaf" you are interested in from the root, is equivalent to a standard proof.
```
rvagg, line 84: > is equivalent to a standard proof

Yeah, if you're willing to deal with the resolution of the intermediates. The approach of just providing a root and a path is fine if you expect the verifier to navigate through the data but the simple form of a Merkle proof provides the intermediate dependent hashes all the way up to the root so you can reconstruct it from just the data you have. You could do that here if you're just going to build a standard Merkle tree where intermediates are tuples of hashes (as you've suggested above) by providing the paired CIDs that couple with the derived CIDs of the leaf you want to prove - so leaf + CIDs of all non-relevant branches on the way up to the root. Then you can reencode those intermediates from the leaf you have and couple it with the provided CIDs to the original Merkle root which, if it matches is the "proof".
But that's not the approach being taken here, which is fine but assumes the availability and navigability of all intermediates. Is that correct?

CAIPs/caip-ipld_anchor_event.md
```diff
@@ -0,0 +1,129 @@
...
+IPLD Merke Tree Node Type
+
+```tsx
+type NodeList [Link]
+```
+
+Example 1
+
+```tsx
+[CID(bagcqcerav5zyrlelnnx24blbgob6stz6y4oiwenn3hodbmyf7bnnguh6gdeq), CID(bagcqcera4nnpzmwsk2bbww6n4bgq3cwne6gmhkouciam44nyjbindkxpz2eq)
```
rvagg, line 52: This is a dag-jose CID, is that intentional or just for example purposes?

CAIPs/caip-ipld_anchor_event.md
zachferland: yeah roughly but the intention is to not assume anything about the network (ie just IPLD, not IPFS), this just describing the data structure in IPLD and in IPLD a natural way to verify is just to traverse the path. The data could already be local, over another network, or an implementation may not rely on any intermediaries by packing the entire proof and path subset of the tree (all necessary ipld blocks) in a CAR file if they choose. Probably makes sense to change wording a bit (ie resolving) and maybe add a note. thanks for commenting

CAIPs/caip-ipld_anchor_event.md
zachferland: not intentional, doesnt really matter how leaves are encoded, nodes are likely to be dag-cbor, so may change here, but also not required

CAIPs/caip-ipld_anchor_event.md
```diff
@@ -0,0 +1,129 @@
...
+```
+
+Example 2
+```tsx
+{
+  root: CID(bafyreiaxdhr5jbabn7enmbb5wpmnm3dwpncbqjcnoneoj22f3sii5vfg74)
+  chainID: "eip155:1"
+  txHash: CID(bagjqcgzanbud4sqdsywfp2mckuj57qsffsovgyjhh7sxebkqwr335hzy2zbq)
+}
+```
```
oed, line 76: We likely will need an additional property here to be able to distinguish between different transaction types. We basically need to know how to take the raw tx data and convert it to a CID. For example if it's a tx to a smart contract it will need a different way of decoding than if it's a regular tx.
```suggestion
```tsx
type BlockchainAnchor struct {
  root Link
  chainID String 
  txHash Link
  txType String
}
\```

Example 2
```tsx
{
  root: CID(bafyreiaxdhr5jbabn7enmbb5wpmnm3dwpncbqjcnoneoj22f3sii5vfg74)
  chainID: "eip155:1"
  txHash: CID(bagjqcgzanbud4sqdsywfp2mckuj57qsffsovgyjhh7sxebkqwr335hzy2zbq)
  txType: 'eth-raw'
}
\```
```

looks like the suggestion system doesn't really work over quote blocks...

oed:
For example, I'm planning to register `f(bytes32)` for the `eip155` namespace, which would map to: "The bytes32 param is the multihash of a dag-cbor CID".

chunningham:
If there's a registry then that's fine, I was kind of thinking that the role of discriminant would be duplicated here between codec and string, but I suppose that the string is more an indicator of which links in the block actually point to/include the root of the merkle tree than of the structure of the block? By 'errors' I meant blocks with e.g. codec is `raw` but the discriminant says `eth-contract` then that's allowed but nonsensical, or at least hard to parse correctly.

CAIPs/caip-ipld_anchor_event.md
bumblefudge: update example to dag-cbor as discussed on today's call?

bumblefudge:
Todo:
- [ ] Adding language to specify that per-namespace profiles are needed.  
- [ ] Add an eip155 profile to namespaces/eip155/caip168
- [ ] Tell @bumblefudge on discord or here when ground is broken on the sample implementation to bump to `Review`

bumblefudge:
NB @Dominikwoelk as per our convo last night

CAIPs/caip-ipld_anchor_event.md
zachferland: updated

CAIPs/caip-ipld_anchor_event.md
zachferland: txType added, and ref to namespace for them

zachferland:
Minor edits made, references to namespace, txType added, and eip155 namespace PR opened - https://github.com/ChainAgnostic/namespaces/pull/42 cc @bumblefudge

oed:
Also an implementation of validation of this CAIP can be seen here: 
https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts

Should address all of your points above @bumblefudge :)

bumblefudge:
Hey guys! Sorry for the delay, I did an editorial pass on both CAIP and Namespaces/eip155/CAIP168.  I had to open new PRs against Ceramic's fork, so if you merge those two I'll merge here.  

Also note: if there's working code that works against this final version of the spec, we can skip draft and go straight to `Review` 🎉

zachferland:
editorial pass merged

bumblefudge:
oh sorry I didn't link this one, also needed: https://github.com/ceramicnetwork/namespaces/pull/1

zachferland:
thanks, merged the namespace pass as well now

PR #172, opened 2022-11-14
Update caip-122.md status to review

bumblefudge:
in the EIP process, we're ready to mark as "In review" as soon as the authors mark it as such.  here, I wonder if there should be some kind of "in prod" and/or "two implementations" requirement, but that's a bigger question for thursday  :D


oed (commented):
lgtm, thanks!

PR #171, opened 2022-11-08
Add a CAIP describing opened session tokens

ritave:

PR #170, opened 2022-11-08
Pre-Draft: Browser-Session Data Model

bumblefudge:

Issue #141, opened 2022-09-26
CAIP-25: Should there be a notion of sessions, and should they expire?


rekmarks:
Ref: #138 

## Problem

CAIP-25 currently does not mention anything about the length of time for which a handshake persists, or what we may call "sessions". Should CAIP-25 specify some notion of "sessions" and whether/how they expire?

kdenhartog:
+1 we've found that time based permissioning can be extremely helpful for our browser permissions and I'd think this would be really useful for the wallet as well. The current options for time to reset/re-ask for consent are:

![Screen Shot 2022-09-29 at 4 28 26 PM](https://user-images.githubusercontent.com/23125059/192932023-9b014598-7790-43ec-90a8-3f1d4075a856.png)

depatchedmode:
@kdenhartog Do you happen to know how many people end up using any of time based expiry for permissions?

kdenhartog:
I don't believe we gather any data on this. Here's the list of P3A data we gather for telemetry if the user opts in for this: https://github.com/brave/brave-browser/wiki/P3A

bumblefudge:
In the interest of moving 25 to `Review`, we might delegate/punt this to the #170 CAIP and wait for interested parties to harmonize session management there, if everyone is happy with the assumptions codified in the more topical/external #171.

bumblefudge:
merging further updates to CAIP-25 today, but still no progress on CAIP-170, so [removing the CAIP-170 constraint](https://github.com/ChainAgnostic/CAIPs/pull/217/commits/0de8414b5f4d5b9708fc447f6a76427fe93d4977) for now.  Leaving this issue open in case session design makes quick progress, but most likely the session object's properties will remain undefined in 25 and be constrained by an optional extensions later down the road, if at all.

kdenhartog:
leaving this as an optional extension point will likely impact interoperability. It is possible to achieve, but the downside here is that on the DApp side logic has to fail gracefully here in a much more complex way. Especially if we plan to support things like multi-transaction flows or multi userOps (for AA style wallets) because in these case the content length may be larger than what would fit within a single request.

I suspect we could look to what HTTP has done over it's various revisions to find what the right pattern would be here if we don't want to constrain this to 170 right now.

PR #173, opened 2022-11-16
Adjusting CAIP-171 as per last meeting

bumblefudge:
- generalize "token" to "identifier" to be more platform-neutral
- made assumptions more explicit
- add stability assumption (identifier doesn't change as session state changes)

NB @ritave  (can't tag you for review directly)

partly addresses #141 , although #170 is probably the place to further specify session management for shared assumpions.  ideally the choice to use #170 session **objects** would be separate from the choice to use #171 session **identifiers**, and wallets or users would know what they can assume based on one or both of the standards being conformed to?


CAIPs/caip-171.md
```diff
@@ -0,0 +1,64 @@
...
+`required` front-matter property, it SHALL be interpreted as reference to this
+specification.
+
+> Notice that there are no code constraints on the token value except
+> that it be expressed as a string with few reserve characters, under 256
+> characters in length. This is by design and the value is
+> implementation-dependent.
+
+```typescript
+type SessionToken = string;
```
ritave, line 50:
```suggestion
type SessionIdentifier = string;
```

CAIPs/caip-171.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+Properties of the `sessionIdentifier` are as follows:
```
ritave, line 53:
```suggestion
Properties of the `SessionIdentifier` are as follows:
```

CAIPs/caip-171.md
```diff
@@ -0,0 +1,64 @@
...
+2119](https://www.ietf.org/rfc/rfc2119.txt)
+
+### Definition
+
+Whenever a CAIP uses a name `SessionIdentifier` and has this CAIP in the
+`required` front-matter property, it SHALL be interpreted as reference to this
+specification.
+
+> Notice that there are no code constraints on the token value except
+> that it be expressed as a string with few reserve characters, under 256
```
ritave, line 45: Remove the size limit from a non-normative section and let's put it into specification below

CAIPs/caip-171.md
```diff
@@ -0,0 +1,64 @@
...
+
+```typescript
+type SessionToken = string;
+```
+
+Properties of the `sessionIdentifier` are as follows:
+1. It MUST uniquely identify an open and stateful session. 
+1. It MUST identify a closeable session, and it MUST become invalid
+   after a session is closed.
+1. It MUST remain the same as the identified session's state changes.
```
ritave, line 57: I don't understand this point.

CAIPs/caip-171.md
```diff
@@ -0,0 +1,64 @@
...
+> that it be expressed as a string with few reserve characters, under 256
+> characters in length. This is by design and the value is
+> implementation-dependent.
+
+```typescript
+type SessionToken = string;
+```
+
+Properties of the `sessionIdentifier` are as follows:
+1. It MUST uniquely identify an open and stateful session.
```
ritave, line 54: Please actually increment the bulled points

CAIPs/caip-171.md
```diff
@@ -0,0 +1,64 @@
...
+```typescript
+type SessionToken = string;
+```
+
+Properties of the `sessionIdentifier` are as follows:
+1. It MUST uniquely identify an open and stateful session. 
+1. It MUST identify a closeable session, and it MUST become invalid
+   after a session is closed.
+1. It MUST remain the same as the identified session's state changes.
+1. It MUST be serializable into JSON. Serialization and later deserialization using
```
ritave, line 58: `1. It MUST be between inclusive 1 and 256 characters in length.`

CAIPs/caip-171.md
```diff
@@ -0,0 +1,64 @@
...
+JSON MUST result in the same value.
+
+## Copyright
+
+Copyright and related rights waived via
+[CC0](https://creativecommons.org/publicdomain/zero/1.0/).
```
ritave, line 64: Also I'd suggest having the CC0 in the LICENSE file of this repo as well.

That's what we do in [Snaps Improvement Proposals](https://github.com/MetaMask/SIPs/blob/60c212d1da607005bef989fd364565f21c66f111/sip-template.md?plain=1#L37)

CAIPs/caip-25.md
```diff
@@ -137,9 +137,11 @@ The valid error messages codes are the following:
     * code = 5102
     * message = "Requested events are not supported"
 
-## Links
+## Changelog
 
-n/a
+- 2022-11-26: add mandatory indexing by session identifier (i.e. CAIP-171 requirement)
```
ritave, line 142: It's a Draft, imho the changelog shouldn't be in it until it's Final, where we could add Errata.

ritave (approved):
Generally looks good with minor asks

CAIPs/caip-171.md
bumblefudge: 🤦‍♀️

CAIPs/caip-171.md
bumblefudge: the identifier remains constant when the state changes, i.e. not content-addressable/hash-based.  better phrasing appreciated!

CAIPs/caip-171.md
bumblefudge: oops let's leave that in the separate PR so it doesn't block this one

CAIPs/caip-171.md
bumblefudge: good point, better not to rely on markdown for people that will be reading it in emacs/vim/etc

PR #177, opened 2022-11-17
Add CC0 license

ligi:
as it came up in the meeting today
inline to not rely on the external link

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,121 @@
+Creative Commons Legal Code
+
+CC0 1.0 Universal
+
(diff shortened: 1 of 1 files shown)


ukstv (approved):

ritave (changes requested):
I'd say to get this in we need to update the template, sip-1 process, and all CC0 links in all SIPs?

ligi:
SIP?

ritave:
My bad, I said Snaps Improvement Proposals, when I meant CAIPs

PR #180, opened 2022-11-18
Support underscores in chain ID reference

delaaxe:
Adding support for underscores in the chain id reference allows using StarkNet chain ids as-is (`SN_MAIN`, `SN_GOERLI`) without having to do error-prone translations in both dapps and wallets.

Such a change shouldn't impact the existing requirements for the reference.

### DIFF
--- a/CAIPs/caip-10.md
+++ b/CAIPs/caip-10.md
@@ -32,7 +32,7 @@ The `account_id` is a case-sensitive string in the form
 
 ```
 account_id:        chain_id + ":" + account_address
-chain_id:          [-a-z0-9]{3,8}:[-a-zA-Z0-9]{1,32}
+chain_id:          [-a-z0-9]{3,8}:[-_a-zA-Z0-9]{1,32}
 account_address:   [a-zA-Z0-9]{1,64}
 ```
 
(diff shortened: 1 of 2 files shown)


oed:
Could you also add a PR to add starknet to the namespaces repo?
https://github.com/ChainAgnostic/namespaces/

delaaxe:
> Could you also add a PR to add starknet to the namespaces repo? https://github.com/ChainAgnostic/namespaces/

Will do the PR.

Drafting this one for now as there may be better ways of representing StarkNet chain ids that don't require this change.

oed:
@pedrouid any reason not to add support for `_`?

delaaxe:
@oed @pedrouid here is PR for `namespaces`: https://github.com/ChainAgnostic/namespaces/pull/39

pedrouid (approved):
LGTM 👍

bumblefudge (approved):

bumblefudge:
@delaaxe this LGTM but maybe we can get the namespace finalized by thursday the 15th so that we don't have an example in a CAIP that isn't backed up by a namespace entry?

PR #181, opened 2022-11-20
Formal fixes

ligi:
Currently working on a CAIP checker that can check headers for correctness - this removes some inconsistencies

### DIFF
--- a/CAIPs/caip-25.md
+++ b/CAIPs/caip-25.md
@@ -7,7 +7,7 @@ status: Draft
 type: Standard
 created: 2020-10-14
 updated: 2022-10-26
-requires: ["2", "10"]
+requires: 2, 10
 ---
 
 ## Simple Summary
--- a/CAIPs/caip-25.md
+++ b/CAIPs/caip-25.md
@@ -7,7 +7,7 @@ status: Draft
 type: Standard
 created: 2020-10-14
 updated: 2022-10-26
-requires: 2, 10
+requires: [2, 10]
 ---
 
 ## Simple Summary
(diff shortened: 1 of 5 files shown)


CAIPs/caip-25.md
```diff
@@ -7,7 +7,7 @@ status: Draft
 type: Standard
 created: 2020-10-14
 updated: 2022-10-26
-requires: ["2", "10"]
+requires: 2, 10
```
bumblefudge, line 10: It appears the `"`s were unnecessary, but the `[]`s *were* necessary.  Not sure if this is all Yaml, or the specific version of Yaml Jekyll uses, or just Jekyll, but:

https://talk.jekyllrb.com/t/solved-front-matter-array-with-multiple-values/4426

CAIPs/caip-3.md
```diff
@@ -4,7 +4,7 @@ title: Blockchain Reference for the EIP155 Namespace
 author: Simon Warta (@webmaster128), ligi <ligi@ligi.de>, Pedro Gomes (@pedrouid)
 discussions-to: https://github.com/ChainAgnostic/CAIPs/issues/3, https://github.com/ChainAgnostic/CAIPs/pull/1
 status: Draft
```
bumblefudge, line 6:
```suggestion
status: Superseded
```

bumblefudge: (Embarrassing! I'll cop to it, it was me that changed the wrong property to Superseded haha)

CAIPs/caip-25.md
ligi: OK then the other way around - but for now we had both styles which was not good:

CAIPs/CAIPs/caip-25.md:requires: ["2", "10"]
CAIPs/CAIPs/caip-76.md:requires: 10, 75

CAIPs/caip-3.md
ligi: if this is a real state we need to add it to CAIP-1

bumblefudge (approved):

CAIPs/caip-25.md
bumblefudge: Fixed!

bumblefudge: And a comment added to CAIP-1 and to the template.md

PR #182, opened 2022-11-20
Add workflow to check CAIPs

ligi:
#181 needs to be merged first


ritave:
Wouldn't you prefer to use something already built instead of rebuilding a new one?

I've written a re-usable markdown linter for SIPS, which operates on Abstract Syntax Tree of markdown and allows writing custom rules per repository.

For example, this rule checks any external links for 400 statues by [visiting specific nodes in AST](com.github.komputing:XIP-checker).

I could release the linter into MIT and NPM and allow it to load CAIPs specific plugins.

ligi:
This is reused - I used it for DIPs before. Link checker there I used another one before - gaurav-nelson/github-action-markdown-link-check@v1 - but actually removed it from this step as Link checking is not deterministic.
Feel free to add yours afterwards - this already found stuff and can guard until something else is in place.

ritave:
I'm good with whichever if we're not writing one anew again.

jackromo888:
8

obstropolos:
@ligi  - is this all working?

Issue #139, opened 2022-09-25
CAIP-25: `extensions` considered harmful?


rekmarks:
Ref: #138

## Problem

When the dapp submits a CAIP-25 request, it may desire to request some methods or events for specific chains only, as opposed to the entire namespace. Given that top-level keys in the request object are namespaces, this presents a challenge. Consider:

```json
{
  "eip155": {
    "chains": ["eip155:137", "eip155:1"],
    "methods": ["eth_sign"],
    "events": ["accountsChanged"],
  },
}
```

The current WalletConnectV2 implementation of CAIP-25, [solves this by adding a field named `extensions`](https://github.com/WalletConnect/walletconnect-docs/blob/fa87a0b572d0f8d3e522686cb8ca42f0026b4a74/docs/specs/sign/session-namespaces.md#example-proposal-namespaces-request) to the namespace object:

```json
{
  "eip155": {
    "chains": ["eip155:137", "eip155:1"],
    "methods": ["eth_sign"],
    "events": ["accountsChanged"],
    "extensions": [
      {
        "chains": ["eip155:137"],
        "method": ["personalSign"],
        "events": ["chainChanged"]
      }
    ]
  },
}
```

While this certainly works, it introduces complexity on the part of the wallet, which must essentially perform set operations (IIRC, the word used was "diffing") on the configuration for chains for which extensions are specified. Multiple people objected to this, and argued that a configuration format should explicitly state the configuration for each chain/namespace.

## Proposed Solution

Instead of the `extensions` field, attendees suggested that top-level keys can be chain ids in addition to namespaces. This makes the object "literal" and ensures that no set operations have to be performed by the wallet during parsing:

```json
{
  "eip155:1": {
    "methods": ["eth_sign"],
    "events": ["accountsChanged"],
  },
  "eip155:137": {
    "methods": ["eth_sign", "personalSign"],
    "events": ["accountsChanged", "chainChanged"],
  },
  "cosmos": {
    "chains": ["cosmos:cosmoshub-4", "cosmos:cosmoshub-5"],
    "methods": ["cosmos_signDirect"],
    "events": ["someCosmosEvent"]
  }
}
```

orenyomtov:
My 2 cents on the subject:

Let's keep the JSON-RPC API spec to be as close to a 1-to-1 representation of the data that it is conveying.

[...]

PR #185, opened 2022-11-22
alternate extensions model

bumblefudge:
One possible solution to #139 , which proposes a mix of namespace-scoped and chainId-scoped authorizations to enable complex multi-chain authorizations (or chain-specific methods in multi-chain namespaces) without extensions.  See also the notes from the [RPC meeting 9 nov](https://github.com/ChainAgnostic/CASA/issues/52)

Signed-off-by: bumblefudge <jcaballero@centre.io>

### DIFF
--- a/CAIPs/caip-25.md
+++ b/CAIPs/caip-25.md
@@ -97,7 +97,7 @@ An example of a successful response should match the following format:
   "id": 1,
   "jsonrpc": "2.0",
   "result": {
-    "session": "0xdeadbeef",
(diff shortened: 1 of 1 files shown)


bumblefudge:
One possible solution to #139 , which proposes a mix of namespace-scoped and chainId-scoped authorizations to enable complex multi-chain authorizations (or chain-specific methods in multi-chain namespaces) without extensions.  See also the notes from the [RPC meeting 9 nov](https://github.com/ChainAgnostic/CASA/issues/52)

PR #186, opened 2022-11-24
Fleshing out CAIP-27

bumblefudge:
I'm thinking that many of these tasks can't be finished until CAIP-25 is in `Review`, but here's a checklist for now:
- [X] clarify session management and division of labor between 25/27 on this point
- [ ] define the response
- [ ] since the example method (evm `personal_sign` has [assumptions](https://stackoverflow.com/a/72348743) which may vary wildly across VMs, add explicit assumptions about `account` behavior to help people understand where/whether this matches their own use-cases.  better yet, just add a ### Use-cases section after the ## Specification intro
- [ ] add a test vector to help people test implementations.  Not sure this is even possible, but my idea would be:
    + human-readable `personal_sign` message and dummy-address seed phrase --> sample request with hexcode params --> sample response --> sample CACAO

### DIFF
--- a/CAIPs/caip-27.md
+++ b/CAIPs/caip-27.md
@@ -7,19 +7,19 @@ status: Draft
 type: Standard
 created: 2020-12-12
 updated: 2022-11-16
-requires: ["2", "25", "171"]
+requires: [2, 25, 171]
 ---
 
 ## Simple Summary
 
-CAIP-27 defines a standard JSON-RPC method for requesting methods mapped to a
-target chain.
+CAIP-27 defines a standard JSON-RPC method for requesting specific RPC
+methods, mapped to one specific target chain, one per call.
 
 ## Abstract
 
 This proposal has the goal to define a standard method for decentralization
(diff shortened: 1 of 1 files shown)


bumblefudge:
Some thoughts from a brief meeting today:
- [ ] Should the ## Motivation section be updated?
- [ ] Should CAIP-27 explicitly be limited to methods pre-authorized by a CAIP-25 call/session object?  i.e. should "method not previously authorized" be an error response?
- [ ] Do some methods require an account, and thus a separate error message? Or is that best left outside of the scope of 27, for caip 25 and/or the node itself to provide?
- [ ] No idempotency? What happens if multiple reqs sent, e.g. cuz response is slow in coming?

Error messages (straw man):
- invalid params (sessionId, chainId, etc)
- inadequate authZ captured in session
- no way for PROVIDER to validate params? Provider versus node errors/parsing

bumblefudge:
overtaken by events -- see #217

Issue #144, opened 2022-09-27
CAIP-25: The behavior regarding accounts is confusing


rekmarks:
Ref: #138 

## Problem

The relationship between accounts and CAIP-25 is, for lack of a better word, strange. The stated motivation of the specification is to define how "to expose accounts and define the expected JSON-RPC methods to be used by an application through a provider connecting to a wallet." Yet, it has very little to say about what accounts are, what the dapp must do to ask for them, and how they are approved by the user.

To see why this may be a problem, consider the only existing CAIP-25 implementation, WalletConnectV2. Per @ganchoradkov and @bkrem, WCv2 will reject `caip_handshake` requests if `eth_accounts` is not in the `methods` array for the `eip155` namespace, or if the user does not select any accounts to expose.  This is a perfectly reasonable design if you are familiar with Ethereum applications, but you could in no way predict this behavior from the text of CAIP-25.

During the 2022-09-15 CASA gathering, participants suggested that:
1. The request for accounts should be explicit in the `caip_handshake` request in order for accounts to be selected and returned.
2. That accounts should not be returned as part of the result at all.
3. That requesting accounts should always be optional.
    - (This was my point.)

Whatever the case, the concept and behavior of "accounts" in CAIP-25 merits further definition.

bumblefudge:
One quick fix would be to change the language from 
```
The accounts returned as a result should match the requested `chainId`s and
should be an array of CAIP-10 compliant `accountId`s.
``` 
to
```
The accounts returned as a result should match the requested `chainId`s and
should be an array of zero or more CAIP-10 compliant `accountId`s.
```

But that doesn't really solve the problem of how a dapp communicates to the wallet whether an empty array is acceptable or not!  
1. Should an optional parameter be added to the request? 
2. Should an error message be added if the dapp requests 1 or more accounts and the user chooses not to give any?

PR #187, opened 2022-11-24
Strawman of a CAIP-25 split into required/optional with session properties attached and verbose response

bumblefudge:
As per yesterday's RPC meeting, this is a straw man of a more verbose CAIP-25 request dividing its params between `required` and `optional` arrays, the latter of which includes session properties.  

This allows a simplified negotation-by-subset of optional authorizations, and a way for providers to override, ignore, or impose their own constraints to the session metadata.  This may offer enough flexibility for everything requested in #141, at the expense of some complexity around corner cases.  It also outsources to #170 the definition of valid properties and their valid values.  Note that there is currently no versioning for #170 objects.

We should decide where "accounts" go as per #144 -- I took the liberty of throwing them into the namespace/chain objects, and allowing all-but-one of them to be empty arrays, but this was simply for the sake of having a straw man to debate.  Opinions, corner-cases, UX nightmares, and new requirements welcome!

Signed-off-by: bumblefudge <jcaballero@centre.io>

### DIFF
--- a/CAIPs/caip-25.md
+++ b/CAIPs/caip-25.md
@@ -229,6 +229,7 @@ The valid error messages codes are the following:
   consolidated variants across issues and forks post-Amsterdam Gathering
 
 [RFC3339]: https://datatracker.ietf.org/doc/html/rfc3339#section-5.6
+[CAIP-170]: https://chainagnostic.org/CAIPs/caip-170
 
 ## Copyright
 
--- a/CAIPs/caip-25.md
+++ b/CAIPs/caip-25.md
@@ -39,36 +39,57 @@ forthcoming one.
 Within that session model, this interface outlines the authorization of an
 injected provider per namespace.
 
+The application interfaces with a provider to populate a session with a base
+state describing authorized chains, methods, event, and accounts.  This
+negotation takes place by sending the application's REQUIRED and REQUESTED
+properties of the session. If any requirements are not met, a failure response
+expressive of one or more specific failure states will be sent (see below).
+Conversely, a succesful response will contain all the required properties *and
+the provider's choice of the optional properties* expressed as a unified set of
+parameters.
+
 ### Request
 
-The application would interface with a provider to authorize a provider with a
-given set of parameters by calling the following JSON-RPC request
+Example:
 
-```
+```jsonc
 {
-    "id": 1,
-    "jsonrpc": "2.0",
-    "method": "provider_authorization",
-    "params": {
-        "eip155": {
-            "chains": ["eip155:1", "eip155:137"],
-            "methods": ["eth_sendTransaction", "eth_signTransaction", "eth_sign", "personal_sign"]
-            "events": ["accountsChanged", "chainChanged"]
+  "id": 1,
+  "jsonrpc": "2.0",
+  "method": "provider_authorization",
+  "params": {
+    "required": {
+      "eip155": {
+        "chains": ["eip155:1", "eip155:137"],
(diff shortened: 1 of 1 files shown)


bumblefudge:
On a 1:1 call with a CASA member discussing this PR, they mentioned some semantic tweaks:

1.) renaming `required` and `optional` to something more semantically clear vis-a-vis their contents, i.e. `scopes_required`/`scopes_requested` or `namespaces_required`/`namespaces_optional`.

2.) moving `sessionProperties` out to a top-level param (sister of `required` and `optional`) rather than inside of `optional` so that each namespace object would just contain namespace properties, i.e.

```jsonc
...
    },
    "optional":{
      "eip155:42161": {
        "methods": ["eth_sendTransaction", "eth_signTransaction", "get_balance", "personal_sign"],
        "events": ["accountsChanged", "chainChanged"]
      },
    },
    "sessionProperties": {
      "expiry": "2022-12-24T17:07:31+00:00",
      "auto-refresh": "true"        
    }
  },
```

CAIPs/caip-25.md
```diff
@@ -39,64 +39,136 @@ forthcoming one.
...
     },
-
+    "optional":{
+      "eip155:42161": {
+        "methods": ["eth_sendTransaction", "eth_signTransaction", "get_balance", "personal_sign"],
+        "events": ["accountsChanged", "chainChanged"]
+      },
+      "sessionProperties": {
+        "expiry": "2022-12-24T17:07:31+00:00",
+        "auto-refresh": "true"
```
pedrouid, line 82: Personally I'm not a fan of auto-refresh. I think the session should expire if there is no action taken by the requester

CAIPs/caip-25.md
```diff
@@ -39,64 +39,136 @@ forthcoming one.
...
+      }
     },
-
+    "optional":{
+      "eip155:42161": {
+        "methods": ["eth_sendTransaction", "eth_signTransaction", "get_balance", "personal_sign"],
+        "events": ["accountsChanged", "chainChanged"]
+      },
+      "sessionProperties": {
+        "expiry": "2022-12-24T17:07:31+00:00",
```
pedrouid, line 81: I would put this in the response instead of the request

IMO it should always be the responder that defines it and/or calculates it

Alternatively there should be an upper bond for the expiry (eg. 30 days)

bumblefudge: Hehe, scroll down, in my example the responder picked a 7-day expiry and overrode the request :D

CAIPs/caip-25.md
bumblefudge: hehe in my example the responder rejected that boolean.  but maybe there are other booleans like 
```
"enforce CAIP-154": "true",
```
?

CAIPs/caip-25.md
```diff
@@ -136,13 +208,28 @@ The valid error messages codes are the following:
...
+    * message = "Scope/chain mismatch"
+* When a badly-formed request defines one `chainId` two ways
+    * code = 5104
+    * message = "ChainId defined in two different scopes"
+* Invalid Session Properties Object
+    * code = 5200
+    * message = "Invalid Session Properties requested"
+* Required Session Properties 
+    * code = 5201
+    * message = "Session Properties can only be optional"
```
bumblefudge, line 222:
```suggestion
    * Required Session Properties 
    * code = 5201
    * message = "Session Properties can only be optional"
```
No longer child object

bumblefudge:
Self-merging after editorial clean-up as agreed on today's call

PR #195, opened 2022-12-16
Add optional Credential_Manifest method to CAIP-169

bumblefudge:
To allow more complex issuance.

Also rearranges methods to be more readable (and move all optional methods to the end)

### DIFF
--- a/CAIPs/caip-169.md
+++ b/CAIPs/caip-169.md
@@ -134,90 +134,6 @@ Claim Format registry][] if they are expecting reasonable interoperability in
 the near term, and to carefully construct their [presentation_definition
 object][] accordingly.
 
-### Store
-
-Stores the given VC in the CP.
-
-#### Method:
-
-`wallet_creds_store`
-
-##### Params:
-
-- `vc` - A Verifiable Credential.
-
-##### Returns:
-
-- `error` - OPTIONAL. If `vc` was malformed or does not comply with the
-  Verifiable Credentials Profile defined in this specification. Note that some
-  wallets will call `wallet_creds_verify` locally or remotely and pass back an
-  error message received thereby, depending on security context.
-    + 400 - invalid parameters
-    + 500 - processing or internal error
-    + 501 - invalid cryptographic primitives (see `wallet_creds_metadata`)
-
-### Verify
-
-Verifies the proof section of a single verifiable credential after dereferencing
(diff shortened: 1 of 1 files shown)


CAIPs/caip-169.md
```diff
@@ -308,6 +224,93 @@ Optionally, holder binding can also be requested.
...
+`wallet_creds_issue`
+
+##### Params:
+
+- `credential_application` - REQUIRED. This can vary from the full contents of
+  the payload of the to-be-issued credential to a mere consent event per
+  use-case, but in either case MUST be formated as a valid
+  [credential_application object][] as specified in the [Credential Manifest][]
+  specification.
+- `preferred_proofs` - OPTIONAL. An **ordered** array (from most to least
```
kimdhamilton, line 293: [Note: this is hard to review -- I can't figure out what's going on with the diffs. Visually it looks like there's no difference between store and verify, but they are getting pulled in. I tried messing with whitespace settings but no dice.]

I think there may be collision with this and credential application's [format](https://identity.foundation/credential-manifest/#credential-application) property: 
```
The object MUST have a format property if the related [Credential Manifest](https://identity.foundation/credential-manifest/#term:credential-manifest) specifies a format property. Its value MUST be a subset of the format property in the [Credential Manifest](https://identity.foundation/credential-manifest/#term:credential-manifest) that this [Credential Submission](https://identity.foundation/credential-manifest/#term:credential-submission) is related to. This object informs the [Issuer](https://identity.foundation/credential-manifest/#term:issuer) which formats the [Holder](https://identity.foundation/credential-manifest/#term:holder) wants to receive the [Claims](https://identity.foundation/credential-manifest/#term:claims) in.
```

It looks like this spec is assuming the presence of a credential manifest, I think it shouldn't need `preferred_proofs`.

bumblefudge: Agreed -- I think that was an OIDC interop thing, like the wallet metadata getter.  

@awoie would it make sense to warn readers of the spec that `preferred_proofs` is ONLY required for OIDC backwards compat and that the preferred way of communicating this is in the 
 `formats` property of the `credential_application`?

bumblefudge: (review most recent commit)

awoie (changes requested):
Can you describe a typical flow how a developer could use that for issuance and presentation of credentials?

PR #207, opened 2023-01-30
Add CSPRNG usage requirement

kdenhartog:
Some options to use here would be UUIDv4 (122 bits of entropy - 128 bits with 6 bits fixed), a SHA-256 hash of the ephemeral public key, or a CSPRNG generated number.


kdenhartog:
cc @pedrouid @haardikk21 @bumblefudge related to #206

bumblefudge:
Makes sense. I can add a security considerations section to 171 if the group agrees on this req. 

Where does ["randomness-aaS"](https://github.com/drand/drand) fit in the threat model? Hehe

kdenhartog:
> Where does ["randomness-aaS"](https://github.com/drand/drand) fit in the threat model? Hehe

In my mind that seems like a valid way to generate a session Id, but likely an implementation detail.

PR #211, opened 2023-02-02
[CAIP-25] Add CAIP-211 ("Rpc Authority Negotiation") as extension spec to CAIP-25

bumblefudge:
Tried to capture the skeleton of the discussion yesterday and propose one path forward.  Note the changes to CAIP-25 as well!

If anyone feels like adding anything or filling in TODO sections, be my guest! use "changes requested", I suppose?

NB @shanejonas

### DIFF
--- a/CAIPs/caip-207.md
+++ b/CAIPs/caip-207.md
@@ -1,0 +1,236 @@
+---
+caip: 207
+title: JSON-RPC Authority Negotiation
+author: Pedro Gomes (@pedrouid), Hassan Malik (@hmalik88)
+discussions-to: https://github.com/ChainAgnostic/CAIPs/pull/207
+status: Draft
+type: Standard
+created: 2023-02-02
+updated: 2023-02-02
+requires: [2, 10, 25, 171]
+---
+
+## Simple Summary
+
+CAIP-207 extends CAIP-25 to allow callers and respondents to negotiate RPC
+authorities and RPC semantics before authorizing RPC terms. 
+
+## Abstract
+
+CAIP-207 defines additional properties that CAIP-25 can be used
+iteratively/progressively to layer custom or local RPC semantics and/or routing
+onto a session. Since CAIP-25 respondents ignore unknown properties, respondents
+that conform to CAIP-25 but not to CAIP-207 should be carefully considered and
+accomodated by implementers. 
+
+## Motivation
+
+While some core methods and notifications are foundational to a namespace and
+universally defined out-of-band (meaning all callers and respondents agree to
+them already), others are specific to chains or even to subsets of wallets and
+dapps on a given chain, which requires semantic negotiation and/or network
+routing negotiation before authorization can occur.
+
+## Specification
+
+Two properties are added to the scope objects requested in `optionalScopes` and
+optionally returned in `sessionScopes`.  These are both [ordered] strings of arrays:
+
+1. `rpcEndpoints` is an array of zero or more URLs of RPC endpoints that
+   the caller would prefer the respondent to use, ordered by preference. Each
+   must be a valid URL that addresses an RPC endpoint. The respondent may
+   return it empty, reordered, with less, the same, or even more conformant URLs
+   than received.
+2. `rpcDocuments` is an array of zero or more URLs of machine-readable RPC
+   documents that the caller would prefer the respondent to use, ordered by
+   preference. This set of document collectively defines at least syntactically
+   if not also semantically any methods and/or notifications authorized by the
+   CAIP-25 authorization, in DESCENDING heirarchical authority. (For example,
+   any methods or notifications defined differently by multiple authorites will
+   be interpreted by whichever authority is closer to the 0-index of the array).
+   Each must be a valid URL that addresses a valid openRPC document. The
+   respondent may return it empty, reordered, with less, the same, or even more
+   conformant URLs than received.
+
+### Request
+
+A CAIP-207 request is a valid CAIP-25 except for the two additional properties. 
+
+Example:
+
+```jsonc
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "method": "provider_authorization",
+  "params": {
+    "requiredScopes": {
+      "eip155": {
+        "chains": ["eip155:1", "eip155:137"],
+        "methods": ["eth_sendTransaction", "eth_signTransaction", "eth_sign", "get_balance", "personal_sign"],
+        "notifications": ["accountsChanged", "chainChanged"]
+      },
+      "eip155:42069": {
+        "methods": ["get_balance", "chainChanged", "42069_sEcReTbAlAnCe"],
+        "rpcDocuments": ["https://openrpc.42069-chain.org/"],
+        "rpcEndpoints": ["https://node1.42069-chain.org/"]
+      },
+      "cosmos": {
+        ...
+      }
+    },
+    "optionalScopes":{
+      "eip155:42161": {
+        "methods": ["eth_sendTransaction", "eth_signTransaction", "get_balance", "personal_sign"],
+        "events": ["accountsChanged", "chainChanged"]
+    },
+    "sessionProperties": {
+      "expiry": "2022-12-24T17:07:31+00:00",
+      "caip154-mandatory": "true"
+    }         
+  }
+}
+```
+
+### Response
+
+The wallet can respond to this method with either a success result or an error message.
+
+#### Success
+
+A succesful result will be a conformant successful result according to
+[CAIP-25][], with the additional presence (if authorized) of the new properties
+in one or more scope objects.
+
+An example of a successful response follows:
+
+```jsonc
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "result": {
+    "sessionId": "0xdeadbeef",
+    "sessionScopes": {
+      "eip155": {
+        "chains": ["eip155:1", "eip155:137"],
+        "methods": ["eth_sendTransaction", "eth_signTransaction", "get_balance", "eth_sign", "personal_sign"]
+        "events": ["accountsChanged", "chainChanged"],
+        "accounts": ["eip155:1:0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb", "eip155:137:0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb"]
+      },
+      "eip155:10": {
+        "methods": ["get_balance"],
+        "events": ["accountsChanged", "chainChanged"],
+        "accounts:" []
+      },
+      "eip155:42069": {
+        "methods": ["get_balance", "chainChanged", "42069_sEcReTbAlAnCe"],
+        "rpcDocuments": ["https://openrpc.42069-chain.org/"],
+        "rpcEndpoints": ["https://node1.42069-chain.org/"]
+      }
+        //...
+    },      
+    "sessionProperties": {
+      "expiry": "2022-11-31T17:07:31+00:00"          
+    }
+  }
+}
+```
+
+#### Failure States
+
+##### TODO: 
+
+An example of an error response should match the following format:
+
+```jsonc
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "error": {
+    "code": 5000,
+    "message": "Unknown error"
+  }
+}
+```
+
+The additional error messages codes to be supported in addition to CAIP-25s are the following:
+
+##### TODO: replace following with novel errors
+
+Possible error messages (to be discussed with WG):
+- rpcDocuments not conformant syntactically (not openRPC, not served as mime type JSON, etc)
+
+Are these error messages required at protocol level or are they implementation-specific?
+- rpcDocuments rejected by user input
+- rpcDocuments rejected by policy/in principle
+- rpcDocuments unreachable/404
+- rpcEndpoints rejected by user input
+- rpcEndpoints rejected by policy/in principle
+- rpcEndpoints unreachable/404
+- rpcEndpoints URL malformed
+
+* Unknown error OR no requested scopes were authorized
+    * code = 5000
+    * message = "Unknown error"
+* When user disapproves accepting calls with the request methods
+    * code = 5001
+    * message = "User disapproved requested methods"
+* When user disapproves accepting calls with the request events
+    * code = 5002
+    * message = "User disapproved requested events"
+* When wallet evaluates requested chains to not be supported
+    * code = 5100
+    * message = "Requested chains are not supported"
+* When wallet evaluates requested methods to not be supported
+    * code = 5101
+    * message = "Requested methods are not supported"
+* When wallet evaluates requested events to not be supported
+    * code = 5102
+    * message = "Requested events are not supported"
+* When a badly-formed request includes a `chainId` mismatched to scope
+    * code = 5103
+    * message = "Scope/chain mismatch"
+* When a badly-formed request defines one `chainId` two ways
+    * code = 5104
+    * message = "ChainId defined in two different scopes"  
+* Invalid Session Properties Object
+    * code = 5200
+    * message = "Invalid Session Properties requested"
+* Session Properties requested outside of Session Properties Object 
+    * code = 5201
+    * message = "Session Properties can only be optional and global"
+
+## Security Considerations
+
+##### TODO
+- what happens if wallet reorders Documents and dapp ignores that? should we
+  make more explicit how important the ordering of those arrays is, and that
+  neither party can guarantee how the other is ordering them?
+
+## Privacy Considerations
+
+##### TODO
+
+## Changelog
+
+## Links
+
+- [CAIP-2][] - Chain ID Specification
+- [CAIP-10][] - Account ID Specification
+- [CAIP-25][] - JSON-RPC Provider Request
+- [CAIP-75][] - Blockchain Reference for the Hedera namespace
+- [CAIP-171][] - Session Identifier Specification
+
+[CAIP-2]: https://chainagnostic.org/CAIPs/caip-2
+[CAIP-10]: https://chainagnostic.org/CAIPs/caip-10
+[CAIP-25]: https://chainagnostic.org/CAIPs/caip-25
+[CAIP-75]: https://chainagnostic.org/CAIPs/caip-75
+[CAIP-104]: https://chainagnostic.org/CAIPs/caip-104
+[CAIP-171]: https://chainagnostic.org/CAIPs/caip-171
+[namespaces]: https://namespaces.chainagnostic.org
+[RFC3339]: https://datatracker.ietf.org/doc/html/rfc3339#section-5.6
+[CAIP-170]: https://chainagnostic.org/CAIPs/caip-170
+
+## Copyright
+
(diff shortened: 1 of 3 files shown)


CAIPs/caip-207.md
```diff
@@ -0,0 +1,236 @@
...
+CAIP-207 extends CAIP-25 to allow callers and respondents to negotiate RPC
+authorities and RPC semantics before authorizing RPC terms. 
+
+## Abstract
+
+CAIP-207 defines additional properties that CAIP-25 can be used
+iteratively/progressively to layer custom or local RPC semantics and/or routing
+onto a session. Since CAIP-25 respondents ignore unknown properties, respondents
+that conform to CAIP-25 but not to CAIP-207 should be carefully considered and
+accomodated by implementers.
```
kdenhartog, line 24: I think these two could be combined into one subsection called abstract that reads like the following:

```
CAIP-207 extends CAIP-25 to allow callers and respondents to negotiate RPC
authorities and RPC semantics before authorizing RPC terms. This can be done 
progressively to layer custom or local RPC semantics and/or routing onto a session. 
Since CAIP-25 respondents ignore unknown properties, respondents that conform 
to CAIP-25 but not to CAIP-207 should be carefully considered and accommodated
by implementers. 
```

CAIPs/caip-207.md
```diff
@@ -0,0 +1,236 @@
...
+
+CAIP-207 defines additional properties that CAIP-25 can be used
+iteratively/progressively to layer custom or local RPC semantics and/or routing
+onto a session. Since CAIP-25 respondents ignore unknown properties, respondents
+that conform to CAIP-25 but not to CAIP-207 should be carefully considered and
+accomodated by implementers. 
+
+## Motivation
+
+While some core methods and notifications are foundational to a namespace and
```
kdenhartog, line 28: Are we still using the term "namespace"?

CAIPs/caip-207.md
```diff
@@ -0,0 +1,236 @@
...
+
+Two properties are added to the scope objects requested in `optionalScopes` and
+optionally returned in `sessionScopes`.  These are both [ordered] strings of arrays:
+
+1. `rpcEndpoints` is an array of zero or more URLs of RPC endpoints that
+   the caller would prefer the respondent to use, ordered by preference. Each
+   must be a valid URL that addresses an RPC endpoint. The respondent may
+   return it empty, reordered, with less, the same, or even more conformant URLs
+   than received.
+2. `rpcDocuments` is an array of zero or more URLs of machine-readable RPC
```
kdenhartog, line 44: What's the expected behavior if zero `rpcEndpoints` and zero `rpcDocuments` are provided?

CAIPs/caip-207.md
```diff
@@ -0,0 +1,236 @@
...
+      "expiry": "2022-12-24T17:07:31+00:00",
+      "caip154-mandatory": "true"
+    }         
+  }
+}
+```
+
+### Response
+
+The wallet can respond to this method with either a success result or an error message.
```
kdenhartog, line 97: We should probably be defining a test suite for this to make sure implementations are actually conformant. I know this is just a first cut at this but between CAIP-25 and CAIP-207 I see a lot of underspecified states for the protocol occurring between the RPC responder and the RPC requester.

CAIPs/caip-207.md
bumblefudge: I think we're attached to the two sections both existing but i've incorporated the wordsmithing, thanks!

CAIPs/caip-207.md
bumblefudge: heh, the reason we moved to `scope objects` for CAIP-25 namespaces is to disambiguate them from folders in the /namespaces/  repo, which we ARE still using for the time being 🙈

CAIPs/caip-207.md
bumblefudge: couldn't agree more.  I think test vectors might be easier to write into each profile, tho-- i.e.  `/namespaces/eip155/caip211.md#test-vectors`.  Perhaps namespace profiles per-CAIP should have a ##Test-Vectors section instead of an ## Examples section more generally?

CAIPs/caip-207.md
bumblefudge: I threw together a straw man in a new commit but we should probably iterate this a bit, [it's kinda clunky as-is](https://github.com/ChainAgnostic/CAIPs/pull/211/commits/8703b014734361eccf3bb41cfbd9283b641f6f3a)

shanejonas:
for a wallet, is `rpcEndpoints` enough? 

for example [`wallet_addEthereumChain` has `nativeCurrency`](https://eips.ethereum.org/EIPS/eip-3085) so that it can pull out what the name, symbol, decimals to display its currency within the wallet.

CAIPs/caip-211.md
```diff
@@ -0,0 +1,275 @@
...
+  "params": {
+    "requiredScopes": {
+      "eip155": {
+        "chains": ["eip155:1", "eip155:137"],
+        "methods": ["eth_sendTransaction", "eth_signTransaction", "eth_sign", "get_balance", "personal_sign"],
+        "notifications": ["accountsChanged", "chainChanged"]
+      },
+      "eip155:42069": {
+        "methods": ["get_balance", "chainChanged", "42069_sEcReTbAlAnCe"],
+        "rpcDocuments": ["https://openrpc.42069-chain.org/", "https://ethereum.github.io/execution-apis/api-documentation/"],
```
shanejonas, line 93: this would not point to a documentation site, which is what https://ethereum.github.io/execution-apis/api-documentation/ is.

it would point to the actual openrpc document which is located here: https://raw.githubusercontent.com/ethereum/execution-apis/assembled-spec/refs-openrpc.json

bumblefudge: good catch!  curious that `https://ethereum.github.io/execution-apis/assembled-spec/refs-openrpc.json` doesn't work-- would it make sense to configure the gitpages to publish to, e.g., docs.ethereum.org so that the link would survive a migration off of (Microsoft-owned) infra?

(I'll update it with the accurate raw.githubusercontent.com URL for now regardless)

bumblefudge, line 93:
```suggestion
        "rpcDocuments": ["https://openrpc.42069-chain.org/", "https://raw.githubusercontent.com/ethereum/execution-apis/assembled-spec/refs-openrpc.json"],
```

CAIPs/caip-211.md
```diff
@@ -0,0 +1,275 @@
...
+        "accounts": ["eip155:1:0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb", "eip155:137:0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb"]
+      },
+      "eip155:10": {
+        "methods": ["get_balance"],
+        "events": ["accountsChanged", "chainChanged"],
+        "accounts:" []
+      },
+      "eip155:42069": {
+        "methods": ["get_balance", "chainChanged", "42069_sEcReTbAlAnCe"],
+        "rpcDocuments": ["https://ethereum.github.io/execution-apis/api-documentation/", "https://openrpc.42069-chain.org/"],
```
bumblefudge, line 145:
```suggestion
        "rpcDocuments": ["https://raw.githubusercontent.com/ethereum/execution-apis/assembled-spec/refs-openrpc.json", "https://openrpc.42069-chain.org/"],
```

bumblefudge:
> for a wallet, is `rpcEndpoints` enough?
> 
> for example [`wallet_addEthereumChain` has `nativeCurrency`](https://eips.ethereum.org/EIPS/eip-3085) so that it can pull out what the name, symbol, decimals to display its currency within the wallet.

Hey @shanejonas!  Speaking only for myself, I would tent to think that EIP-3085 is a slightly different use-case, since 3085 is for a multi-chain EVM dapp to request/suggest to an EVM wallet adding a chain that the wallet might not already know, a decision which might in turn be passed on to the user over an already-authorized connection.  I think enabling the `wallet_addEthereumChain` method qua method would still work over a connection established by CAIP-25, and would probably be the most logical path in the case where the wallet does not already know/support that chainID.  To put it another way, `rpcEndpoints` isn't enough for adding new [possibly unknown] chains, for that first authorizing and then calling `wallet_addEthereumChain` sounds like a better fit.

Conversely, authorizing additional chains with initial or progressive CAIP-25 authorizations make more sense in cases where the chain is already known to the wallet, particularly if that chain has additional methods/capabilities (like a ZK L2 or a non-EVM/EVM-superset L1 or L2).  For these, CAIP-25 is asking a more fundamental question which is less about user authorization and more about wallet capability.  Here, I should hope rpcEndpoints and/or rpcDocuments *IS* enough, and user consent/interaction is left for a later stage, whether using `wallet_addEthereumChain` or not!

pedrouid:
No objections on my end 👍

hmalik88:
PR should be renamed with [CAIP-211]

CAIPs/caip-211.md
```diff
@@ -0,0 +1,275 @@
...
+CAIP-25 but not to CAIP-211 should be carefully considered and accomodated by
+implementers. 
+
+## Motivation
+
+While some core methods and notifications are foundational to entire namespaces
+and thus almost universally defined out-of-band (meaning all callers and
+respondents agree to them already), others are specific to chains or even to
+subsets of wallets and dapps on a given chain.  This requires scope objects in
+CAIP-25 requests to negotation semantic anchors and/or network routing
```
hmalik88, line 33:
```suggestion
CAIP-25 requests to negotate semantic anchors and/or network routing
```

CAIPs/caip-211.md
```diff
@@ -0,0 +1,275 @@
...
+
+Two properties are added to the scope objects requested in `optionalScopes` and
+optionally returned in `sessionScopes`.  These are both [ordered] strings of arrays:
+
+1. `rpcEndpoints` is an array of zero or more URLs of RPC endpoints that
+   the caller would prefer the respondent to use, ordered by preference. Each
+   must be a valid URL that addresses an RPC endpoint. The respondent may
+   return it empty, reordered, with less, the same, or even more conformant URLs
+   than received.
+2. `rpcDocuments` is an array of zero or more URLs of machine-readable RPC
```
hmalik88, line 62:
```suggestion
2. `rpcDocuments` is an array of zero or more URLs of machine-readable OpenRPC
```

CAIPs/caip-211.md
```diff
@@ -0,0 +1,275 @@
...
+but empty in a request signals that a caller is requesting that custom endpoints
+or  definitions NOT be considered in CAIP-25 authorizations. Present but empty
+arrays in a response confirms this behavior, while its absence can be taken to
+mean the wallet either does not support custom endpoints/definitions or the user
+has opted out of that degree of trust, which for privacy/fingerprinting reasons,
+should not be handled by distinct codepaths (See Privacy Considerations).
+
+## Specification
+
+Two properties are added to the scope objects requested in `optionalScopes` and
```
hmalik88, line 54: These properties need to be allowed in `requiredScopes` as well, doesn't make sense to have them in one and not the other.

CAIPs/caip-211.md
```diff
@@ -0,0 +1,275 @@
...
+   must be a valid URL that addresses an RPC endpoint. The respondent may
+   return it empty, reordered, with less, the same, or even more conformant URLs
+   than received.
+2. `rpcDocuments` is an array of zero or more URLs of machine-readable RPC
+   documents that the caller would prefer the respondent to use, ordered by
+   preference. This set of document collectively defines at least syntactically
+   if not also semantically any methods and/or notifications authorized by the
+   CAIP-25 authorization, in DESCENDING heirarchical authority. (For example,
+   any methods or notifications defined differently by multiple authorites will
+   be interpreted by whichever authority is closer to the 0-index of the array).
```
hmalik88, line 68: This is not entirely clear to me, is it possible to re-word this?

CAIPs/caip-211.md
```diff
@@ -0,0 +1,275 @@
...
+```
+
+### Response
+
+The wallet can respond to this method with either a success result or an error message.
+
+#### Success
+
+A succesful result will be a conformant successful result according to
+[CAIP-25][], with the additional presence (if authorized) of the new properties
```
hmalik88, line 120: I don't know if it makes sense to include the new properties in the response, what use it to spit back what the dapp has requested from you? The OpenRPC doc would be just used by the wallet for discovery.

CAIPs/caip-211.md
```diff
@@ -0,0 +1,275 @@
...
+  }
+}
+```
+
+The additional error messages codes to be supported in addition to CAIP-25s are the following:
+
+##### TODO: replace following with novel errors
+
+Possible error messages (to be discussed with WG):
+- rpcDocuments not conformant syntactically (not openRPC, not served as mime type JSON, etc)
```
hmalik88, line 188: There's an OpenRPC doc to describe the OpenRPC spec, that would be helpful 😄

CAIPs/caip-25.md
```diff
@@ -93,20 +130,28 @@ Example:
...
 - events - array of JSON-RPC message/events expected to be emitted during the
   session
 
-The `requiredNamespaces` array MUST contain 1 or more of these objects, if present; the `optionalNamespaces` array MUST contain 1 or more of them, if 
-present.
+The `requiredScopes` array MUST contain 1 or more of these objects, if present;
+the `optionalScopes` array MUST contain 1 or more of them, if present.
+
+There is one special-case scope object, named `wallet`, which refers to methods
+and events independent of any namespace but, crucially, no chains or CASA
```
hmalik88, line 149: This essentially blocks wallet methods that are chain-related then no?

CAIPs/caip-25.md
```diff
@@ -127,24 +172,25 @@ The wallet can respond to this method with either a success result or an error m
...
-properties requested in the response or not (at the discretion of the provider) but MUST
-conform to the properties names and value constraints described in [CAIP-170][]; any other 
-MUST be dropped by the requester.
+The first is called `sessionScopes` and contains 1 or more scope objects.
+* All required scope objects and all, none, or some of the optional scope object
+(at the discretion of the provider) MUST be included if successful.  
+* As in the request, each scope object object MUST contain `methods` and
+`events` objects, and a `chains` object if a specific chain is not specified in
+the object's index.
+* Unlike the request, each scope object MUST also contain an `accounts` array,
```
hmalik88, line 184:
```suggestion
* Unlike the request, each scope object apart from `wallet` MUST also contain an `accounts` array,
```

hmalik88, line 184: Can we reword this to make it clear that accounts is not a MUST for `wallets` and that it's a MAY?

CAIPs/caip-211.md
bumblefudge: That's fair - I hadn't really been considering that use case, but there will probably be dapps that you can't even (or wouldn't even want to) connect your wallet to without mutually-supported custom RPC, right? I thought the examples should only have them in `optionalScopes` because that will be the most common pattern.

CAIPs/caip-211.md
bumblefudge: my thinking was that the ORDERING of the array might be changed by the wallet (i.e., I'll trust this RPC extension doc, but wherever it redefines something in my core doc, e.g. execution-apis, I'll take the latter's definition). This allows for some flexibility between binary accepting/rejecting each element in the array.

CAIPs/caip-211.md
bumblefudge: Sure! The error message could even LINK to it (some WC error messages currently link to CAIP URLs!) although I'm not sure where to draw the normativity line on things like error messages

CAIPs/caip-25.md
bumblefudge: no, `wallet_changeEthereumChain` would still be a valid method, but it would be in an `eip155` or `eip155:{chainId}` scope object!  the semantics/naming gets a little wonky but hey no one was thinking about non-ethereum wallets or offchain capabilities when the ethereum RPC methods were being named :D

CAIPs/caip-25.md
bumblefudge: good catch, done

CAIPs/caip-211.md
hmalik88: On further thought, it might make sense to include the `rpcEndpoints` field in an ordered manner like you mentioned for `rpcDocuments`. If a provider so chooses to use the provided endpoints, it can specify usage + fallbacks in order or otherwise return an empty array (if its using endpoints of its own choice). 

Ideally,  there should only be one source of truth for the `rpcDocument`, lets call it that because well it should be one. Imagine a scenario where for some reason the multiple endpoints are not updated to be in line with each other, we have a scenario where the wallet doesn't know what document to trust. Anytime a request has an OpenRPC doc, the provider should be using that to determine/provide functionality, including for core-apis.

bumblefudge: Good point.  I guess I was assuming that multiple RPC docs don't necessarily undermine THE RPC Doc.  

Example: custom endpoints might find it useful to have RPCDocs reinforcing or making more explicit what's custom about them-- whether the custom RPC doc overrides The One is a matter of policy, easily expressed in ordering? Why not let additional RPC docs *extend* THE RPC Doc, like a class?

Maybe that's too footgunny, though.  The real question is whether that ordering mechanism is too prone to human error or too hackable... let's talk about it at the next meeting and see what others in the group have to say, maybe? Feel free to bring Shane or anyone at MM who has opinions and/or topical devrel experience from EthDenver :D

bumblefudge: See also this exchange with ligi, which refers to a discussion with TimDaub about eth clients at the last editorial meeting: 
https://github.com/ChainAgnostic/CAIPs/pull/206#discussion_r1102520837

It seems multiple *only very slightly different* RPC documents might be useful until there is 100% conformance across all eth clients!  I can definitely imagine that on other namespaces (e.g. Polkadot) where chains can customize their runtimes and add crates/pallets, it would be very useful for cross-chain apps to have multiple RPC docs...

bumblefudge: Perhaps namespace-specific profiles of 211 would be the place to recommend (or even constrain) behavior here, like "never authorize more than 1 RPC Doc" or "Never put any RPC ahead of The One when authorizing"?

CAIPs/caip-211.md
bumblefudge: reworded in [newest commit](https://github.com/ChainAgnostic/CAIPs/pull/211/commits/6748a7316ce6615f5bdd22a499819b8dde4b3f0e)

CAIPs/caip-211.md
bumblefudge: fixed [here](https://github.com/ChainAgnostic/CAIPs/pull/211/commits/f91abb189234dc4ae0733dfcfb266396ebbfc231)

bumblefudge:
In the interest of simplicity, I'm tempted to just close this in favor of the far less verbose #217 . The example responses illustrating reordering of the arrays might show up again in a future implementation guide or other auxiliary documentation, perhaps even a Informational CAIP?

obstropolos:
Closing this in favor of continued work on #217

Issue #213, opened 2023-02-07
[CAIP-122] Add guideline to match on `domain` term


haardikk21:
In EIP-4361, there is a guideline that wallets "MUST prevent phishing attacks by matching on the domain term". This stance is missing in CAIP-122, and seems reasonable to add.

bumblefudge:
closed by #214

PR #214, opened 2023-02-07
Add wallet implementer considerations

haardikk21:
Solves #213

### DIFF
--- a/CAIPs/caip-122.md
+++ b/CAIPs/caip-122.md
@@ -147,7 +147,7 @@ Not applicable.
 
 ### Verifying `domain` binding
 - Wallet implementers MUST prevent phishing attacks by matching on the `domain` term when processing a signing request. For example, when processing the message beginning with `"service.invalid wants you to sign in..."`, the wallet checks that the request actually originated from `service.invalid`.
-- The domain SHOULD be read from a trusted data source such as the browser window.
+- The domain is expected to be read from a trusted data source such as the browser window.
 
 ## References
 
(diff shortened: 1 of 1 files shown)


CAIPs/caip-122.md
```diff
@@ -143,6 +143,12 @@ c2VydmljZS5vcmcgd2FudHMgeW91IHRvIHNpZ24gaW4gd2l0aCB5b3VyIFNvbGFuYSBhY2NvdW50OgpH
 
 Not applicable.
 
+## Wallet Implementer Steps
+
+### Verifying `domain` binding
+- Wallet implementers MUST prevent phishing attacks by matching on the `domain` term when processing a signing request. For example, when processing the message beginning with `"service.invalid wants you to sign in..."`, the wallet checks that the request actually originated from `service.invalid`.
+- The domain SHOULD be read from a trusted data source such as the browser window.
```
kdenhartog, line 150:
```suggestion
- The domain is expected to be read from a trusted data source such as the browser window.
```

I don't believe this is a testable statement so removing the normative statement here, but I generally think it's a good idea to state this. I've found that reserving normative statements for things that can be tested (e.g. put in a test suite) makes for a better way to be able to quickly assess whether an implementation is conformant with the spec via a test suite. So since this doesn't appear to be testable, I don't think a normative statement here is good.

kdenhartog (changes requested):

bumblefudge (approved):
LGTM

kdenhartog (approved):

PR #217, opened 2023-03-02
Decouple CAIP-27 from CAIP-25 and introduce CAIP-217 for Authorization Scopes

bumblefudge:
To-do before merge
- [X] Overhauled to reorg CAIP-25 into composable blocks 💪 
- [X] Closed #211 - no longer needed, `rpcEndpoints` and `rpcDocuments` are just optional properties now 😌 
- [x] Introduced ### Validation section - should this be more constrained? feels a little #yolo as-is
- [x] ### Response section feels too open-ended - should forwarded responses be wrapped in metadata? 
- [x] Should error messages be added?
- [x] Should references to CAIP-170 be removed, for now, as #170 is still in draft and there's been no discussion on it?

@hmalik88 @pedrouid please opine on the questions above and I can throw some more commits in this bad boy


CAIPs/caip-25.md
```diff
@@ -92,21 +122,21 @@ Example:
 }
 ```
 
-The JSON-RPC method is labelled as `provider_authorization` and both the
-"requiredNamespaces" and "optionalNamespaces" arrays are populated with 
-`namespace` objects each named after the scope of authorization:
-1. EITHER an entire ChainAgnostic [namespace][] 
-2. OR a specific [CAIP-2][] in that namespace.
+The JSON-RPC method is labelled as `provider_authorize` and its `params` object
```
hmalik88, line 125:
```suggestion
The JSON-RPC method is labeled as `provider_authorize` and its `params` object
```

CAIPs/caip-27.md
```diff
@@ -44,8 +45,9 @@ The application would interface with a provider to make request as follows:
   "jsonrpc": "2.0",
   "method": "caip_request",
   "params": {
-    "chainId": "eip155:1",
     "session": "0xdeadbeef",
+    "scope": "eip155",
+    "chainId": "eip155:1",
```
hmalik88, line 50: I thought we were removing this? It's redundant and doesn't fit the non-blockchain use cases that we spoke about. I think `session` followed with `scope` is sufficient.

CAIPs/caip-27.md
```diff
@@ -57,15 +59,23 @@ The application would interface with a provider to make request as follows:
 }
 ```
 
-The JSON-RPC method is labelled as `caip_request` and expects three parameters:
+The JSON-RPC method is labelled as `caip_request` and expects
```
hmalik88, line 62:
```suggestion
The JSON-RPC method is labeled as `caip_request` and expects 
```

CAIPs/caip-27.md
```diff
@@ -57,15 +59,23 @@ The application would interface with a provider to make request as follows:
...
-  specific chain or network within it
-- session - [CAIP-171][] `SessionToken` to identify the session opened or
+- **session** - [CAIP-171][] `SessionToken` to identify the session opened or
   updated by a [CAIP-25][] interaction.
-- request - an object containing the fields:
-  - method - JSON-RPC method to request
-  - params - JSON-RPC parameters to request
+- **scope** - a `scopeObject` authorized by a [CAIP-25][] response and persisted in
+  the session by both caller and respondent
+- *chainId* - [CAIP-2][]-defined `chainId` including both namespace and a
```
hmalik88, line 70: I think we should remove this and update the scope definition to be either a non-blockchain scope or a CAIP-2 `chainId`, because the request will have to either map to a specific chain or not, never just a namespace.

CAIPs/caip-27.md
bumblefudge: that's why I made it optional--- but then I thought about the corner case where "eip155" is the key of the scope object with multiple entries in the `chains` array, or for that matter any other namespace where (unlike polkadot or cosmos) a chainId is needed to route each RPC call...

Inelegant but fits CAIP-25's scopeObjects snugly.

CAIPs/caip-27.md
bumblefudge: Not exactly never-- the `wallet_` namespace within eip155 and many methods in polkadot and cosmos ARE namespace-wide!

hmalik88: ok fair, but you still wouldn't need the `chainId`, the `scope` would then just be the namespace.

CAIPs/caip-27.md
hmalik88: See my other comment.

hmalik88 (approved):
LGTM.

bumblefudge (commented):
LGTM, @pedrouid final check?

bumblefudge:
Majorly refactored; new initial comment above

CAIPs/caip-27.md
```diff
@@ -1,51 +1,57 @@
...
 ### Request
 
-The application would interface with a provider to make request as follows:
+The application would interface with an RPC provider to make request as follows:
 
 ```jsonc
 {
   "id": 1,
   "jsonrpc": "2.0",
   "method": "caip_request",
```
pedrouid, line 51:
```suggestion
  "method": "provider_request",
```

CAIPs/caip-27.md
```diff
@@ -57,20 +63,30 @@ The application would interface with a provider to make request as follows:
 }
 ```
 
-The JSON-RPC method is labelled as `caip_request` and expects three parameters:
+The JSON-RPC method is labeled as `caip_request` and expects
```
pedrouid, line 66:
```suggestion
The JSON-RPC method is labeled as `provider_request` and expects 
```

pedrouid (commented):
Just left a single comment to replace in `caip_request` with `provider_request` for CAIP-27

Other than that... This is looking amazing! 🙏 

I can feel CAIP-25 and CAIP-27 being Final very very soon 🚀

CAIPs/caip-217.md
```diff
@@ -0,0 +1,91 @@
...
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" written in
+uppercase in this document are to be interpreted as described in [RFC
+2119](https://www.ietf.org/rfc/rfc2119.txt)
+
+### Definition
+
+The syntax is as follows:
+
+```json
```
pedrouid, line 47: This removes the "awkward" red highlighting that Github puts on this code block

```suggestion
```jsonc
```

CAIPs/caip-27.md
```diff
@@ -57,20 +63,30 @@ The application would interface with a provider to make request as follows:
...
+three **required parameters**:
 
-- chainId - [CAIP-2][]-defined `chainId` to identify both a namespace and a
-  specific chain or network within it
-- session - [CAIP-171][] `SessionToken` to identify the session opened or
-  updated by a [CAIP-25][] interaction.
-- request - an object containing the fields:
-  - method - JSON-RPC method to request
-  - params - JSON-RPC parameters to request
+- **session** - [CAIP-171][] `SessionId` referencing a known, open session
```
pedrouid, line 69:
```suggestion
- **sessionId** - [CAIP-171][] `SessionId` referencing a known, open session
```

pedrouid:
Some extra comments to fix how markdown is being rendered plus making it more consistent with CAIP-25 naming

Akaslove99:
@pedrouid #
Duplicate of #

Akaslove99:
```html

``` 
```python

```

Akaslove99 (approved):

CAIPs/caip-217.md
```diff
@@ -0,0 +1,91 @@
...
+- `string` (conditional) = A namespace identifier string registered in the CASA [namespaces][] registry to authorize multiple chains with identical properties OR a single, valid [CAIP-2][] identifier, i.e., a specific `chain_id` within a namespace.
+- `scopes` (conditional) = An array of 0 or more [CAIP-2][] `chainId`s. Setting multiple `scopes` is equivalent to making multiple copies of the `scopeObject` for each member of `scopes`, simply in compact form.
+  - This property MUST NOT be present if the object is already scoped to a single `chainId` in the string value above.
+  - This property MUST NOT be present if the scope is an entire [namespace][namespaces] in which `chainId`s are not defined.
+  - This property MAY be present if the scope is an entire [namespace][namespaces] in which `chainId`s are defined.
+- `methods` = An array of 0 or more JSON-RPC methods that an application can call on the agent and/or an agent can call on an application.
+- `notifications` = An array of 0 or more JSON-RPC notifications that an application send to or expect from the agent.
+- `accounts` (optional) = An array of 0 or more [CAIP-10][] identifiers valid within the scope of authorization.
```
hmalik88, line 70: What's the context here for providing `accounts`? Does this pertain to the request or response?

bumblefudge: it's optional because usually empty or unset in the CAIP-25 request and set by responder in most CAIP-25 contexts today... but in theory this data model could be used independently of CAIP-25 so maybe left vague here?

CAIPs/caip-217.md
```diff
@@ -0,0 +1,99 @@
...
+- `scopes` (conditional) = An array of 0 or more [CAIP-2][] `chainId`s. Setting multiple `scopes` is equivalent to making multiple copies of the `scopeObject` for each member of `scopes`, simply in compact form.
+  - This property MUST NOT be present if the object is already scoped to a single `chainId` in the string value above.
+  - This property MUST NOT be present if the scope is an entire [namespace][namespaces] in which `chainId`s are not defined.
+  - This property MAY be present if the scope is an entire [namespace][namespaces] in which `chainId`s are defined.
+- `methods` = An array of 0 or more JSON-RPC methods that an application can call on the agent and/or an agent can call on an application.
+- `notifications` = An array of 0 or more JSON-RPC notifications that an application send to or expect from the agent.
+- `accounts` (optional) = An array of 0 or more [CAIP-10][] identifiers, each valid within the scope of authorization.
```
hmalik88, line 74: How are we addressing account matching for namespaced scopes? Also shouldn't this be 1 or more CAIP-10 identifiers (in the response)?

CAIPs/caip-211.md
```diff
@@ -18,7 +18,12 @@ within one) and RPC endpoints (i.e. preferential routing for specific nodes).
 
 ## Abstract
 
-
+Without a profile of this CAIP which defines implicit values for a namespace,
+setting the `rpcDocuments` and `rpcEndpoints` values in the `scopeObject`s of a
+[CAIP-25][] negotiation does not make much sense. Once those implicit values
```
hmalik88, line 23: I would argue this to not be true, because a rpc document could be introducing new methods and not necessarily overriding implicit values.

bumblefudge: Good point--- I'm trying to telegraph that this introduces potential security risks (since an explicit one COULD override an implicit one) so using this property is a little riskier until/unless a profile has been written... will rewrite in the meantime, in any case

CAIPs/caip-217.md
bumblefudge: 1 or more in eip155, but not necessarily in all namespaces (for example, the hypothetical `wallet` namespace!). I think we'd need to merge this and then open a PR to discuss in two weeks adding `eip155/caip25.md` to namespaces which adds the "always 1 account in every `sessionScopes` member keyed to eip155 or to an eip155 CAIP-2..." rule :D

hmalik88: Ok, I'm fine with keeping it at 0 or more, but we do need to include some sort of language and logic around account matching, we talked about how the namespace scope was a shorthand for chain specific scopes and how the `accounts` logic doesn't match up with that for that exact reason. What if a chain returns more than one account, what if a chain doesn't return any accounts? We have no idea how to match based on an array. I think omitting this leaves this CAIP incomplete especially when it's introducing this concept of `accounts`. We're making the incorrect assumption that namespaces share accounts with the current spec.

CAIPs/caip-211.md
```diff
@@ -0,0 +1,151 @@
...
+    notifications: [D, E, F],
+    rpcEndpoints: [Y1, Y2, Y3, U1, U2, U3],
+    rpcDocuments: [Z, V]
+  }
+}
+```
+
+But the latter SHOULD be returned as response if dropping the implicit values in
+favor of the requested ones is unacceptable and extending the implicit values is
+preferred.
```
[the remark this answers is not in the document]
bumblefudge: Unclear # 1: by unacceptable, I meant "allowed to extend e.g. execution APIs but I neither want to allow your documents to contradict/re-define execution APIs NOR do I want to manually confirm each connection that your documents don't contradict/redefine.  I can either make this more explicit in another commit... or change course if it's a faulty use-case assumption!

Unclear # 2: it's extraneous if it's in every request, in the sense that it verbosely states the obvious, but I wouldn't drop it from my responses if I was a provider! what a CAIP-211 profile does is give additional confidence to the providers who want to interpret no setting as equivalent to execution-apis, and let them respond without adding it.  I guess I was assuming legacy verbosity would be tolerable and just peter out over time harmlessly?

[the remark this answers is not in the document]
bumblefudge: unclear # a: I was assuming whole RPC document -- if machine-readable, "subsetting" that document or extracting, say, ABI signatures for function names can happen automatically, if human-readable (a permalink to a namespace's official dev docs, for example), this might be a little more onerous...

CAIPs/caip-217.md
bumblefudge: right-- I think maybe the only implicit "matching" logic was the requirement that `accounts` are "valid" in the given scope of a `scopeObject`, but you're describing a corner-case of that squishy "validity". For example:

```
"eip155": {
...
   "scopes": ["chainid1", "chainid2", "chainid3"],
   "accounts"["eip155:chainid1:acct1", "eip155:chainid2:acct2", "eip155:chainid3:acct3"]
...
}
```
all three accounts are valid in `eip155` but `acct3` is valid only in `chainid3`, so if the `scopes` array is actually a shorthand for condensing 3 distinct caip2-specific scopes, all 3 of them have 2 invalid accounts in their `accounts` array, perhaps?

I think maybe we're trying to solve an eip155 problem at the CAIP layer, and the desired matching behavior for this corner-case might best be specified by the `eip155/caip25.md` profile? Or do we need to expand the concept (or the syntax) of `accounts` for all namespaces?  Even if namespaces like polkadot can "convert" a universal address to a CAIP2-specific address and thus sidestep the problem of matching `accounts` to `scopes`, Solana and some other namespaces have an additional matching problem, where each address exposed needs to be matched to a public key...

bumblefudge: So can we merge this for the sake of updating `main` (and the .org website) and return to this issue later, after we've tried writing the `eip155/caip25.md` profile? It might be worth detailing how to handle "invalid" `accounts` (drop? ignore? drop-or-ignore?) but it's hard to specify until we have an "end to end example" , i.e. a stable CAIP-25 and a profile defining validity in namespace-specific ways :D

CAIPs/caip-211.md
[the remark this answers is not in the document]
bumblefudge: > This should just be a rejected request then.

OK I'll push another commit aligned with this different behavior.

CAIPs/caip-217.md
hmalik88: yes, i'm fine with leaving this part as is.

CAIPs/caip-217.md
```diff
@@ -0,0 +1,99 @@
...
+   notifications: [(notification_name)+],
+   *accounts: [(account_id)+]
+   *rpcDocuments: [(rpcDocument)+],
+   *rpcEndpoints: [(rpcEndpoint)+]
+}
+```
+
+Where:
+
+- `string` (conditional) = A namespace identifier string registered in the CASA [namespaces][] registry to authorize multiple chains with identical properties OR a single, valid [CAIP-2][] identifier, i.e., a specific `chain_id` within a namespace.
```
rekmarks, line 67: What is `string` referring to here?

hmalik88: Looks like a typo, should be referring to `scope`

cc: @bumblefudge

CAIPs/caip-217.md
```diff
@@ -0,0 +1,99 @@
...
+- `scopes` (conditional) = An array of 0 or more [CAIP-2][] `chainId`s. Setting multiple `scopes` is equivalent to making multiple copies of the `scopeObject` for each member of `scopes`, simply in compact form.
```
hmalik88, line 68: let's get rid of this language of setting multiple scopes as being shorthand for multiple copies of the `scopeObject` for each member of `scopes` as we've discussed that's not the case :)

CAIPs/caip-217.md
bumblefudge: sorry, I didn't think through this part very well.  this is the string that summarizes the whole object, i.e. `eip155` or `eip155:1`.  Hassan is right.  Maybe it would be more readable to write:

```
` ` ` jsonc
scopeString: {
   * scopes: ...
}
` ` `
Where:
- `scopeString` = A namespace identifier...
```

bumblefudge: ^ If so, glad to edit accordingly

CAIPs/caip-217.md
bumblefudge: Closing in the hopes that [this commit](https://github.com/ChainAgnostic/CAIPs/pull/217/commits/8a92f2b79f34ac64aaeb0eab6b29b3ce0ca55e80) leaves validity to be defined in profiles of the spec

CAIPs/caip-217.md
bumblefudge: Done [here](https://github.com/ChainAgnostic/CAIPs/pull/217/commits/8a92f2b79f34ac64aaeb0eab6b29b3ce0ca55e80)

CAIPs/caip-217.md
bumblefudge: clarified by edits in [this commit](https://github.com/ChainAgnostic/CAIPs/pull/217/commits/50763733fa4e739af2ceb5aa3cd5d54fce2ac504) and [this commit](https://github.com/ChainAgnostic/CAIPs/pull/217/commits/10378531ea47bfa50b63555aaa8e474d88ad5c84).

Issue #209, opened 2023-01-31
User feedback: No way to address a specific transaction


bumblefudge:
In the past, we've discussed block heights (whether as a new CAIP or as a "query parameter").  A more fundamental use case might be addressing across multiple namespaces a specific transaction, i.e. on both sides of an atomic swap or bridging transaction.  

Another possible use-case, whether for the same or a separate CAIP, would be addressing the **metadata** of a specific transaction, such as transaction type, or even arbitrary metadata like `input data`, method calls, memos, etc.  This could a path (i.e. another `:` followed by, say `parameters[0]` in the case of input data), a query param, or a separate CAIP.  Would like to hear from more implementers before committing to anything that might be un-profilable or un-implementable on many namespaces, tho!

Perhaps related issue: How would we unambiguously address miner's/validators messages in blocks, i.e., `The Times 03/Jan/2009 Chancellor on brink of second bailout for banks.`? `bip122:000000000019d6689c085ae165831e93:666666:op_return` ?

titusz:
Cross chain addressing of transactions should be fairly simple with a pattern like:
`<chain-id>:tx:<tx-hash>`

Examples:
`eip155:1:tx:0x1cc5b5193db6f41c6b830660fe6e8e14a4c9bce66381956f065a5584272754b8`
`bip122:tx:f1d481508ce99ade1427b581ce0e240dd7258358bbfdee89ed08450de23c5e6c`

Another possibility, to make references much shorter, one could take advantage of blockchain immutability and total ordering and use an index/location based addressing scheme:

`<chain-id>:tx:<block-height>:<tx-index>`

The above examples would be reduced to:
`eip155:1:tx:15670909:268`
`bip122:tx:635359:1110`

On UTXO chains you could also reference a specific outputs (like an OP_RETURN within a transaction by index:
`<chain-id>:tx:<block-height>:<tx-index>:<output-index>`

Example:
`bip122:tx:635359:1110:0`

would point to the first output within the transaction (an OP_RETURN) in this case:
https://mempool.space/tx/f1d481508ce99ade1427b581ce0e240dd7258358bbfdee89ed08450de23c5e6c

microbecode:
Interesting. I wasn't aware of this discussion before I started a process for creating a similar [standard for EVM chains](https://ethereum-magicians.org/t/a-new-standard-for-encoding-chain-id-transaction-hash/23782/7).

[...]

Issue #208, opened 2023-01-31
User feedback: No way to address a NON-ASSET deployed smart contract in a namespaced way


bumblefudge:
CAIP-19 only works for assets controlled/registered at a smart contract; non-asset smart contracts are not addressable by it.  It feels like `asset_type` should not be generalized to `contract_type`, particularly since there are already many users and adopters in the wild using it specifically as a translation-layer/interop tool for asset-specific interactions/queries.  

This implies to me that a new, distinct CAIP would , particularly since in many of the alread-registered namespaces assets are a subset of contracts with different validation rules/addressing systems, etc.

Similarly, if someone wanted to address a specific transaction (or better yet, the metadata of a specific transaction, such as transaction type, or even arbitrary metadata like `input data`, method calls, memos, etc.

Perhaps related issue: How would we unambiguously address miner's/validators messages in blocks, i.e., `The Times 03/Jan/2009 Chancellor on brink of second bailout for banks.`?

Shout out to @sposth for raising this

Nightmare77769:
[CAIPs-main.zip](https://github.com/ChainAgnostic/CAIPs/files/13806924/CAIPs-main.zip)
[eip155-82.json](https://github.com/ChainAgnostic/CAIPs/files/13806925/eip155-82.json)

1. [](url)

Nightmare77769:
[Uploading CAIPs-main.zip…]()

Nightmare77769:
[eip155-82.json](https://github.com/ChainAgnostic/CAIPs/files/13806927/eip155-82.json)
[eip155-83.json](https://github.com/ChainAgnostic/CAIPs/files/13806928/eip155-83.json)
[eip155-100 (2).json](https://github.com/ChainAgnostic/CAIPs/files/13806929/eip155-100.2.json)
[eip155-100 [
[Uploading CAIPs-main.zip…]()
]((1).json]https://github.com/ChainAgnostic/CAIPs/files/13806930/eip155-100.1.json))

Nightmare77769:
[Uploading eip155-82.json…]()

Nightmare77769:
[Uploading pm-master.zip…]()

Nightmare77769:
[Uploading tokenlist.schema (1).json…]()
[Uploading tokenlist.schema.json…]()

Nightmare77769:
@

Missjones2829:
M

PR #218, opened 2023-03-04
caip: chainproof

oed:
[📃Preview](https://github.com/oed/CAIPs/blob/caip/chainproof/CAIPs/caip-218.md)

### DIFF
--- a/CAIPs/caip-218.md
+++ b/CAIPs/caip-218.md
@@ -1,0 +1,194 @@
+---
+caip: CAIP-218
+title: ChainProofs - Fresh Facts Straight from the Blockchain
+author: Joel Thorstensson (joel@3box.io), Wayne Chang (wayne@spruceid.com)
+discussions-to: https://github.com/ChainAgnostic/CAIPs/pull/218
+status: Draft
+type: Standard
+created: 2023-03-04
+updated: 2023-03-04
+requires: 196
+---
+
+# TODO BEFORE MERGE
+
+- [ ] Fully define ChainProof EVM Profile mechanisms for identifying minimal states necessary from `eth_getProof` for EVM verification of contract calls and blocks. currently under "Instruction things".
+- [ ] Flesh out ChainProof Method definition and examples, especially the EVM NFT Ownership method.
+- [ ] Example code and test vectors.
+- [ ] Clean up ChainProof Methods section, move EVM-specific stuff to the EVM Profile.
+- [ ] Discussion of ChainProof inclusion within ReCaps, UCANs, and W3C Verifiable Credentials, including their relationship with cryptographic signatures in these formats.
+- [ ] Complete remaining empty sections.
+
+<!--You can leave these HTML comments in your merged EIP and delete the visible duplicate text guides, they will not appear and may be helpful to refer to if you edit it again. This is the suggested template for new EIPs. Note that an EIP number will be assigned by an editor. When opening a pull request to submit your EIP, please use an abbreviated title in the filename, `eip-draft_title_abbrev.md`. The title should be 44 characters or less.-->
+## Simple Summary
+
+<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the CAIP.-->
+ChainProofs are a cross-chain mechanism for proving human understandable statements using evidence from blockchain state.
+
+## Abstract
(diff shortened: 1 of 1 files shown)


oed:
@wyc

bumblefudge:
This is very interesting! The one thing that this spec makes me wonder about is the precision and universality of block-, [transaction](https://github.com/ChainAgnostic/CAIPs/issues/209)- and [non-asset](https://github.com/ChainAgnostic/CAIPs/issues/208)-contract- addressing.  it feels to me from the diagram provided for the EVM profile that there is a `{namespace}:{blockId}:{txnID}:{code/storageslot/etc}` URN implicit here... that we could maybe make explicit first.  Lord knows people have been proposing their use-cases (linked above) for this URN anyways!

Having been mulling on this for months, I think we probably need two new CAIPs, that chainProofs could extend/require:

- `{namespace}:{blockId}:{txnID}{:XOR?}XXX` where XXX is an optional (and namespace-specific) syntax for code, storage addressing (including arbitrary transaction "memo" strings, mentioned in the issues above), etc.
and
- `{namespace}:{blockId}{:XOR?}XXX - because blocks themselves have a lot of metadata.  this might extend Ligi's block-explorer uniform routing CAIP, since obviously what metadata each block has is extremely namespace-specific.

And let's just name one elephant in the room: maybe the Ordinals people would be interested in the above.  There hasn't been a ton of interest from the BTC family tree so far, this might be a good opportunity for outreach to them!

oed:
For EVM [`debug_traceCall`](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#debugtracecall) will likely be useful when trying to find the appropriate storage slots.

wyc:
I think txid-based addressing could be useful, but we don't need it to prove NFT ownership on Ethereum for example, so I'm not sure it should be a requirement. It could be useful if you wanted to prove that a particular transaction was included in a specific block or something like that, but you could alternatively define what that means in the EVM profile without a new namespace at CASA. Also, how would you refer to blocks and transactions with graph-based ledgers?

bumblefudge:
> we don't need it to prove NFT ownership on Ethereum for example, so I'm not sure it should be a requirement. 
Right, EVM and ERC721 make it pretty easy/standardized to navigate to the storage from the contract address.  My question was whether this would be more extensible if there were a CAIP-constrained string that encoded that same navigation for non-EVM contracts/environments.  Didn't mean to imply it was a gating requirement!  At most a stitch-in-time situation, but more likely me overthinking :D 

> how would you refer to blocks and transactions with graph-based ledgers?
I'm no expert but most of the DAGs I've looked at have something pretty similar to transactions, it's more the "block"-addressing that would break down (although most DAGs have some form of supernode/milestone/checkpoint/etc).  I'm actually working on these two CAIPs today, so maybe it'll be less annoying to discuss with a strawman to point to, but we definitely shouldn't bake the assumption of blocks into the syntax for transactions, since basically everything we want a namespace for has transaction-queryable state... I've crossed "block" out of the micro-strawman above!

oed:
Updated the PR with an example chainproof for the ownership of an ENS name. See the CAR file in the `assets` folder.

It doesn't yet use multidid for the `iss/aud`, other than that I think it should be a fully compliant NFT chainproof.

Created using this script: https://gist.github.com/oed/a590fe2932099310bcbf922d1d25b646

oed:
**A note on size:**
The ChainProof above for proving ENS name ownership is actually quite big (21 kB). For more complicated proofs size will be even larger. To make ChainProofs more usable it might actually be worth it to compress them using ZKPs. Experimentation needed!

oed:
Great news, after a conversation with Richard Meissner of Safe it turns out that there's an easy way to get all storage slots associated with an `eth_call`. 

The method is called [`eth_createAccessList`](https://docs.infura.io/networks/ethereum/json-rpc-methods/eth_createaccesslist#:~:text=Creates%20an%20EIP-2930%20access%20list%20that%20you%20can,storage%20keys%20that%20a%20transaction%20intends%20to%20access.) and returns all the storage slots across all contracts that were accessed by a specific call. 

This is all information we need to be able to construct a ChainProof for any EVM smart contract call!

obstropolos:
@oed - doing some quick housekeeping - where did this end up?

PR #220, opened 2023-03-20
New CAIP - Block Addressing model

bumblefudge:

PR #221, opened 2023-03-20
new CAIP - Transaction Object Addressing

bumblefudge:
Design questions still pending input:
- Note: [Titusz' proposal](https://github.com/ChainAgnostic/CAIPs/issues/209#issuecomment-1416002818) was for addressing transactions within blocks, by slot #; here, instead, I was thinking specifically of transactions that can be queried directly from indexers and/or block explorers. 
- [ ] Should transactions-within-blocks be added to #220 ? If so, does that change anything about what's in and out of scope for this CAIP scheme?
- [ ] Block explorers tend to display information like position-in-block, which can't be derived from the transaction object itself; are those worth including in such a scheme? I assume not, particularly if transactions-within-a-block are addressable from #220, as mentioned above. But worth checking with the group
- [ ] Should a property that's an array be queryable (i.e. `..123deadbeef4.inputs[1]`), or should the scheme just allow you to fetch the whole array and you need to dive into it after getting back the whole array? 

Would you all be open to a meeting to discuss, @TimDaub @Titusz @ntn-x2 @sposth ?

### DIFF
--- a/CAIPs/caip-221.md
+++ b/CAIPs/caip-221.md
@@ -37,7 +37,7 @@ may be important to cross-namespace use cases.
 Transactions are addressed as follows:
 
 ```
-block_address:        chain_id + ":" [ + "block:"]? + "txn/" + transaction_id + ["." + property]?
+block_address:        chain_id + ":" [ + "block:"]? + "tx/" + transaction_id + ["." + property]?
 chain_id:             [-a-z0-9]{3,8}:[-_a-zA-Z0-9]{1,32} (See [CAIP-2][])
 transaction_id:       [-%a-zA-Z0-9]{1,128}
 property (optional):  (signer|recipients|type|outputs|inputs)


CAIPs/caip-221.md
```diff
@@ -0,0 +1,107 @@
+---
+caip: CAIP-X
+title: Transaction Object Addressing
+author: Juan Caballero (@bumblefudge), Sebastian Posth <sebastian@posth.com>, and Antonio Antonioni (@ntn-x2)
```
ntn-x2, line 4: 😁

```suggestion
author: Juan Caballero (@bumblefudge), Sebastian Posth <sebastian@posth.com>, and Antonio Antonino (@ntn-x2)
```

CAIPs/caip-221.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Rationale
+<!--The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages. The rationale may also provide evidence of consensus within the community, and should discuss important objections or concerns raised during discussion.-->
+TBD - will just summarize PR debates if good input is forthcoming!
+
+## Test Cases
+<!--Please add diverse test cases here if applicable. Any normative definition of an interface requires test cases to be implementable. -->
+
+```
+# bip122:000000000019d6689c085ae165831e93:666666:tx/057954bb28527ff9c7701c6fd2b7f770163718ded09745da56cc95e7606afe99.data
```
ntn-x2, line 68: `data` is not part of the `property` property.

ntn-x2 (commented):
Don't have much context on the origin of this PR, but overall it looks a good starting point. 

One main grip I have, as I also had for CAIP-19 identifiers for the `erc`-based namespaces, is the possibility to link between identifiers that refer to the same resource. Specifically, I think that a tx referred to by a CAIP of this category and a tx referred to by a CAIP as in https://github.com/ChainAgnostic/CAIPs/pull/220/files should semantically be considered the same resource. I don't see any resolution process defined in these CAIPs, meaning that they only deal (as far as I understand) with identifiers, and not with resolving any aliases between them.

bumblefudge:
> I don't see any resolution process defined in these CAIPs, meaning that they only deal (as far as I understand) with identifiers, and not with resolving any aliases between them.

But wouldn't that have to live in each namespace's profile of this CAIP, since there no common/universal assumptions that can be made even about the resolvability or uniqueness of these identifiers?

XstRUhayabusa (approved):

CAIPs/caip-221.md
```diff
@@ -0,0 +1,107 @@
...
+translation layer across those syntaxes, with per-namespace profiles needed to
+further define and caveat the values and add namespace-specific properties that
+may be important to cross-namespace use cases.
+
+## Specification
+<!--The technical specification should describe the standard in detail. The specification should be detailed enough to allow competing, interoperable implementations. -->
+Transactions are addressed as follows:
+
+```
+block_address:        chain_id + ":" [ + "block:"]? + "txn/" + transaction_id + ["." + property]?
```
friedger:
```suggestion
block_address:        chain_id + ":" [ + "block:"]? + "tx/" + transaction_id + ["." + property]?
```
This is shorter and also used in the test cases below

friedger (commented):
replace txn/ with tx/

PR #222, opened 2023-04-07
CAIP-222: Wallet Authenticate

pedrouid:
Inspired by WalletConnect's Auth API: https://medium.com/walletconnect/introducing-walletconnect-auth-one-click-wallet-login-to-simplify-web3-ux-9cc11fd18aba

The motivation for creating a standard is to enable other wallet providers to offer the same UX without using the WalletConnect protocol.

**Auth = Connect + Sign**

This standard will enable wallets to authenticate blockchain accounts in a single-step without requiring a persistent connection to be established.


CAIPs/caip-222.md
```diff
@@ -0,0 +1,249 @@
+---
+caip: 222
+title: Wallet Authenticate JSON-RPC Method
+author: Pedro Gomes (@pedrouid)
+discussions-to: https://github.com/ChainAgnostic/CAIPs/discussions/222
+status: Draft
+type: Standard
+created: 2022-04-07
```
kdenhartog, line 8:
```suggestion
created: 2023-04-07
```

CAIPs/caip-222.md
```diff
@@ -0,0 +1,249 @@
...
+  }
+}
+```
+
+The JSON-RPC method is labelled as `wallet_authenticate` and expects the following parameters:
+
+- cacaov - Cacao version number
+- type - Cacao header message type
+- chainId - [CAIP-2][]-defined `chainId` to identify specific chain or network.
+- domain - [RFC 4501][rfc 4501] `dnsauthority` that is requesting the signing.
```
kdenhartog, line 68: @awoie we should make sure whatever we land on for SIWE to match here.

@pedrouid the current thinking is that we’ll stay with domain here because that’s what’s already deployed.

In any case, we should make sure the cited RFC matches as well, because IIRC these are different right now.

CAIPs/caip-222.md
```diff
@@ -0,0 +1,249 @@
...
+- aud - [RFC 3986][rfc 3986] URI referring to the resource that is the subject of the signing.
+- version - Current version of the message.
+- nonce - Randomized token to prevent signature replay attacks.
```
kdenhartog, line 71: We should specify the requirements of the nonce here too.

CAIPs/caip-222.md
```diff
@@ -0,0 +1,249 @@
...
+- requestId (optional) - System-specific identifier used to uniquely refer to the authentication request.
+- resources (optional) - List of information or references to information the user wishes to have resolved as part of the authentication by the relying party; express as [RFC 3986][rfc 3986] URIs and separated by `\n`.
+
+### Response
+
+The wallet will prompt the user with a dedicated UI to display the app requesting the authentication and allow the user to select which blockchain account to sign with.
+
+#### Success
+
+If approved, the walllet will return a signed CACAO with the selected blockchain account.
```
kdenhartog, line 85:
```suggestion
If approved, the wallet will return a signed CACAO with the selected blockchain account.
```

CAIPs/caip-222.md
```diff
@@ -0,0 +1,249 @@
...
+      "iss": string,
+      "aud": string,
+      "exp": string,
+      "iat": string,
+      "nbf": string,
+      "nonce": string,
+      "domain": string,
+      "version": string,
+      "requestId": string,
+      "resources": string[],
```
kdenhartog, line 105: I prefer this being an strong array rather than a string, but we shouldn’t new line separate here if it’s going to be an array.

CAIPs/caip-222.md
```diff
@@ -0,0 +1,249 @@
...
+  }
+}
+```
+
+All payload parameters MUST match the same parameters as the ones provided in the request and the following will be added:
+
+- h - header object according to [CAIP-74][caip-74]
+  - t - cacao message type (matching `type` in request)
+- p - payload object according to [CAIP-74][caip-74]
+  - iss - [did:pkh][did:pkh] method identifying [CAIP-10][caip-10] blockchain account
```
kdenhartog, line 121: I like the idea of using did:pkh here, but does it address chain forks? I think forks could present security issues here

CAIPs/caip-222.md
```diff
@@ -0,0 +1,249 @@
...
+- s - signature object according to [CAIP-74][caip-74]
+  - t - signature type
+  - s - signature bytes
+
+#### Failure
```
kdenhartog, line 126: We should probably consider fingerprinting concerns here too

CAIPs/caip-222.md
pedrouid: Sounds good! For the record these parameter descriptions were not inherited from SIWE (EIP-4361) but from SIWX (CAIP-122)

CAIP-222 is essentially a combination of CAIP-122 for request and CAIP-74 for response in a JSON-RPC exchange

So there is also going to be some changes inherited from CACAOv3 PR that is currently pending (#179)

CAIPs/caip-222.md
pedrouid: Wouldn't that be part of CAIP-122 scope instead?

CAIPs/caip-222.md
pedrouid: I'm not sure I follow... isn't this the same interface for both SIWx and CACAO?

CAIPs/caip-222.md
pedrouid: That's a deeper problem than did:pkh or even CAIP-10

For example, EIP-155 doesn't handle chain forks really well 

That means that one of the chains must change it's chainId

CAIPs/caip-222.md
pedrouid: Yeah I did this section a bit too quick TBH

CAIPs/caip-222.md
```diff
@@ -0,0 +1,249 @@
...
+
+Request will fail if rejected by the user or if paramaters fail validation.
```
ligi, line 128:
```suggestion
Request will fail if rejected by the user or if parameters fail validation.
```

CAIPs/caip-222.md
```diff
@@ -0,0 +1,249 @@
...
+
+This standard provides both benefits to users and developers by compiling two widely adopted patterns into a single request: connecting a wallet and signing an authentication message.
+
+Users have the ability to quickly authenticate into applications without establishing a persistent session before they have decided to actually expose a two-way communication channel to sign more messages and/or transactions.
+
+Developers have the ability to verify ownership of users' blockchain accounts similarly to how other authentication standards like OAuth and OIDC using also a more familiar interface that is commonly found in many social applications today.
+
+This also incentives more applications to prevent impersonation by verifying ownership of blockchain accounts upfront without needing to trust the wallet provider interface that optimistically exposes account on connection.
+
+This interface can be used with multiple accounts by simply calling it multiple times with a difference nonce and prompting the waller user to authenticate each account individually without establishing a persistent connection.
```
ligi, line 149:
```suggestion
This interface can be used with multiple accounts by simply calling it multiple times with a difference nonce and prompting the wallet user to authenticate each account individually without establishing a persistent connection.
```

ligi (changes requested):
Minor typos

CAIPs/caip-222.md
kdenhartog: If we reference directly to 122 as the normative definition then yeah that makes sense. I think that's actually the best way to do this too. I can help double check it to make sure we got this normatively defined well.

CAIPs/caip-222.md
kdenhartog: It may be, I've been reviewing on mobile so wasn't too easy to double check SIWE/SIWX/CACAO specs. I think in SIWE the intention was to encode the entire message structure into a string which is where new line charters were needed. I'm not sure if that was kept in SIWX, but I saw above in the resources definition that we utilized the new line character as the list delimiter character. So, as it currently reads this could be interpreted as needing to have a new line character at the end of each string inside a string array which I don't think would be the intention. That was the main thing I wanted to call out.

CAIPs/caip-222.md
kdenhartog: All good, just calling out so we can address it during the review stage. Getting a first draft is a great first start to iterate on. These concerns typically get addressed during more of the review process anyways.

CAIPs/caip-222.md
```diff
@@ -0,0 +1,249 @@
...
+// Request
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "method": "wallet_authenticate",
+  "params": {
+      "cacaov": "2",
+      "type": "eip4361",
+      "chainId": "eip155:1",
+      "iss": "did:pkh:eip155:1:0xBAc675C310721717Cd4A37F6cbeA1F081b1C2a07",
```
ligi, line 165:
```suggestion
```

I think the request should not have iss - only the response

PR #230, opened 2023-06-05
Quick-fix for jekyll rendering bug

bumblefudge:
I'm having trouble getting to the bottom of this typing chaos across yaml and random ruby/liquid dependencies of jekyll, so in the meantime i'm just manually reformatting the offending yaml array and putting a guide into the template so that PR authors don't fall into the same trap.  Going on record here to say that ducktyping is bad.


bumblefudge:
[Self-merging because .org isn't displaying the new PRs!]

PR #236, opened 2023-07-13
editorial: Clarify string representation of CACAO signing input

ukstv:
As noted in https://github.com/ChainAgnostic/namespaces/issues/41#issuecomment-1500659237 and https://github.com/ChainAgnostic/namespaces/issues/66 CAIP-122 is ambiguous.

We define `address` in data model as CAIP-10 account. String representation of singing input though uses plain `${address}` field, and `${chain-id}` that the data model does not even mention.

The PR here resolves the ambiguity. Data model remains the same, with `address` being CAIP-10. String representation is clarified to reflect two fields (previously `${address}` and `${chain-id}`) are derived from data model's `address`.

### DIFF
--- a/CAIPs/caip-122.md
+++ b/CAIPs/caip-122.md
@@ -68,7 +68,7 @@ The proposed string representation format, inspired from [EIP-4361][eip-4361], s
 
 ```
 ${domain} wants you to sign in with your **blockchain** account:
-${address}
+${account_address(address)}
 
 ${statement}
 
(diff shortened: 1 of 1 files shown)


haardikk21 (approved):
Thank you 🙏

bumblefudge (commented):
This seems reasonable to me, since it's the net-same information displayed in a more readily signable message.  ChainID will still be illegible in namesapces where CAIP-2 is counterintuitive or not part of the native end-user's mental model, but at the very least this is less scary than having an incomprehensible bit of hash prepended to the address (which is what the EIP editors insisted on in the EIP). 

Do we have any sense of how many non-Ceramic implementers this could break, tho? Is it worth adding some mention of the previous state to "## Legacy considerations?"

ukstv:
> This seems reasonable to me, since it's the net-same information displayed in a more readily signable message. ChainID will still be illegible in namesapces where CAIP-2 is counterintuitive or not part of the native end-user's mental model, but at the very least this is less scary than having an incomprehensible bit of hash prepended to the address (which is what the EIP editors insisted on in the EIP).
> 
> Do we have any sense of how many non-Ceramic implementers this could break, tho? Is it worth adding some mention of the previous state to "## Legacy considerations?" I.e. if Haardik's library and Ceramic's library are being used to generate the messages AND verify them in 99% of contexts, then it's a no-brainer, but it's still worth discussing downstream risks insofar as they're known, making it a major-version change in package managers, etc etc...

AFAIK, that only changes wording, not the implementations.

bumblefudge:
hehe by this i meant, throw currently-conformant implementations out of conformance (and interop)

bumblefudge (approved):

silverpill (approved):
LGTM

bumblefudge (approved):

PR #248, opened 2023-10-19
feat(caip-222): multi-chain wallet authentication

ganchoradkov:
Updated [CAIP-222](https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-222.md) to support multi-chain authentication.

Added ability to 
- request signatures for multiple chains in a single request
- specify supported signing types via `signatureTypes` request parameter
- respond with a list of signed CACAOs

### DIFF
--- a/CAIPs/caip-222.md
+++ b/CAIPs/caip-222.md
@@ -77,7 +77,7 @@ The JSON-RPC method is labelled as `wallet_authenticate` and expects the followi
 - statement (optional) - Human-readable ASCII assertion that the user will sign. It _MUST NOT_ contain `\n`.
 - requestId (optional) - System-specific identifier used to uniquely refer to the authentication request.
 - resources (optional) - List of information or references to information the user wishes to have resolved as part of the authentication by the relying party; express as [RFC 3986][rfc 3986] URIs and separated by `\n`.
-- signatureTypes (optional) - Object specifying the supported signing algorithms by the application for each namespace. The namespace MUST be defined in the key, while the supported signing algoriths as `string[]` and MUST contain one or more values if present.
+- signatureTypes (optional) - Object specifying the supported signing algorithms by the application for each namespace. The namespace MUST be defined in the key, while the supported signing algorithms as `string[]` and MUST contain one or more values if present.
 
 Example of `signatureTypes`
 
--- a/CAIPs/caip-222.md
+++ b/CAIPs/caip-222.md
@@ -32,7 +32,7 @@ This JSON-RPC method can be requested to a wallet provider without prior knowled
 
 The requester will provide parameters required by [CAIP-122][caip-122] plus a CACAO header type as specified by [CAIP-74][caip-74]
 
-The responder will return a signed CACAO(s) with a header type and payload matching the requested parameters
+The responder will return zero, one, or more signed CACAO(s) with a header type and payload matching the requested parameters.
 
 ### Request
 
--- a/CAIPs/caip-222.md
+++ b/CAIPs/caip-222.md
@@ -12,7 +12,7 @@ requires: 2, 10, 74, 122
 
 ## Simple Summary
 
-This CAIP defines a JSON-RPC method to authenticate wallet ownership/control of a blockchain account(s).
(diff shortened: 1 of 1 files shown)


CAIPs/caip-222.md
```diff
@@ -75,20 +77,30 @@ The JSON-RPC method is labelled as `wallet_authenticate` and expects the followi
 - statement (optional) - Human-readable ASCII assertion that the user will sign. It _MUST NOT_ contain `\n`.
 - requestId (optional) - System-specific identifier used to uniquely refer to the authentication request.
 - resources (optional) - List of information or references to information the user wishes to have resolved as part of the authentication by the relying party; express as [RFC 3986][rfc 3986] URIs and separated by `\n`.
+- signatureTypes (optional) - Object specifying the supported signing algorithms by the application for each namespace. The namespace MUST be defined in the key, while the supported signing algoriths as `string[]` and MUST contain one or more values if present.
+
+Example of `signatureTypes`
+
+```
+signatureTypes: {
+  "eip155": ["eip191", "eip1271" ],
```
oed, line 86: So if I support both eip191 and eip1271 I can just not define this field and get either?

bumblefudge: i think wildcards are kind of unsafe and likely to lead to bad UX deadends; better for dapps to have to explicitly list which kinds they support, I would think.

ganchoradkov: The v1 (merged version) is very optimistic that the sender application would be able to validate the returned signature without knowing the signing type especially now that you could request non-evm signatures. This parameter lets you specify exactly which types you're supporting. Whether you can omit it, its up for debate. I've set it as optional to reduce devex friction.

CAIPs/caip-222.md
```diff
@@ -45,7 +46,7 @@ The application would interface with a provider to make request as follows:
   "params": {
       "cacaov": string,
       "type": string,
-      "chainId": string,
+      "chains": string[],
```
arein, line 49: should this be optional and the wallet can just respond with any of those it supports?

ganchoradkov: The rationale is that the dapp would request signatures for all chains included in the `chains` parameter and the wallet would respond with signatures for those it supports as outlined [here](https://github.com/ChainAgnostic/CAIPs/blob/7087562ed1f3a49e62ca6eb739bc32b67961d1bc/CAIPs/caip-222.md#not-supported-chains)
>The wallet SHOULD ignore all unsupported chains and SHOULD NOT auto reject the request if there are supported chains to sign without explicit user rejection.

In v1, `chainId` is required parameter so it makes sense to preserve that to an extended.
Additionally, setting `chains` to explicitely empty `[]` array is valid and kinda allows to not set any chains already.

CAIPs/caip-222.md
```diff
@@ -75,20 +77,30 @@ The JSON-RPC method is labelled as `wallet_authenticate` and expects the followi
 - statement (optional) - Human-readable ASCII assertion that the user will sign. It _MUST NOT_ contain `\n`.
 - requestId (optional) - System-specific identifier used to uniquely refer to the authentication request.
 - resources (optional) - List of information or references to information the user wishes to have resolved as part of the authentication by the relying party; express as [RFC 3986][rfc 3986] URIs and separated by `\n`.
+- signatureTypes (optional) - Object specifying the supported signing algorithms by the application for each namespace. The namespace MUST be defined in the key, while the supported signing algoriths as `string[]` and MUST contain one or more values if present.
```
ganchoradkov, line 80:
```suggestion
- signatureTypes (optional) - Object specifying the supported signing algorithms by the application for each namespace. The namespace MUST be defined in the key, while the supported signing algorithms as `string[]` and MUST contain one or more values if present.
```
typo

bumblefudge:
I'm still wondering if it makes more sense for the multi-CAIP222 to be CAIP-248 or CAIP-222; would love input from potential or early implementers.

ganchoradkov:
> Member

@bumblefudge imho, it should be an update as it complely encapsulates the merged CAIP-222. Additionally, having them as different standards would cause conflics if one attempts to implement them at the same time in their current form forcing implementers to choose one OR the other

llbartekll (approved):

bumblefudge (approved):

PR #261, opened 2024-02-06
New CAIP - Web of Trust Primitives (Split of former: Community-powered trust assessment in software components)

dayksx:
Initial proposal for a standardized data framework designed to facilitate community-powered trust evaluation in software components.


mirceanis (commented):
The proposal looks good for bootstrapping such an ecosystem.

I have some comments about adherence to the Verifiable Credential data model:

1. Any subject `id` must be a URI according to the VC spec. This would at the very least help implementers figure out what the subject is, as in the case of opaque IDs it's anyone's guess as to what a review or endorsement is about.
2. When referencing software components I think it's important to specify some form of versioning and recommend that it be as granular as possible, ideally uniquely identifying a particular release or commit hash.
3. The spec proposes that only `did:pkh` be used for peer IDs and that `did:pkh` or `did:key` to refer to trust computers. I believe it's important to require a minimum set of issuer DID methods for better interop but to also allow allow other DID methods to participate. I'd argue that issuers be limited to DIDs even if the Verifiable Credential data model allows any URI to be used, as the verification process is well known in case of DIDs.
4. The examples don't detail any `proof`. It's important to stress the importance of cryptographically verifiable proofs that bind the assertions to the issuer such that tampering is evident and impersonations are impossible, especially since the exact sources of credentials are intentionally open.
5. In the Verifiable Credential model a `credentialStatus` method can also be specified which allows a verifier to determine if the credential was revoked by the issuer. It may be worth mentioning this mechanism here too. This can also be achieved by an issuer "disputing" their own credential, but it raises a question of availability.
7. An optional detail but potentially relevant during implementation: for better compatibility with EIP712, arrays of objects should adhere to the same schema. For example, an optional `reason` property for an endorsement should be an empty array instead of being omitted.

I also took the liberty of suggesting fixes for some typos that I could spot. Please ignore them if it's too early for such corrections.

dbcfd:
One thing that is not clear to me is if the issued assessments are updateable, and if so, which fields are updateable. 

A given assessment by an issuer has the possibility to change over time. It feels like the assessment itself should change, other than issuer and subject, rather than trying to build a linking from previous assessment to current assessment.

mirceanis:
> One thing that is not clear to me is if the issued assessments are updateable, and if so, which fields are updateable.
> 
> A given assessment by an issuer has the possibility to change over time.

That's an excellent point!

Verifiable credentials are static objects, they are not updateable, but they can point to mechanisms of refresh or revocation.

An outdated credential should be revoked, possibly through its `credentialStatus` property.

The new one can contain a reference to the previous, in which case the trust computer can decide to override the previous as if it is revoked.

Vid201:
Some feedback from my side as well. First of all, kudos for putting the proposal together!

1. There could be some advantages to using DIDs for software components as well. Some DID methods allow the insertion of additional data into the DID Document, which could link to a website or some technical documentation of the software component (better discoverability). If there is only a technical identifier (such as checksum), it's hard to find more information about the software component from the identifier directly.
2. What is the role of software entities? Is there some verifiable connection between the software component and the owner/author of the software? It would be great to expand on that.
3. Verifiable Credentials are missing the credentialSchema and context fields.
4. Some things are strictly defined in the specification, while some are not strictly defined and are entirely open. It may make sense to split the proposal into two parts - one with exactly described things and the second part documenting what is not defined (and maybe include some suggestions and examples or even a reference example of how all components fit together in the Snaps model/system).
5. I'm not sure if this should be a part of the specs, but some sections on registries could be included, such as where to find what credential/JSON schemas and values exist.
6. Adding to Mircea's comment about credentialStatus: I agree it would make sense to include it, although some considerations should be included on how to handle revoked attestations.

dayksx:
Thank you all for your valuable feedback. I am currently reviewing it.

CAIPs/caip-x.md
```diff
@@ -0,0 +1,424 @@
...
+By standardizing data to form a universally applicable trust graph reusable in any context, we strengthen the reliability of software components assessments powered by the communities.
+
+## Specification
+<!--The technical specification should describe the standard in detail. The specification should be detailed enough to allow competing, interoperable implementations. -->
+### Identification
+Decentralized Identifiers ([DID](https://www.w3.org/TR/did-core/)) or Content Identifier (CID) are used to identify subjects such as peers, software components as well as the claims / assertions themselves.
+They can also identify any issuers, which could be peers or a software entity like a trust computers.
+
+The standardized data framework specifies the use of the following identifiers:
+- **Peers:** `PKH` DID method (e.g. `did:pkh:eip155:1:<publicAddress>`, `did:pkh:bip122:<publicAddress>`, `did:pkh:solana:<publicAddress>`);
```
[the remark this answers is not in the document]
dayksx: Indeed, the goal is to accommodate any peer identifier. 
I changed it:

>Peers: Any key pair based DID such as PKH DID method (e.g. did:pkh:eip155:1:<publicAddress>, did:pkh:bip122:><publicAddress>, did:pkh:solana:<publicAddress>), KEY DID method, etc.;

CAIPs/caip-x.md
```diff
@@ -0,0 +1,424 @@
...
+- **Software components:** Custom identifiers for software components such as the checksum (e.g. `snap://<checksum>`, `did:pkh:eip155:1:<contractAddress>`);
+- **Assertions:** CID generated based on their contents according to [RFC 8785
+JSON Canonicalization Scheme (JCS)](https://www.rfc-editor.org/rfc/rfc8785);
+- **Software entities:** `KEY` or `PKH` DID method.
```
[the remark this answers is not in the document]
dayksx: Indeed, I will list them as options rather than specifications. However, do you suggest any other relevant identifiers for software components and for assertions?

CAIPs/caip-x.md
```diff
@@ -0,0 +1,424 @@
...
+},
+"proof": {}
+```
+
+```json
+"type": ["VerifiableCredential", "ReviewCredential"],
+"issuer": "did:pkh:eth:0x44dc4E3309B80eF7aBf41C7D0a68F0337a88F044",
+"credentialSubject":
+{
+  "id": "d6f7052b6f28912f2703066a912ea577f2ce4da4caa5a5fbd8a57286c345c2f2",
```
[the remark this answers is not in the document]
dayksx: Indeed, that makes sense. I have augmented all the isolated content identifiers with the IPFS URI.

CAIPs/caip-x.md
```diff
@@ -0,0 +1,424 @@
...
+- A Software component, by issuing **Security Report** or **Review** assertions,
+- A Security report, by issuing **Review** assertions.
+
+![image](https://github.com/dayksx/CAIPs/assets/77788154/96161b44-6cfa-477d-9069-cc47cf9fbf38)
+
+*View - Software component Trust Assessment Metamodel*
+
+All subsequent documents adhere to the [Verifiable Credential Data Model](https://www.w3.org/TR/vc-data-model/) for representation purposes. 
+However this standard does not prescribe any specific document type, even though internationally recognized standards are recommended.
+The standard presumes that both the `issuer` property and the complete content of the `credentialSubject` will be only utilized once the wire-formats and signed-envelopes have been verified.
```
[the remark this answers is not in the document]
dayksx: Yes, clarification is needed. Thank you for pointing that out; I have included this note accordingly. [#1066f4d](https://github.com/dayksx/CAIPs/commit/1066f4d2f4df97fe7a26fac31be2a9f069af5b22)

dayksx:
> The proposal looks good for bootstrapping such an ecosystem.
> 
> I have some comments about adherence to the Verifiable Credential data model:
> 
> 1. Any subject `id` must be a URI according to the VC spec. This would at the very least help implementers figure out what the subject is, as in the case of opaque IDs it's anyone's guess as to what a review or endorsement is about.
> 2. When referencing software components I think it's important to specify some form of versioning and recommend that it be as granular as possible, ideally uniquely identifying a particular release or commit hash.
> 3. The spec proposes that only `did:pkh` be used for peer IDs and that `did:pkh` or `did:key` to refer to trust computers. I believe it's important to require a minimum set of issuer DID methods for better interop but to also allow allow other DID methods to participate. I'd argue that issuers be limited to DIDs even if the Verifiable Credential data model allows any URI to be used, as the verification process is well known in case of DIDs.
> 4. The examples don't detail any `proof`. It's important to stress the importance of cryptographically verifiable proofs that bind the assertions to the issuer such that tampering is evident and impersonations are impossible, especially since the exact sources of credentials are intentionally open.
> 5. In the Verifiable Credential model a `credentialStatus` method can also be specified which allows a verifier to determine if the credential was revoked by the issuer. It may be worth mentioning this mechanism here too. This can also be achieved by an issuer "disputing" their own credential, but it raises a question of availability.
> 6. An optional detail but potentially relevant during implementation: for better compatibility with EIP712, arrays of objects should adhere to the same schema. For example, an optional `reason` property for an endorsement should be an empty array instead of being omitted.
> 
> I also took the liberty of suggesting fixes for some typos that I could spot. Please ignore them if it's too early for such corrections.



> One thing that is not clear to me is if the issued assessments are updateable, and if so, which fields are updateable.
> 
> A given assessment by an issuer has the possibility to change over time. It feels like the assessment itself should change, other than issuer and subject, rather than trying to build a linking from previous assessment to current assessment.

dayksx:
> A given assessment by an issuer has the possibility to change over time. 


> An outdated credential should be revoked, possibly through its credentialStatus property.
> 
> The new one can contain a reference to the previous, in which case the trust computer can decide to override the previous as if it is revoked.

@dbcfd @mirceanis Indeed, you've raised a valid point. In our ongoing experiments, we are operating under the assumption that any new type of credential issued by a specific issuer to a particular subject will supersede the previous ones.

@mirceanis would it be considered a best practice to include a `credentialStatus` field (with explicit values like `update` or `revocation`), and maybe a `previousCredential` field referring to the updated credential, within the newly issued credential to indicate that it supersedes an older one?

```json
"type": ["VerifiableCredential", "TrustCredential"],
"issuanceDate": "2024-02-29T14:31:56.273Z",
"issuer": "did:pkh:eip155:1:0x44dc4E3309B80eF7aBf41C7D0a68F0337a88F044",
"credentialSubject":
{
  "id": "did:pkh:eip155:1:0xfA045B2F2A25ad0B7365010eaf9AC2Dd9905895c",
  "credentialStatus": "Update",
  "previousCredential": "ipfs://QmPTqvH3vm6qcZSGqAUsq78MQa9Ctb56afRZg1WJ5sKLiu",
  "trustworthiness":
  [
    {
      "scope": "Honesty",
      "level": 1,
      "reason": ["Alumnus"]
    },
    {
      "scope": "Software development",
      "level": 1,
      "reason": ["Software engineer", "Ethereum core developer"]
    },
    {
      "scope": "Software security",
      "level": 1,
      "reason": ["White Hat", "Smart Contract Auditor"]
    }
  ]
},
"proof": {}
```

dayksx:
> 1. There could be some advantages to using DIDs for software components as well. Some DID methods allow the insertion of additional data into the DID Document, which could link to a website or some technical documentation of the software component (better discoverability). If there is only a technical identifier (such as checksum), it's hard to find more information about the software component from the identifier directly.

Sure, I think opening up the standard to DIDs for software components is good.
Even knowing that DID could be based on checksum, it is also possible to use another identifier that is not self containing. Of course this is acceptable, but imply additional work. We need to specify that creating a new DID require to use a verifiable data registry in order to manage the DID lifecycle, for creation, updating, resolving, and deactivating DIDs.

dayksx:
> 2. What is the role of software entities? Is there some verifiable connection between the software component and the owner/author of the software? It would be great to expand on that.
Software entities has been introduced to represent software able to issue assertions, such as trust computers, that issue true scores. 
Why do you think it will be important to have a verifiable connection between the software component and the owner/author of the software? Here, for example the trust computer would be known with a DID, the controller of this DID would be the owner/author of the trust computer.

dayksx:
> 4. Some things are strictly defined in the specification, while some are not strictly defined and are entirely open. It may make sense to split the proposal into two parts - one with exactly described things and the second part documenting what is not defined (and maybe include some suggestions and examples or even a reference example of how all components fit together in the Snaps model/system).

Could you specify which particular aspects you're referring to? I am seeing the entire computational component that has not been strictly defined. What else do you observe?

dayksx:
> 5. I'm not sure if this should be a part of the specs, but some sections on registries could be included, such as where to find what credential/JSON schemas and values exist.

Do you mean the registries where all issued assertions and their corresponding schemas can be found? If that's the case, I have outlined key requirements related to registries in the 'Data and Trust Score Storage' section. Do you think we should expand on this?

Vid201:
> > A given assessment by an issuer has the possibility to change over time.
> 
> > An outdated credential should be revoked, possibly through its credentialStatus property.
> > The new one can contain a reference to the previous, in which case the trust computer can decide to override the previous as if it is revoked.
> 
> > Adding to Mircea's comment about credentialStatus: I agree it would make sense to include it, although some considerations should be included on how to handle revoked attestations.
> 
> @dbcfd @mirceanis @Vid201 Indeed, you've raised a valid point. In our ongoing experiments, we are operating under the assumption that any new type of credential issued by a specific issuer to a particular subject will supersede the previous ones.
> 
> @mirceanis would it be considered a best practice to include a `credentialStatus` field (with explicit values like `update` or `revocation`), and maybe a `previousCredential` field referring to the updated credential, within the newly issued credential to indicate that it supersedes an older one?
> 
> ```json
> "type": ["VerifiableCredential", "TrustCredential"],
> "issuanceDate": "2024-02-29T14:31:56.273Z",
> "issuer": "did:pkh:eip155:1:0x44dc4E3309B80eF7aBf41C7D0a68F0337a88F044",
> "credentialSubject":
> {
>   "id": "did:pkh:eip155:1:0xfA045B2F2A25ad0B7365010eaf9AC2Dd9905895c",
>   "credentialStatus": "Update",
>   "previousCredential": "ipfs://QmPTqvH3vm6qcZSGqAUsq78MQa9Ctb56afRZg1WJ5sKLiu",
>   "trustworthiness":
>   [
>     {
>       "scope": "Honesty",
>       "level": 1,
>       "reason": ["Alumnus"]
>     },
>     {
>       "scope": "Software development",
>       "level": 1,
>       "reason": ["Software engineer", "Ethereum core developer"]
>     },
>     {
>       "scope": "Software security",
>       "level": 1,
>       "reason": ["White Hat", "Smart Contract Auditor"]
>     }
>   ]
> },
> "proof": {}
> ```

`credentialStatus` is a top level field of the VC data model (https://www.w3.org/TR/vc-data-model-2.0/#status). The old credential is revoked, and the new credential is issued. So something like that could make sense:

```
"type": ["VerifiableCredential", "TrustCredential"],
"issuanceDate": "2024-02-29T14:31:56.273Z",
"issuer": "did:pkh:eip155:1:0x44dc4E3309B80eF7aBf41C7D0a68F0337a88F044",
"credentialStatus": {
    "id": "https://snaps.example/credentials/status/3#94567",
    "type": "BitstringStatusListEntry",
    "statusPurpose": "revocation",
    "statusListIndex": "94567",
    "statusListCredential": "https://snaps.example/credentials/status/3",
},
"credentialSubject":
{
  "id": "did:pkh:eip155:1:0xfA045B2F2A25ad0B7365010eaf9AC2Dd9905895c",
  "trustworthiness":
  [
    {
      "scope": "Honesty",
      "level": 1,
      "reason": ["Alumnus"]
    },
    {
      "scope": "Software development",
      "level": 1,
      "reason": ["Software engineer", "Ethereum core developer"]
    },
    {
      "scope": "Software security",
      "level": 1,
      "reason": ["White Hat", "Smart Contract Auditor"]
    }
  ]
},
"proof": {}
```

Vid201:
> > 3. Verifiable Credentials are missing the credentialSchema and context fields.
> 
> Does this imply that we should include the @context field for each credential, as illustrated below, correct?
> 
> ```json
> "@context": [
>     "https://www.w3.org/2018/credentials/v1"
>   ]
> ```

Yes.

Vid201:
> > 5. I'm not sure if this should be a part of the specs, but some sections on registries could be included, such as where to find what credential/JSON schemas and values exist.
> 
> Do you mean the registries where all issued assertions and their corresponding schemas can be found? If that's the case, I have outlined key requirements related to registries in the 'Data and Trust Score Storage' section. Do you think we should expand on this?

I meant only the schemas. Since the standard defines how the attestations look like (structure/JSON schema), it could make sense to have some registry for the schemas, while the data storage for the actual attestations themselves is left to the implementors of the system.

Vid201:
> > 2. What is the role of software entities? Is there some verifiable connection between the software component and the owner/author of the software? It would be great to expand on that.
> 
> Software entities has been introduced to represent software able to issue assertions, such as trust computers, that issue true scores.
> 
> Why do you think it will be important to have a verifiable connection between the software component and the owner/author of the software? Here, for example the trust computer would be known with a DID, the controller of this DID would be the owner/author of the trust computer.

Hmm, I think some more clarification and differences (technical) between peers and software entities would be good to have.

Vid201:
> > 4. Some things are strictly defined in the specification, while some are not strictly defined and are entirely open. It may make sense to split the proposal into two parts - one with exactly described things and the second part documenting what is not defined (and maybe include some suggestions and examples or even a reference example of how all components fit together in the Snaps model/system).
> 
> Could you specify which particular aspects you're referring to? I am seeing the entire computational component that has not been strictly defined. What else do you observe?

Yes, I agree. For example, type of actors in the system and type of attestations are strictly defined, while data storage, computational component, and some other things are left to ones that will implement this system. So there are some things that are strictly defined and some that are not.

CAIPs/caip-x.md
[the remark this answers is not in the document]
bumblefudge: lots of package managers identify individual binaries by a hash, if so this seems to be a pretty neutral/generic way of URN-ifying hashes that works across some common hashes:
https://datatracker.ietf.org/doc/html/draft-thiemann-hash-urn-01

or multiformats?

bumblefudge:
> > > 5\. Verifiable Credentials are missing the credentialSchema and context fields.
> > 
> > 
> > Does this imply that we should include the @context field for each credential, as illustrated below, correct?
> > ```json
> > "@context": [
> >     "https://www.w3.org/2018/credentials/v1"
> >   ]
> > ```
> 
> Yes.

So, technically, `@Context`  is mandatory and ` credentialSchema`  is optional, but in this use-case I would recommend [using both](https://w3c.github.io/vc-data-model/#differences-between-contexts-types-and-credentialschemas) because the latter forces everyone to be much more explicit as they add new assertion formats or variables to the system.  more layers of validation is always good in an open-world system where all kinds of folks will be piling in with their messy data and unicode issues and who knows what else 😏

bumblefudge:
> > 1. There could be some advantages to using DIDs for software components as well. Some DID methods allow the insertion of additional data into the DID Document, which could link to a website or some technical documentation of the software component (better discoverability). If there is only a technical identifier (such as checksum), it's hard to find more information about the software component from the identifier directly.
> 
> Sure, I think opening up the standard to DIDs for software components is good. Even knowing that DID could be based on checksum, it is also possible to use another identifier that is not self containing. Of course this is acceptable, but imply additional work. We need to specify that creating a new DID require to use a verifiable data registry in order to manage the DID lifecycle, for creation, updating, resolving, and deactivating DIDs.

Not all did methods require a verifiable data registry per se: [did:web](https://w3c-ccg.github.io/did-method-web/) just requires a website, while [did:dns](https://danubetech.github.io/did-method-dns/) just requires a website and access to advanced or manual DNS registration. Both allow the insertion of any additional data into the did document that doesn't break DID doc syntax. 

It's also worth mentioning that registering a mobile app to an appstore, or registering as a publisher with most package managers, requires... a website. 

It might make sense to model software actors as 1-to-many instances of a given binary/build, and binaries/builds as 1-to-many outputs of stably-identified publishers of software, not just for the software being assessed but also for the trust computers synthetisizing other trust docs into trust scores and derived/synthetic trust docs.  Doing so doesn't necessarily mean you have to make a valid, live `did:web` a gating function or anything, but it might be an interesting SHOULD statement or example-pattern?

CAIPs/caip-261.md
```diff
@@ -0,0 +1,475 @@
...
+"issuer": "did:pkh:eth:0x44dc4E3309B80eF7aBf41C7D0a68F0337a88F044",
+"credentialSubject":
+{
+  "id": "snap://CLwZocaUEbDErtQAsybaudZDJq65a8AwlEFgkGUpmAQ=",
+  "securityStatus": "Secured"
+},
+"proof": {}
+```
+
+- The `securityStatus` is the final result of the security assessment, that can be either `Secured` or `Unsecured`.
```
ra-phael, line 232: Should this be optional? Because I'm wondering if it isn't too reductive. With most software applications we use we accept to take on some level of risk, but it's not black or white. 
The smart contract audits that I've seen have findings with different levels of criticality, but I guess auditors don't want to be forced to draw the line somewhere to declare that a piece of software is `Secured` or `Unsecured` when the reality is more nuanced than that.

CAIPs/caip-261.md
```diff
@@ -0,0 +1,475 @@
...
+These three data elements can be used independently of one another.
+For building a community-powered trust assessment mechanism, it's necessary to identify and gather the relevant data in accordance with the requirements of the intended trust scores computation.
+These data can then be leveraged to compute synthetic trust scores which reflect the overall sentiment of the community.
+
+The data framework modeled below incorporates the following basic primitives as inputs:
+
+- **Trust List:** This allows individuals to define their trusted peers, shaping their trust graph;
+- **Expert Report:** This enables experts to publish insights and conclusions about a given resource (e.g. an audit in the case of software components);
+- **Peer Review:** This allows individuals to endorse or dispute claims made about resources by peers;
+- **Trust Score:** This enables any trust computer to publish computed trust scores about resources and/or about the actors behind the other 3 primitives.
```
ra-phael, line 36: Might be useful to define "trust computer"?

CAIPs/caip-261.md
```diff
@@ -0,0 +1,475 @@
...
+
+```json
+"type": ["VerifiableCredential", "ReviewCredential"],
+"issuanceDate": "2024-02-15T07:05:56.273Z",
+"issuer": "did:pkh:eth:0x44dc4E3309B80eF7aBf41C7D0a68F0337a88F044",
+"credentialSubject":
+{
+  "id": "snap://CLwZocaUEbDErtQAsybaudZDJq65a8AwlEFgkGUpmAQ=",
+  "currentStatus": "Disputed",
+  "reason": ["Scam", "Phishing"]
```
ra-phael, line 285: What if someone wants to provide a longer explanation for the reason? To hack my way around that personally I would post the explanation on hackmd for example and put the link as one of the items in this array  😄

CAIPs/caip-261.md
bumblefudge: maybe auditors would be more comfortable with a boolean like `audited`, and even then they'd probably want it applied only to a specific binary 😅

dayksx:
> I meant only the schemas. Since the standard defines how the attestations look like (structure/JSON schema), it could make sense to have some registry for the schemas, while the data storage for the actual attestations themselves is left to the implementors of the system.

Yes it makes sense, so it would be a resource pointed in the "context" field in the context of VC?

dayksx:
As mentioned in some comments, we have decided to ultimately split this CAIP into three parts, each addressing the following aspects:

1. **CAIP-261:** Web of Trust Primitives (this one)
2. **CAIP-x:** Community-Powered Trust Assessment (PR to be created)
3. **CAIP-x:** Community-Powered Trust Assessment in software components (PR to be created)

All above comments should have been addressed in this new CAIP-261 version.

diagram

cc @bumblefudge

CAIPs/caip-261.md
dayksx: The primary purpose of this credential is to facilitate quick reactions to another assertion or any discrete resource, akin to the thumb up/thumb down, upvote/downvote, or emojis used in social networks. Given this context, perhaps renaming the credential to "ReactionCredential" would more accurately reflect its intended use, as opposed to "ReviewCredential".

For more detailed analyses, we would utilize the "Expert Report" credential. This format is designed to comprehensively address issues, including descriptions and references to external resources. Considering your feedback, it might be appropriate to consider names like "ReportCredential" or retain "ReviewCredential" for this purpose.

Please note, this data element has been removed from CAIP-261 and will be included in the forthcoming [CAIP-x](https://github.com/dayksx/CAIPs/blob/community-powered-trust-assessment/CAIPs/caip-x.md).

dayksx: Here is what this modeling would look like.

diagram


**Report Assertion**

*A report presents a detailed presentation of factual information and objective analysis.*
```json
"type": ["VerifiableCredential", "ReportCredential"],
"issuanceDate": "2024-02-15T07:05:56.273Z",
"issuer": "did:pkh:eth:0x44dc4E3309B80eF7aBf41C7D0a68F0337a88F044",
"credentialSubject":
{
  "id": "ipfs://QmPTqvH3vm6qcZSGqAUsq78MQa9Ctb56afRZg1WJ5sKLiu",
  "type": "Security",
  "result": -1,
  "issues": [
    {
      "criticality": 1,
      "type": "Key leak",
      "description": "`snap_getBip44Entropy` makes the parent key accessible",
      "uri": "https://hackmd.io/@ra-phael/DummySnap5KeyLeak"
    },
    {
      "criticality": 0.5,
      "type": "Buffer Overflow"
      "uri": "https://hackmd.io/@ra-phael/DummySnap5BufferOverflow"
    },
    {
      "criticality": 0.25,
      "type": "Phishing"
      "uri": "ipfs://QmElreH3vm6qcASGqAUsq78MQa9Ctb56afRZg1WJ5sKLpl"
    },
    {
      "criticality": 0,
      "type": "Data leak",
      "description": "API can communicate data to a centralized server"
    },
  ]
},
"proof": {}
```
*Note that the Report Assertion could pertain to matters beyond 'Security,' including 'User Experience,' 'Compliance,' and 'Performance' reports, among others.*

**Reaction Assertion**

*A reaction represents a quantifiable expression of agreement or disagreement with the report's content, typically reflecting the collective sentiment of the community.*
```json
"type": ["VerifiableCredential", "ReactionCredential"],
"issuanceDate": "2024-02-15T07:05:56.273Z",
"issuer": "did:pkh:eth:0x44dc4E3309B80eF7aBf41C7D0a68F0337a88F044",
"credentialSubject":
{
  "id": "ipfs://QmPTqvH3vm6qcZSGqAUsq78MQa9Ctb56afRZg1WJ5sKLiu",
  "reaction": "Endorsed",
  "reason": ["Provide important context", "Vulnerabilities clearly defined"],
},
"proof": {}
```
The example above illustrates a reaction to a Report, but it can also apply to a reaction towards a discrete resource, such as a software component.

@ra-phael @bumblefudge @Vid201 @mirceanis @dbcfd Wdyt about this representation?

Vid201:
> > I meant only the schemas. Since the standard defines how the attestations look like (structure/JSON schema), it could make sense to have some registry for the schemas, while the data storage for the actual attestations themselves is left to the implementors of the system.
> 
> Yes it makes sense, so it would be a resource pointed in the "context" field in the context of VC?

Yes, both in context and credentialSchema. So both context files and credential schemas goes into some registry.

CAIPs/caip-261.md
bumblefudge: i like these representations!  my only bikeshedding would be to suggest new property names:
* `type` is too easy to confuse with VC, filesystem, programming language, and other uses of the term; i'd use something that makes more explicit that it's an issue type, like `issueType`, `issueClass`, `tag`, `category`, etc.
* `uri` is a type (or a "domain" in some old-school contexts), i.e. a constraint on the structure of the string that the value of that property can be-- since the *function* of this particular URI is to anchor semantics or disambiguate/explain/enrich the `issueType` preceding it, maybe a more explicit name would be `typeDefinition` or `classDefinition` etc, or `tagContext` or `definition`, etc?

bumblefudge:
> > > I meant only the schemas. Since the standard defines how the attestations look like (structure/JSON schema), it could make sense to have some registry for the schemas, while the data storage for the actual attestations themselves is left to the implementors of the system.
> > 
> > 
> > Yes it makes sense, so it would be a resource pointed in the "context" field in the context of VC?
> 
> Yes, both context and credentialSchema. So both context files and credential schemas goes into some registry.

Note that both `@Context` and `credentialSchema` are arrays, so a given credential can have multiple of either.  I would imagine that, realistically, a registry worth building for these kinds of receipts would have many credentialSchemas but only a few `@Context` files (i.e., most new registrations could get away with just reusing the original `@Context` or maybe extending it by adding one or two more entries or even inlining those, but most registrations would add a new JSON Schema). Not sure any of this has to go into this document, per se, just providing additional color in case there is an actual registry being designed in short-order here :D

CAIPs/caip-261.md
```diff
@@ -0,0 +1,384 @@
...
+- Data integrity,
+- Data ownership,
+- Interoperabiltiy,
+- Scalability,
+- Censorship resistance,
+- Sustainability.
+
+#### Trust Update
+When an issuer needs to update a trust assertion, they simply generate a new assertion containing the revised information. 
+This new assertion, when issued, will supersede any previous assertions of the same type, issued by the same entity, and pertaining to the same subject.
```
bumblefudge, line 191: is there any tombstoning system for assertions that have been superceded? could/should there be, at least optionally? people unfamiliar with IPFS may not realize they have to do their own garbage collecting or dropping of records that have been replaced but haven't expired yet...

CAIPs/caip-261.md
```diff
@@ -0,0 +1,384 @@
...
+#### Signature Verification
+
+The veracity and integrity of trust assertions are paramount to ensuring unbiased insights. 
+All trust assertions SHOULD be cryptographically signed by the issuer using strong cryptographic methods and verified prior to consumption. 
+The standard supports any strong signature methods, such as: ECDSA, EdDSA, Schnorr Signatures, RSA...
+
+EIP-712 should be considered a complementary cryptographic proof method alongside others like ECDSA, EdDSA, Schnorr, and RSA for on-chain verifiable credentials. 
+Its inclusion emphasizes the importance of user-friendly, secure, and efficient interactions with blockchain-based identity and credential systems.
+It is noteworthy that EIP-712 mandates the presence of all fields, even if some are left empty in order to enable their verification.
+
```
bumblefudge, line 231: This example looks a little iffy? https://w3c-ccg.github.io/ethereum-eip712-signature-2021-spec/ makes it sound like there are some mandatory fields missing here?

CAIPs/caip-261.md
bumblefudge: oh wait all comments won't make sense here until the commit i'm working on get merged-- sorry, mixing substantial and editorial in one mega commit again 🤦

CAIPs/caip-261.md
```diff
@@ -0,0 +1,384 @@
...
+### Trust Modeling
+
+1. Trust in an individual or entity is based on their qualities, or their abilities; it is not binary and evolves over time,
+2. Distrust assertions allow for the capture of suspicious behaviors,
+
+### Verification
+The verification process is crucial to ensure the use of valid and up-to-date assertions, thus maintaining the trust system's reliability and accuracy. 
+This step ensures the integrity and reliability of the assertions, which is essential for supporting trust and interoperability within the ecosystem.
+
+## Test Cases
```
bumblefudge, line 331: Mmm, these aren't test cases so much as possible use-cases.  It might be good to mark this section as non-normative explicitly, or leave it out of the CAIP itself and just add these in, say, the eth-magicians thread promoting the CAIP?

bumblefudge: we use "test cases" more in the sense of "test vectors" -- I don't think you really need this section anyways since the examples provided throughout are pretty illustrative and seem (unless mircea and raphael missed any syntax nits) to be valid both as VCs and as examples of your data model?

CAIPs/caip-261.md
```diff
@@ -0,0 +1,384 @@
...
+- Security,
+- Data integrity,
+- Data ownership,
+- Interoperabiltiy,
+- Scalability,
+- Censorship resistance,
+- Sustainability.
+
+#### Trust Update
+When an issuer needs to update a trust assertion, they simply generate a new assertion containing the revised information.
```
bumblefudge: is there any tombstoning system for assertions that have been superceded? could/should there be, at least optionally? people unfamiliar with IPFS may not realize they have manage their own garbage collection in a system where superceding/updating records are submitted often...

CAIPs/caip-261.md
bumblefudge: YES PLEASE (basically the only change needed before I can approve and merge this)

CAIPs/caip-261.md
bumblefudge: ^ Not blocking changes to merge but highly recommended before this goes FINAL :D

bumblefudge:
I added [a big commit which you might want to review](https://github.com/ChainAgnostic/CAIPs/pull/261/commits/47d9cefbd6d892f9c09c5bba9a2d4e2509eaa933) in case I changed anything you disagree with; almost ready to merge except for adding "trust computer" definition to terminology list and maybe another 1 or 2 minor things int he above conversation?

dayksx (commented):
All good for me.

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+It is noteworthy that [EIP-712][] mandates the presence of all fields, even if some are left empty in order to enable their verification.
+Note that composing [EIP712][] with [W3C Verifiable Credentials][VC] requires the [EthereumEip712Signature2021][] signature suite.
+
+##### Example EIP-712 proof object for use in a W3C Verifiable Credential
+
+```json
+  "proof": {
+      "verificationMethod": "did:pkh:eip155:59144:0x3892967AA898d7EeBf1B08d3E1F31B2F4C84317A#blockchainAccountId",
+      "created": "2024-01-27T07:05:56.273Z",
+      "proofPurpose": "assertionMethod",
```
bumblefudge, line 280:
```suggestion
      "proofPurpose": "assertionMethod",
      "proofValue": "0x...",
```

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+
+### Trust Representation
+
+In the following diagram, we depict the process where an issuer of a long-lived and verifiable document formally asserts their trust in a subject, highlighting the foundational relationship within a peer-to-peer trust-based system.
+
+![diagram1](https://github.com/dayksx/CAIPs/assets/77788154/e14b4bcb-afc2-4578-9e33-1c0c8bb927e2)
+
+#### Identifier Scheme
+
+##### Peers Identification
```
bumblefudge, line 41:
```suggestion
##### Peer Identification
```

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+
+Identification of peers SHOULD make use of key pair-based mechanism to facilitate interoperability and verifiability.
+- **[Decentralized Identifiers (DIDs)][DID]** with key pair-based methods for off-chain assertions.
+  - **did:pkh (Public Key Hash):** This method is designed to be chain-agnostic, enabling the creation of DIDs based on public key hashes from a variety of blockchains. Its universality promotes interoperability and simplifies identity management across diverse ecosystems. This approach is particularly useful for applications that aim to operate across multiple cryptographically-secured networks (like blockchains) without being tied to any specific one.
+  - **did:key:** This method allows for the generation of DIDs directly from cryptographic public keys, supporting various key types such as ECDSA, EdDSA, Schnorr Signatures, and RSA. It is straightforward and self-contained, with no need for interaction with blockchain networks. This makes it fast and cost-effective for scenarios that do not require decentralized verification.
```
bumblefudge, line 48:
```suggestion
  - **did:key:** This method allows for the generation of DIDs directly from cryptographic public keys, supporting various key types such as ECDSA, EdDSA, Schnorr Signatures, and RSA. It is straightforward and self-contained, with no need for interaction with blockchain networks. This makes it fast and cost-effective for scenarios that do not require on-chain verification or cross-chain use-cases using keytypes not available in the target verifier's toolchain.
```

UCANs use `did:key`s for decentralized verification! The question of whether or not verification is decentralized is orthogonal to how the key is expressed; best to get more precise about who needs to use which did expression where, whether my suggestion works or not.

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+- **On-chain Public Accounts** using the chain-specific address type for smart contract-based assertions (e.g., `address` type in EVM-based chains) 
+
+##### Assertions Identification
+
+Assertions themselves need to be identifiable for referencing or updating purposes. 
+They SHOULD be identified by a immutable identifier.
```
bumblefudge, line 54:
```suggestion
They SHOULD be identified by an immutable identifier.
```

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+- **[Content Identifiers (CID)][CID]**, which is self-describing and facilitates their unique identification and retrieval. If a CID is used, the CID SHOULD be envelopped as a Uniform Resource Identifier (URI) to offer a contextual identifier, such as IPFS (ipfs://).
```
bumblefudge, line 55:
```suggestion
- **[Content Identifiers (CID)][CID]**, which is self-describing and facilitates their unique identification and retrieval. If a CID is used, the CID SHOULD be expressed as a Uniform Resource Identifier (URI) to offer a contextual identifier, such as the IPFS scheme if using an IPFS CID (`ipfs://`).
```

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+  - This is not necessarily embedded as a property within the document itself; it derives from the content of the document and can be calculated (or recalculated, as a checksum of its contents) by anyone.
```
bumblefudge, line 56:
```suggestion
  - This is not necessarily embedded as a property within the document itself; it derives from the content of the document and can be calculated (or recalculated, as a checksum of its contents) by anyone with the document and the checksum algorithm used to create it (e.g. the default canonicalization and encoding of IPFS `kubo`).
```

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+  - This is generated by hashing the content with a cryptographic hash function; This hash, along with metadata about the hash function, is then encoded in the [multihash][] format.
+- **On-chain custom identifiers** generated and assigned by a smart contract based assertion storage, i.e. on-chain attestations identifiers (e.g., Verax, EAS). 
+
+#### Data Model
+
+Subsequent data structures are represented in accordance with the [Verifiable Credential Data Model](https://www.w3.org/TR/vc-data-model/) and, in some cases, in raw format relevant for on-chain assertions. 
+
+Verifiable credentials can be used both off-chain and on-chain for better portability, while raw formats can be utilized in decentralized data storages (e.g., Ceramic Network) or on-chain assertion storages (e.g., Verax, EAS) as a new schema to issue assertions without the need for additional layers of verifiability and interoperability.
```
bumblefudge, line 64:
```suggestion
Verifiable credentials can be used both off-chain and on-chain for better portability, while raw formats can be utilized in decentralized data storage systems (e.g., Ceramic Network) or on-chain assertion systems (e.g., Verax, EAS) as a registered schema to issue assertions without the need for additional layers of verifiability and interoperability.
```

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+However, this standard does not prescribe any specific document type, though it recommends using internationally recognized standards or trusted attestation services."
+
+![diagram2](https://github.com/dayksx/CAIPs/assets/77788154/479df06f-3dd5-4c88-b396-4cfae01ff8d6)
+
+**Peer Trust Assertions:**
+
+Modeling trust and distrust towards an individual or entity can be a complex task due to the subjective and multifaceted nature of trust.
+This standard proposes the following conceptualization for the trust concept:
+
+- `scope`: This defines the applicable trust perimeter. It SHOULD be a noun. Any implementation or profile of this specification COULD enumerate or provide reference lists of valid "scope" values to facilitate interoperability across different systems.
```
bumblefudge, line 75:
```suggestion
- `scope`: This defines the applicable trust perimeter. It SHOULD be a noun. Any implementation or profile of this specification SHOULD enumerate or provide reference lists of valid "scope" values to facilitate interoperability across different systems.
```
The IETF options are only SHOULD and MAY, not COULD; also, I feel strongly that it's a SHOULD here, not a MAY.

https://www.rfc-editor.org/rfc/rfc2119

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+- `level`: This defines the extent of trust. It MUST remain within the following range: [-1,1]. This could be translated as follows: 'Very low' (-1), 'Low' (-0.5), 'Neutral' (0), 'High' (0.5), 'Very High' (1);
+- `reason` (optional): This defines the motivation of the trust. It MAY be based on some tagging system or other dictionary shared across producers and consumers to guarantee interoperability;
+
+**Assertion of trust to a peer:**
```
bumblefudge, line 79:
```suggestion
**Assertion of trust to a peer:**

```

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+    }
+  ]
+},
+"credentialSchema": [{
+  "id": "ipfs://QmcwYEnLysTyepjjtJw19oTDwuiopbCDbEcCuprCBiL7gl",
+  "type": "JsonSchema"
+},
+"proof": {}
+```
+*Raw  Format*
```
bumblefudge, line 113:
```suggestion

**Raw  Format**

```

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+Trust, being inherently dynamic, can be managed within the document by specifying a validity period, after which the credential must be renewed to maintain its validity.
+A validity limitation can be introduced with the fields `validFrom` and `validUntil`.
+
+```json
+"@context": ["https://www.w3.org/2018/credentials/v1"],
+"type": ["VerifiableCredential", "PeerTrustCredential"],
+"issuanceDate": "2024-02-15T07:05:56.273Z",
+"issuer": "did:pkh:eip155:1:0x44dc4E3309B80eF7aBf41C7D0a68F0337a88F044",
+"validFrom": "2024-01-01T19:23:24Z",
+"validUntil": "2025-01-01T19:23:24Z",
```
bumblefudge, line 200:
```suggestion
{
    "@context": ["https://www.w3.org/2018/credentials/v1"],
    "type": ["VerifiableCredential", "PeerTrustCredential"],
    "issuanceDate": "2024-02-15T07:05:56.273Z",
    "issuer": "did:pkh:eip155:1:0x44dc4E3309B80eF7aBf41C7D0a68F0337a88F044",
    "validFrom": "2024-01-01T19:23:24Z",
    "validUntil": "2025-01-01T19:23:24Z"
}
```

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+```
+
+##### On-chain assertions vaidity period
```
bumblefudge, line 203:
```suggestion
##### On-chain assertions validity period

```

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+Most on-chain assertion storages manage expiration times natively without requiring changes to the overall schema.
```
bumblefudge, line 204:
```suggestion
Many on-chain assertion storage systems manage expiration times natively without requiring changes to the overall schema.
```

I find this a little confusing-- are you saying `validFrom` and `validUntil` aren't needed if you use a storage mechanism that already garbage-collections the assertions and makes them 404 after expiration? I would word this different if so:

```suggestion
Many on-chain assertion storage systems manage expiration times natively, and in implementations relying exclusively on systems that do so, adding explicit validity data to each assertion may be redundant and unnecessary, unless its for forward-compatibility or interoperability with outside verifiers.
```

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+
+### Trust Management
+
+Initially, issued trust assertions should be properly persisted to ensure their availability according to the use-case.
+Subsequently, as trust between peers evolves over time, this proposal outlines a method for managing the entire lifecycle of these assertions, encompassing their creation, update, and eventual revocation.
```
bumblefudge, line 209:
```suggestion
Subsequently, as trust between peers evolves over time, this proposal outlines a method for managing the entire lifecycle of these assertions, encompassing their creation, update, and eventual revocation.
```
why is revocation eventual? would `potential` or `optional` be more precise?

bumblefudge, line 209:
```suggestion
Subsequently, as trust between peers evolves over time, this proposal outlines a method for managing the entire lifecycle of these assertions, encompassing their creation, update, and optional revocation.
```

bumblefudge: eventual would mean it happens to every record here, which is not the case and could lead readers to confuse (non-guaranteed!) expiry with manual/intentional early revocation...

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+
+#### Trust Persistance
+
+Trust Assertions issued by peers compose into a verifiable trust graph that can be consumed for trustless reputation-based use cases.
+To fulfill its purpose, these assertions SHOULD be persisted using mechanisms that provide these properties relative to the given use-case requirements such as Security, Data integrity, Data ownership, Interoperabiltiy, Composability, Scalability, Censorship resistance, Sustainability...
```
bumblefudge, line 214:
```suggestion
To fulfill its purpose, these assertions SHOULD be persisted using mechanisms that provide these properties without compromising the given use-case's other technical requirements, such as security, data integrity, data ownership, defined interoperability, composability, scalability, censorship-resistance, sustainability, etc.
```

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+
+#### Trust Update
+
+When an issuer needs to update a trust assertion, they simply generate a new assertion containing the revised information.
+This new assertion, when issued, MUST supersede any previous assertions of the same type, issued by the same entity, and pertaining to the same subject.
+
+The new assertion COULD refer to the overwritten assertion by adding its identifier in the top level of the data structure
```
bumblefudge, line 221:
```suggestion
The new assertion COULD refer to the overwritten assertion by adding its identifier as value for the `previousVersion` key at the top level of the assertion's data structure:

```

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+```json
+    ""previousVersion":": "ipfs://QmcwYEnWysTyepjjtJw19oTDwuiopbCDbEcCuprCBiL7gt",
```
bumblefudge, line 223:
```suggestion
{
    ...
    "previousVersion": "ipfs://QmcwYEnWysTyepjjtJw19oTDwuiopbCDbEcCuprCBiL7gt",
    ...
}
```
lol looks like your JSON got wierdly escaped here

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+```
+
+#### Trust Revocation
+
+When an issuer decides to revoke a previously issued trust assertion without replacing it, they should generate a new assertion that includes a `credentialStatus` attribute, with a `statusPurpose` set to `revocation`.
```
bumblefudge, line 228:
```suggestion
When an issuer decides to revoke a previously issued trust assertion without replacing it, they should generate a new assertion that includes a `credentialStatus` object whose `id` key contains the identifier for the revoked assertion, and whose `statusPurpose` key set to the literal `revocation`.
```

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+This new assertion should explicitly reference the assertion being revoked.
```
bumblefudge, line 229:
```suggestion
This new assertion should explicitly reference the assertion being revoked using the `credentialStatus.id` mechanism ONLY, and not the `previousVersion` used for update assertions.
```

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+"type": ["VerifiableCredential", "PeerTrustCredential"],
+"issuanceDate": "2024-02-29T14:31:56.273Z",
+"issuer": "did:pkh:eip155:1:0x44dc4E3309B80eF7aBf41C7D0a68F0337a88F044",
+"credentialStatus": {
+    "id": "ipfs://QmcwYEnWysTyepjjtJw19oTDwuiopbCDbEcCuprCBiL7gt",
+    "type": "CredentialStatus",
+    "statusPurpose": "revocation",
+},
+```
+##### On-chain assertions revocation
```
bumblefudge, line 245:
```suggestion

##### On-chain assertions revocation
```

sorry i've become obsessive about linting markdown lately because it sometimes displays wrong in jekyll if you don't have the newlines just right

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+Most on-chain assertion storages manage revocation natively without requiring changes to the overall schema.
```
bumblefudge, line 246:
```suggestion
Many on-chain assertion storage systems manage revocation natively (i.e. as deletions of the original assertion) and thus explicit revocation may be redundant unless forward-compatibility or interoperability with other corpuses are targeted.
```

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+
+#### Trust Graph Expansion
+
+##### Implicit Trust Assertions
+
+While the Peer Trust Assertion model outlined previously focuses on explicit trust signals, the trust graph can be further enriched with additional implicit on-chain and/or off-chain trust signals related to its peers. These can include Proof of Humanity, Proof of Membership, Proof of Contributions, Proof of Attendance, Social Graphs, among others.
```
bumblefudge, line 252:
```suggestion
While the Peer Trust Assertion model outlined previously focuses on explicit trust signals, the trust graph can be further enriched with additional implicit on-chain and/or off-chain trust signals related to its peers. These can include proof of humanity, proof of membership, proof of contribution, proof of attendance, or other forms of open-data query, including social-graph checks.
```

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+
+```json
+  "proof": {
+      "verificationMethod": "did:pkh:eip155:59144:0x3892967AA898d7EeBf1B08d3E1F31B2F4C84317A#blockchainAccountId",
+      "created": "2024-01-27T07:05:56.273Z",
+      "proofPurpose": "assertionMethod",
+      "type": "EthereumEip712Signature2021"
+  }
+```
+##### On-chain assertions verification
```
bumblefudge, line 284:
```suggestion
##### On-chain assertion verification

```

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+On-chain the verification is automaticaly done by the trusted protocol, but can be verified according to the protocol specificities.
```
bumblefudge, line 285:
```suggestion
When computed "on-chain" (i.e. in a node runtime), the verification logic is automaticaly handled by core transaction logic, but can also be verified independently by mirroring this logic elsewhere or after the fact from on-chain records.
```
I'm not sure if this was exactly what you meant, but if not, please be more explicit about whatever that was, it's cryptic in its current form :D

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+
+#### Format Verification
+
+Assertions MUST adhere to the predefined schema to be deemed valid.
+For example, in the context of a [W3C Verifiable Credential][VC], the document MUST include a 'credentialSchema' as a top-level property, with a value that dereferences to a schema known to the verifier and to which the document conforms.
```
bumblefudge, line 290:
```suggestion
For example, in the context of a [W3C Verifiable Credential][VC], the document MUST include an array of 'credentialSchema' objects as a top-level property, one or more of which contains a value that dereferences to a schema known to the verifier, and to which the document conforms.

```

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+Such a property provides verifiers with the necessary information to assess whether the data presented conforms to the established schema(s).
+
+```json
+"@context": ["https://www.w3.org/2018/credentials/v1"],
+  "credentialSchema": [{
+    "id": "ipfs://QmcwYEnLysTyepjjtJw19oTDwuiopbCDbEcCuprCBiL7gl",
+    "type": "JsonSchema"
+  },
```
bumblefudge, line 298:
```suggestion
{
    "@context": ["https://www.w3.org/2018/credentials/v1"],
    "credentialSchema": [{
        "id": "ipfs://QmcwYEnLysTyepjjtJw19oTDwuiopbCDbEcCuprCBiL7gl",
        "type": "JsonSchema"
    }],
    ...
}
```

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+- Verifying if the assertion has been formally revoked, and
+- Confirming that the assertion is within its designated validity period and has not expired.
+
+### Trust Assertions Consumption
+
+After the verification process, the consumer can utilize the trust graph for generating insights relevant to their specific use-case.
+
+#### Processing
+
+The consumption of the trust graph typically involves recursive calculations across the graph, following specific rules, to derive peers insights such as reputation scores.
```
bumblefudge, line 317:
```suggestion
The consumption of the trust graph typically involves recursive calculations across the graph, following specific rules, to derive peer insights such as reputation scores.

```

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+This process can utilize various existing or newly developed algorithms tailored to the unique requirements of the use-case.
+This specification remains unopiniated regarding this processing.
+
+However, this specification offers some guidelines to ensure integrity and transparency in the processing of the trust graph information:
```
bumblefudge, line 321:
```suggestion
However, this specification offers some guidelines to ensure integrity and transparency in the processing of the trust graph information:

```

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+    "trustValue": 0.19191918793049725,
+    "trustResult": 0.7,
+    "creationAt": "2024-02-27T10:28:00.000Z",
+    "trustScoreType": "EigenTrust"
+  },
+  
+},
+"proof": {}
+```
+*Raw Format*
```
bumblefudge, line 357:
```suggestion

**Raw Format**

```

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+  "trustScore": {
+    "trustScoreScope": ["SoftwareDevelopment"],
+    "trustValue": 0.19191918793049725,
+    "trustResult": 0.7,
+    "creationAt": "2024-02-27T10:28:00.000Z",
+    "trustScoreType": "EigenTrust"
+  },
+  
+},
+"proof": {}
```
bumblefudge, line 355:
```suggestion
{
  "@context": ["https://www.w3.org/2018/credentials/v1"],
  "type": ["VerifiableCredential", "PeerTrustScorCredential"],
  "issuanceDate": "2023-11-24T12:24:42Z",
  "issuer": "did:pkh:eip155:1:0x23d86aa31d4198a78baa98e49bb2da52cd15c6f0",
  "credentialSubject":
  {
    "id": "did:pkh:eth:0x44dc4E3309B80eF7aBf41C7D0a68F0337a88F044",
    "trustScore": {
      "trustScoreScope": ["SoftwareDevelopment"],
      "trustValue": 0.19191918793049725,
      "trustResult": 0.7,
      "creationAt": "2024-02-27T10:28:00.000Z",
      "trustScoreType": "EigenTrust"
    }
  },
  proof: {
    ...
  }
```

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+```json
+"trustScoreScope": ["SoftwareDevelopment"],
+"trustValue": 0.19191918793049725,
+"trustResult": 0.7,
+"creationAt": "2024-02-27T10:28:00.000Z",
+"trustScoreType": "EigenTrust"
+```
+
+*The above example represents an assertion issued by the issuer `did:pkh:eip155:1:0x23d86aa31d4198a78baa98e49bb2da52cd15c6f0`.
+It claims a certain level of trust to the subject `did:pkh:eip155:1:0x44dc4E3309B80eF7aBf41C7D0a68F0337a88F044` in the area of software development calculated with the EigenTrust algorithm.*
```
bumblefudge, line 367:
```suggestion
It claims a certain level of trustworthiness for the subject `did:pkh:eip155:1:0x44dc4E3309B80eF7aBf41C7D0a68F0337a88F044` in the area of software development calculated with the EigenTrust algorithm.*
```

CAIPs/caip-261.md
```diff
@@ -0,0 +1,466 @@
...
+Trust in an individual or entity is based on their qualities, or their abilities; it is not binary and evolves over time.
+
+Distrust assertions allow for the capture of suspicious behaviors or subjective impressions.
+
+### Verification
+
+The verification process is crucial to ensure the use of valid and up-to-date assertions, thus maintaining the trust system's reliability and accuracy.
+This step ensures the integrity and reliability of the assertions, which is essential for supporting trust and interoperability within the ecosystem.
+
+## Test Cases
```
bumblefudge, line 400:
```suggestion
## Use Cases

*This section is non-normative.*
```

bumblefudge (changes requested):
LGTM-- any suggestion you disagree with, _but understand the gist of_ and can replace with another clarification, just resolve yourself, and ping me async if you want further explanation of any you _can't understand_!

CAIPs/caip-261.md
dayksx: That's correct. I just wanted to say that trusting the blockchain implementation and immutability make the computation trustless without the need of verification at each compute

bumblefudge (approved):
LGTM, 2nd reviewer and we're golden.  inviting @TimDaub to review because I know he's thought more than I have about EAS and similar onchain record systems, and may have good insights on those sections.

bumblefudge:
@obstropolos got a minute? slower month this time

CAIPs/caip-261.md
```diff
@@ -0,0 +1,492 @@
...
+
+In the following diagram, we depict the process where an issuer of a long-lived and verifiable document formally asserts their trust in a subject, highlighting the foundational relationship within a peer-to-peer trust-based system.
+
+![diagram1](https://github.com/dayksx/CAIPs/assets/77788154/e14b4bcb-afc2-4578-9e33-1c0c8bb927e2)
+
+#### Identifier Scheme
+
+##### Peers Identification
+
+Trust assertions are made about, and by, peers identifiable with identifiers derived from public keys or directly using a public key. This enables peers to sign assertions and be verified.
```
obstropolos, line 43:
```suggestion
Trust assertions are made about, and by, peers identifiable with identifiers derived from public keys or directly using a public key. This enables peers to sign and verify assertions.

```

CAIPs/caip-261.md
```diff
@@ -0,0 +1,492 @@
...
+  - **did:key:** This method allows for the generation of DIDs directly from cryptographic public keys, supporting various key types such as ECDSA, EdDSA, Schnorr Signatures, and RSA. It is straightforward and self-contained, with no need for interaction with blockchain networks. This makes it fast and cost-effective for scenarios that do not require decentralized verification.
+- **On-chain Public Accounts** using the chain-specific address type for smart contract-based assertions (e.g., `address` type in EVM-based chains) 
+
+##### Assertions Identification
+
+Assertions themselves need to be identifiable for referencing or updating purposes. 
+They SHOULD be identified by a immutable identifier.
+- **[Content Identifiers (CID)][CID]**, which is self-describing and facilitates their unique identification and retrieval. If a CID is used, the CID SHOULD be expressed as a Uniform Resource Identifier (URI) to offer a contextual identifier, such as the IPFS scheme if using an IPFS CID (`ipfs://`).
```
obstropolos, line 55:
```suggestion
- **[Content Identifiers (CID)][CID]**, which are self-describing and facilitates their unique identification and retrieval. If a CID is used, the CID SHOULD be expressed as a Uniform Resource Identifier (URI) to offer a contextual identifier, such as the IPFS scheme if using an IPFS CID (`ipfs://`).

```

obstropolos (approved):

PR #268, opened 2024-03-08
Extend CAIP-10 with CAIP-69, designating chainID 0 for EOAs

GriffGreen:


CAIPs/caip-69.md
```diff
@@ -0,0 +1,72 @@
...
+The motivation for proposal stems from the redundancy of assigning an EOA
+as living on every chain that exists. As new chains continue to be launched
+it becomes more and more obvious that assigning an EOA its own specific
+designation will greatly simplify their use in cross chain applications.
+
+
+## Specification
+
+The account id specification will follow CAIP-10, but for EOAs use the 
+prefix `eip155:10:` before teh EOA account address
```
GregTheGreek, line 41: Potential type, do you mean `0` instead of `10`.

I would suggest also adding the case for `1` to be designated as a smart account?

CAIPs/caip-69.md
```diff
@@ -0,0 +1,72 @@
+---
+caip: 69
+title: EOA extension to CAIP-10
+author: Griff Green (@griffgreen), Krati Jain (@jainkrati), Greg-the-Greek (Please edit Greg!)
```
GregTheGreek, line 4:
```suggestion
author: Griff Green (@griffgreen), Krati Jain (@jainkrati), Gregory Markou (@GregTheGreek)
```

CAIPs/caip-69.md
```diff
@@ -0,0 +1,72 @@
...
+CAIP-10 defines a way to identify an address in our multichain world using
+CAIP-2 blockchain id. CAIP-69 extends it to include a special case for EOAs
+using the chain ID of 0.
+
+
+## Abstract
+
+This proposal aims to facilitate specifying EOAs as multichain accounts
+extending CAIP-10 chain id specification. This is useful for both
+decentralized applications to communicate that EOAs can be used for multiple
```
jainkrati, line 22: Possibly you meant this: " Decentralized applications can use this to specify EOA as an address with chainID 0"

CAIPs/caip-69.md
jainkrati: +1

jainkrati: Greg's idea of representing contracts or non-EOA accounts also sounds good. as per CAIP-10, eip155:1:<address> represents ethereum mainnet address. so the case for 1 is already taken. we would need alternative for the case of contract.

GregTheGreek: Actually we're fine, because CAIP-10 only contains one colon, where this CAIP contains two. So you would just need to pre-check the string for number colons to perform the match :)

GregTheGreek:
> I think this makes sense, but in my opinion it should be a PR on the [namespaces](https://github.com/ChainAgnostic/namespaces/) repo.
> 
> In particular I think it could just be added as a new section to [eip155/caip10](https://github.com/ChainAgnostic/namespaces/blob/main/eip155/caip10.md)

Oh wow I didn't know this existed... 

Yep, I think we should namespace against EVM based networks **only**

CAIPs/caip-69.md
GriffGreen: > Potential type, do you mean `0` instead of `10`.
YEP! Fixing!


> I would suggest also adding the case for `1` to be designated as a contract, or non-eoa account? If so lets also add that to the examples

Hmmmm would there be a challenge for backwards compatibility? I think its cleaner to reserve the chainID of 0 for EOAs than to add another param to the standard... however, it might make sense if this is a pattern that we can use for other ecosystems like cosmos or polkadot.

CAIPs/caip-69.md
GriffGreen: Great improvement, will revise

GriffGreen:
> I think this makes sense, but in my opinion it should be a PR on the [namespaces](https://github.com/ChainAgnostic/namespaces/) repo.

GREAT CALL! We will add it there too.


> In particular I think it could just be added as a new section to [eip155/caip10](https://github.com/ChainAgnostic/namespaces/blob/main/eip155/caip10.md)

Yeah we were wondering if we can change it or not since it is marked "Final" I made the assumption that it would need to be extended with a new CAIP instead of amended.... but honestly a new paragraph there is great, if we can make it as simple as that.

oed:
I think we probably need @pedrouid's input on the namespaces PR, but pretty sure it wouldn't break anything. Worst case we can figure out a different way of doing an extension spec in the namespaces.

pedrouid:
Honestly I don't have any objection with allowing ZERO as a chainId to flag EOAs

But this definitely should not be its own CAIP and instead it should be part of the EVM namespace profile

Let's move this conversation to ChainAgnostic/namespaces

bumblefudge:
There isn't a ton of precedent for opening PRs on eip155/caipX.md files, so i'm happy to help on github or in DM/backchannel, @GriffGreen ! I like the basic idea and implementers have been asking for some kind of wildcard/offchain entry in the CAIP-10 namespace for a while, so I'm glad to see someone proposing a path forward.

bumblefudge:
@ligi has anyone ever tried registering a NEGATIVE chainId, or a float/decimal chainId?  it's kind of odd how little eip155 actually says about the data type or domain of chainId qua variable.

@GregTheGreek wow i just saw this for the first time.  you might need to update it or withdraw it, lest some trickster add themself as co-author and revive it 😅 
 https://eips.ethereum.org/EIPS/eip-3788

ligi:
No I have never seen a negative or floating point chainId - my CI for ethereum-lists/chains should also fail in this case.

GregTheGreek:
> @GregTheGreek wow i just saw this for the first time. you might need to update it or withdraw it, lest some trickster add themself as co-author and revive it in a way that confuses or confounds OP's proposed behavior 😅 https://eips.ethereum.org/EIPS/eip-3788

Yaaa I should withdraw that...

CAIPs/caip-69.md
jainkrati: yes, in Test Cases section in https://chainagnostic.org/CAIPs/caip-10, all addresses have two colons. Would appreciate if you could clarify if there are gaps in our understanding @GregTheGreek

jainkrati:
> There isn't a ton of precedent for opening PRs on eip155/caipX.md files, so i'm happy to help on github or in DM/backchannel, @GriffGreen ! I like the basic idea and implementers have been asking for some kind of wildcard/offchain entry in the CAIP-10 namespace for a while, so I'm glad to see someone proposing a path forward.


hey @bumblefudge ! I would like to raise the PR on behalf of @GriffGreen , in the correct repo.. could you please help

bumblefudge:
Closing because addressed by ^ the corresponding PR (now merged) on `/namespaces`

PR #269, opened 2024-03-12
WIP - create extra /final/ permalinks for final CAIPs

bumblefudge:
digging around in jekyll docs, I realize this wasn't easily possible with our setup, so I made a little workaround that requires a little manual intervention when a given CAIP goes to final.

leaving this open in case it's needed, and in case someone knows a cleaner way to achieve the same ends.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,18 +1,18 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    addressable (2.8.1)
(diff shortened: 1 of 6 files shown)


bumblefudge:
For context see https://github.com/ethcatherders/EIPIP issue 320

obstropolos:
@ligi  - whenever you have some time can you take a look here / quick review?

PR #280, opened 2024-05-27
Errata: update links section of CAIP-74

bumblefudge:
No normative changes-- just bringing CAIP-74 up to date with the links section of the styleguide that was added to contributing.md since it was ratified.


obstropolos (approved):
Thanks for the updates here

PR #282, opened 2024-05-30
Browser wallet messaging interface

pedrouid:


CAIPs/caip-282.md
```diff
@@ -0,0 +1,198 @@
...
+
+Standardized messaging for wallet interface in browser environments.
+
+## Abstract
+
+When interfacing with decentralized applications users install wallets in their browser to manage their blockchain accounts which apps require to sign messages and transactions. Leveraging existing browser messaging APIs these are are used for messaging to initiate a dapp-wallet connection in a browser environment.
+
+## Motivation
+
+Currently, decentralized applications are required to install a multitude of SDKs to support many different wallet providers to broaden its userbase with different wallets of choice. However this creates a big overhead for applications to support more and more proprietrary SDKs that increase its size and laso creates vendor lockin for major wallet providers that can allocate resources to reach more applications.
```
chris13524, line 22:
```suggestion
Currently, decentralized applications are required to install a multitude of SDKs to support many different wallet providers to broaden its userbase with different wallets of choice. However this creates a big overhead for applications to support more and more proprietrary SDKs that increase its size and also creates vendor lockin for major wallet providers that can allocate resources to reach more applications.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,198 @@
...
+
+Users are already installing a wallet applicationg in their devices and given its presence in the browser environment it should be unnecessary for application developers to also have to install extra software to support these wallet providers. This situation is only present due to lack of standardization and interoperability between the interfaces and also discovery mechanisms for different wallet providers.
```
chris13524, line 24:
```suggestion
Users are already installing a wallet application on their devices and given its presence in the browser environment it should be unnecessary for application developers to also have to install extra software to support these wallet providers. This situation is only present due to lack of standardization and interoperability between the interfaces and also discovery mechanisms for different wallet providers.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,198 @@
...
+
+Both wallet providers and blockchain libraries must listen to incoming messages that might be published after their initialization. Additionally both wallet providers and blockchain libraries must publish a message to both announce themselves and their intent to connect, respectively.
+
+Here is the expected logic from the blockchain library:
+
+```js
+const wallets = {};
+
+// blockchain library starts listening on init
+window.addEventListener("caipxxx:announceWallet", (event) => {
```
chris13524, line 62: Would this be this CAIP?

```suggestion
window.addEventListener("caip282:announceWallet", (event) => {
```

pedrouid: Good point! I've updated all event names

CAIPs/caip-282.md
```diff
@@ -0,0 +1,198 @@
+---
+caip: 282
+title: Browser Wallet Messaging Interface
```
arein, line 3:
```suggestion
title: Chain Agnostic Browser Wallet Messaging Interface
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,198 @@
...
+discussions-to: https://github.com/ChainAgnostic/CAIPs/pull/282
+status: Draft
+type: Standard
+created: 2023-05-30
+requires: 2, 10, 25, 27, 222, 275
+---
+
+## Simple Summary
+
+Standardized messaging for wallet interface in browser environments.
```
arein, line 14:
```suggestion
Standardized messaging interface for browser wallets.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Abstract
+
+When interfacing with decentralized applications users install wallets in their browser to manage their blockchain accounts which apps require to sign messages and transactions. Leveraging existing browser messaging APIs these are are used for messaging to initiate a dapp-wallet connection in a browser environment.
```
arein, line 18:
```suggestion
When interfacing with decentralized applications users install browser wallets to manage their blockchain accounts which apps require to sign messages and transactions. Leveraging existing browser messaging APIs these are are used for messaging to initiate a dapp-wallet connection in a browser environment.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Motivation
+
+Currently, decentralized applications are required to install a multitude of SDKs to support many different wallet providers to broaden its userbase with different wallets of choice. However this creates a big overhead for applications to support more and more proprietrary SDKs that increase its size and also creates vendor lockin for major wallet providers that can allocate resources to reach more applications.
```
arein, line 22:
```suggestion
Currently, in order for decentralized applications to be able to support all users they need to support all browser wallet APIs. Similarly, in order for browser wallets to support all decentralized applications they need to support all APIs. This is not only complicated but also results in a larger bundle size of applications.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,198 @@
...
+
+Wallet Provider: A user agent that manages accounts and facilitates transactions with a blockchain.
+
+Decentralized Application (DApp): A web page that relies upon one or many Web3 platform APIs which are exposed to the web page via the Wallet.
+
+Blockchain Library: A library or piece of software that assists a DApp to interact with blockchain and interface with the Wallet.
+
+### Messaging APIs
+
+The browser has available two APIs that can be used for messaging across different parts of the stack. Using `window.addEventListener` and `window.postMessage` enables communication between browser window, iframes and extensions.
```
arein, line 46:
```suggestion
The browser exposes two APIs that can be used for messaging across different parts of the stack. Using `window.addEventListener` and `window.postMessage` enables communication between browser window, iframes and extensions.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,198 @@
...
+  uuid: string;
+  name: string;
+  icon: string;
+  rdns: string;
+}
+```
+
+The parameters `name` and `icon` are used to display to the user to be easily recognizable while the `rdns` and `uuid` are only used internally for de-duping while they must always be unique, the `rdns` will always be the same but `uuid` is ephemeral per browser session.
+
+#### Connection
```
arein, line 121:
```suggestion
#### Handshake
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,198 @@
...
+
+TODO
+
+## Privacy Considerations
+
+TODO
+
+## Backwards Compatibility
+
+TODO
```
arein, line 179: Probably needs to be a gradual migration where for the next ~3yrs apps/wallets need to support the old/new standards.

1. As an EVM/Solana/Bitcoin/etc browser wallet I want to support all old standards and this one
2. As EVM/Solana/Bitcoin/etc app I want to support all wallet standards

CAIPs/caip-282.md
```diff
@@ -0,0 +1,198 @@
...
+
+In this proposal, we present a solution that focused on optimizing interoperability for multiple Wallet Providers and fostering fairier competition by reducing the barriers to entry for new Wallet Providers, along with enhancing the user experience across all blockchain networks.
+
+## Specification
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in [RFC-2119].
+
+### Definitions
+
+Wallet Provider: A user agent that manages accounts and facilitates transactions with a blockchain.
```
bumblefudge, line 38: I think it might be worth distinguishing the *wallet* (a piece of software the end-user/account-controller interacts with directly and trusts) from the *wallet provider* (which I take here to mean the javascript abstraction that lets the dapp talk to 1 or more wallets without knowing too much about the wallet).  while a wallet provider might have plenty of additional features (some of my best friends are wallet providers!), it is still basically a bundle of browser APIs at its basic, definitional level, right?

CAIPs/caip-282.md
pedrouid: This feels redundant given that it is a CAIP

CAIPs/caip-282.md
pedrouid: This depends on the definition of a browser wallet... personally I classify a browser wallet as a wallet that is integrated into the browser either natively or as an extension

Would like to avoid limiting this standard to only those wallets since this should support any wallet that can access the browser environment including embedded wallets, cloud wallets, iframe wallets and remote wallets

CAIPs/caip-282.md
pedrouid: Merging this one but still would like to clarify the definition of a "browser wallet"

CAIPs/caip-282.md
pedrouid: Also I would like to add more context also on the definition what classifies as a "browser wallet"

CAIPs/caip-282.md
pedrouid: Correct... we should mention existing standards or patterns that require migration to adopt this standard

CAIPs/caip-282.md
```diff
@@ -0,0 +1,385 @@
...
+discussions-to: https://github.com/ChainAgnostic/CAIPs/pull/282
+status: Draft
+type: Standard
+created: 2023-05-30
+requires: 2, 10, 25, 27, 222, 275
+---
+
+## Simple Summary
+
+Standardized messaging interface for browser wallets.
```
obstropolos, line 14:
```suggestion
CAIP-282 defines a standardized messaging interface for browser wallets.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,385 @@
...
+
+## Abstract
+
+When interfacing with decentralized applications users install browser wallets to manage their blockchain accounts which apps require to sign messages and transactions. Leveraging existing browser messaging APIs these are are used for messaging to initiate a dapp-wallet connection in a browser environment.
```
obstropolos, line 18:
```suggestion
To interface with decentralized applications (dapps), users install browser wallets to manage their blockchain accounts, which are required to sign messages and transactions. Leveraging existing browser messaging APIs, these are used to initiate a dapp-wallet connection in a browser environment.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,385 @@
...
+
+## Motivation
+
+Currently, in order for decentralized applications to be able to support all users they need to support all browser wallet APIs. Similarly, in order for browser wallets to support all decentralized applications they need to support all APIs. This is not only complicated but also results in a larger bundle size of applications.
+
+Users are already installing a wallet application on their devices and given its presence in the browser environment it should be unnecessary for application developers to also have to install extra software to support these wallet providers. This situation is only present due to lack of standardization and interoperability between the interfaces and also discovery mechanisms for different wallet providers.
```
obstropolos, line 24:
```suggestion
Users are already installing a wallet application on their devices, and given its presence in the browser environment, it should be unnecessary for application developers to install extra software to support these wallet providers. This situation is only present due to the lack of standardization and interoperability between interfaces and discovery mechanisms for different wallet providers.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,385 @@
...
+
```
obstropolos, line 25: One additional note - ensure consistency with the usage of `wallet provider` OR `Wallet Provider`

CAIPs/caip-282.md
```diff
@@ -0,0 +1,385 @@
...
+This results not only in a degraded user experience but also increases the barrier to entry for new wallet providers as users are incentivized to use more popular wallet providers that are more widely supported in more applications.
+
+This situation is further aggravated by differences between blockchain networks such as Ethereum, Cosmos, Solana, Tezos, etc. While some solutions amttept to solve this such as WalletConnect, EIP-6963, Solana Wallet Protocol, etc. They are not covering all wallets and are not chain-angostic.
```
obstropolos, line 28:
```suggestion
This situation is further aggravated by differences between blockchain networks such as Ethereum, Cosmos, Solana, Tezos, etc. While some solutions attempt to solve this, such as WalletConnect, [EIP-6963][eip-6963], Solana Wallet Protocol, etc., they do not cover all wallets and are not chain-angostic.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,385 @@
...
+
+For EIP155 (Ethereum) ecosystem there are already interfaces for discoverability of browser wallets through either legacy window.ethereum or EIP-6963 events. These existing mechanisms should be supported in parallel in order to receive incoming requests either through this new messaging interface or legacy ones.
+
+Similarly the Solana and BIP122 (Bitcoin) ecosystems have used similar patterns around window.solana and window.bitcoin respectively plus the wallet-standard events. Yet these can also be supported in parallel without conflict with this new messaging interface.
+
+The Wallet Data exposed in this messaging interface is also compatible with EIP-6963 events and wallet-standard events therefore wallet providers can re-use the same identifiers and assets already being used in these existing integrations.
+
+## Links
+
+- [CAIP-2][caip-2] - Blockchain ID Specification
```
obstropolos, line 370:
```suggestion
- [EIP-6963][eip-6963] - Multi Injected Provider Discovery
- [CAIP-2][caip-2] - Blockchain ID Specification
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,385 @@
...
+- [CAIP-10][caip-10] - Account ID Specification
+- [CAIP-27][caip-27] - Blockchain ID Specification
+- [CAIP-25][caip-25] - Blockchain ID Specification
+- [CAIP-222][caip-222] - Account ID Specification
+
+[caip-2]: https://chainagnostic.org/CAIPs/caip-2
```
obstropolos, line 376:
```suggestion
[eip-6963]: https://eips.ethereum.org/EIPS/eip-6963
[caip-2]: https://chainagnostic.org/CAIPs/caip-2
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,385 @@
...
+
+This results not only in a degraded user experience but also increases the barrier to entry for new wallet providers as users are incentivized to use more popular wallet providers that are more widely supported in more applications.
+
+This situation is further aggravated by differences between blockchain networks such as Ethereum, Cosmos, Solana, Tezos, etc. While some solutions amttept to solve this such as WalletConnect, EIP-6963, Solana Wallet Protocol, etc. They are not covering all wallets and are not chain-angostic.
+
+In this proposal, we present a solution that focused on optimizing interoperability for multiple Wallet Providers and fostering fairier competition by reducing the barriers to entry for new Wallet Providers, along with enhancing the user experience across all blockchain networks.
```
obstropolos, line 30:
```suggestion
In this proposal, we present a solution focused on optimizing interoperability for multiple Wallet Providers, fostering fairer competition by reducing the barriers to entry for new Wallet Providers, and enhancing user experience across all blockchain networks.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,385 @@
...
+
+## Specification
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in [RFC-2119].
+
+### Definitions
+
+Wallet Provider: A user agent that manages accounts and facilitates transactions with a blockchain.
+
+Decentralized Application (DApp): A web page that relies upon one or many Web3 platform APIs which are exposed to the web page via the Wallet.
```
obstropolos, line 40: (another one - dapp vs DApp vs dApp vs derp)

CAIPs/caip-282.md
```diff
@@ -0,0 +1,385 @@
...
+
+Blockchain Library: A library or piece of software that assists a DApp to interact with blockchain and interface with the Wallet.
```
obstropolos, line 42:
```suggestion
Blockchain Library: A library or piece of software that assists a DApp to interact with a blockchain and interface with the Wallet.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,385 @@
...
+
+### Messaging APIs
+
+The browser exposes two APIs that can be used for messaging across different parts of the stack. Using `window.addEventListener` and `window.postMessage` enables communication between browser window, iframes and extensions.
+
+This provides the foundation for any wallet provider to interface with a decentralized application using a blockchain library which implements this standard.
+
+There are different loading times that can be affected by multiple factors which makes it non-deterministic to publish and listen messages from different sources within the browser.
```
obstropolos, line 50:
```suggestion
Different loading times can be affected by multiple factors, which makes it non-deterministic to publish and listen to messages from different sources within the browser.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,385 @@
...
+}
+```
+
+The parameters `name` and `icon` are used to display to the user to be easily recognizable while the `rdns` and `uuid` are only used internally for de-duping while they must always be unique, the `rdns` will always be the same but `uuid` is ephemeral per browser session.
+
+#### Handshake
+
+After the wallet has been selected by the user then the blockchain library MUST publish a message to share its intent to establish a connection. This can be either done as a [CAIP-25][caip-25] request or [CAIP-222][caip-222] authentication.
+
+The communication will use the `uuid` shared by the initial wallet provider announcement payload which the wallet provider will be listening to for any incoming requests and consequently the blockchain library will also use for publishing messages. The same will happen again the other way around but vice-versa where the wallet provider will be the blockchain library will be listening to for any incoming responses and consequently the wallet provider will also use for publishing messages.
```
obstropolos, line 123:
```suggestion
The communication will use the `uuid` shared by the initial wallet provider announcement payload, which the wallet provider will listen to for any incoming requests, and consequently, the blockchain library will also be used for publishing messages. The same will happen again the other way around but vice-versa, where the wallet provider will be the blockchain library that will be listening to any incoming responses, and consequently, the wallet provider will also use it for publishing messages.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,385 @@
...
+// wallet provider publishes for reponses
+window.postMessage("message", {
+  event: "caip282:respond:<wallet_provider_uuid>",
+  data: { ... },
+});
+```
+
+#### Signing
+
+This same channel uuid can then be used for a connected session using the [CAIP-27][caip-27] which then would use the sessionId from the established connection to identify incoming payloads that need to be respond and also which chainId is being targetted.
```
obstropolos, line 157:
```suggestion
This same channel `uuid` can then be used for a connected session using [CAIP-27][caip-27] which then would use the `sessionId` from the established connection to identify incoming payloads that need to be respond to, and also which `chainId` is being targetted.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,385 @@
...
+
+#### UUIDs
+
+The generation of UUIDs is crucial for this messaging interface to work seamlessly for the users.
+
+A wallet provider MUST generate UUIDs always distinctly for each webpage loaded and they must not be re-used without a session being established between the application and the wallet with the user's consent.
```
obstropolos, line 163:
```suggestion
A wallet provider MUST always generate UUIDs distinctly for each web page loaded, and they must not be re-used without a session being established between the application and the wallet with the user's consent.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,385 @@
...
+
+A UUID can be re-used as a sessionId if and only if the CAIP-25 or CAIP-222 has been prompted to the user and the user has approved its permissions to allow the application to make future signing requests.
```
obstropolos, line 165:
```suggestion
A UUID can be re-used as a `sessionId` if and only if the [CAIP-25] or [CAIP-222] procedure has been prompted to the user and the user has approved its permissions to allow the application to make future signing requests.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,385 @@
...
+
+Once established the UUID is used as sessionId also for the CAIP-27 payloads which can verify that incoming messages are being routed through the appropriate channels.
```
obstropolos, line 167:
```suggestion
Once established, the UUID is used as `sessionId` for the [CAIP-27] payloads, which can verify that incoming messages are being routed through the appropriate channels.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,385 @@
...
+
+## Rationale
+
+Browser wallets differentiate themselves because they can be installed by users without the application developer require any further integration. Therefore we optimize for a messaging interface that leverages the two-way communication available to browser wallets to make themselves discoverable and negotiate a set of parameters that enable not only easy human readability with a clear name and icon but also machine-readability using strong identifiers with uuid and rdns.
```
obstropolos, line 171:
```suggestion
Browser wallets differentiate themselves because they can be installed by users without the application developer requiring any further integration. Therefore, we optimize for a messaging interface that leverages the two-way communication available to browser wallets to make themselves discoverable, and negotiate a set of parameters that enable not only easy human readability with a clear name and icon but also machine-readability using strong identifiers with uuids and rdns.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,385 @@
...
+
+The choice for using window.postMessage is motivated by expanding the range of wallet providers it can support which would include not only browser extensions that can alternatively use window.dispatchEvent but instead it would cover also Inline Frames, Service Workers, Shared Workers, etc.
```
obstropolos, line 173:
```suggestion
The choice for using `window.postMessage` is motivated by expanding the range of wallet providers it can support, including browser extensions that can alternatively use `window.dispatchEvent` but instead it would also cover Inline Frames, Service Workers, Shared Workers, and more.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,385 @@
...
+
+The use of UUID for message routing is important because while RDNS is useful for identifying the wallet provider, it causes issues when it comes to the session management of different webpages connected to the same wallet provider or even managing stale sessions which can be out-of-sync. Since UUID generation is derived dynamically on page load then wallet providers can track these sessions more granularly rather than making assumptions around webpage URL and RDNS relationship.
```
obstropolos, line 175:
```suggestion
The use of UUID for message routing is important because while RDNS is useful for identifying the wallet provider, it causes issues when it comes to the session management of different webpages connected to the same wallet provider or even managing stale sessions, which can be out-of-sync. Since UUID generation is derived dynamically on page load, wallet providers can track these sessions more granularly rather than making assumptions around the webpage URL and RDNS relationship.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,385 @@
...
+
+The existing standards around provider authorization (CAIP-25) and wallet authentication (CAIP-222) are fundamental to this experience because they create clear intents for a wallet to "connect" with a webpage url after it's been discovered. This standard does not enforce either one but strongly recommends these standards as the preferred interface for connecting or authenticating a wallet.
+
+Finally the use of CAIP-27 leverages the work above to properly target signing requests that are intended to be prompt to wallet users which will include a sessionId and chainId in parallel with the pre-established sessions using either CAIP-25 or CAIP-222
```
obstropolos, line 179:
```suggestion
Finally the use of CAIP-27 leverages the work above to properly target signing requests that are intended to be prompt to wallet users which will include a `sessionId` and `chainId` in parallel with the pre-established sessions using either CAIP-25 or CAIP-222
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,385 @@
...
+
+## Test Cases
+
+Here is a test case where we demonstrate a scenario with logic from both a blockchain library and a wallet provider
```
obstropolos, line 183:
```suggestion
Here is a test case where we demonstrate a scenario with logic from both a blockchain library and a wallet provider.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,385 @@
...
+    id: request.id, // 2
+    jsonrpc: "2.0",
+    result: "0xe670ec64341771606e55d6b4ca35a1a6b75ee3d5145a99d05921026d1527331"
+  },
+});
+```
+
+## Security Considerations
+
+The advantage of using window.postMessage over existing standards that leverage window.dispatchEvent is the prevention of prototype pollution but that still does not mean that there aren't existing attacks that must be considered:
```
obstropolos, line 344:
```suggestion
The advantage of using `window.postMessage` over existing standards that leverage `window.dispatchEvent` is the prevention of prototype pollution, but that still does not mean that there aren't existing attacks that must be considered:
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,385 @@
...
+
+### Wallet Imitation and Manipulation
+
+Application developers are expected to actively detect for misbehavior of properties or functions being modified in order to tamper with or modify other wallets. One way this can be easily achieved is to look for when the uuid property within two Wallet Data objects match. Applications and Libraries are expected to consider other potential methods that the Wallet Data objects are being tampered with and consider additional mitigation techniques to prevent this as well in order to protect the user.
+
+### Prevent SVG Javascript Execution
+
+The use of SVG images introduces a cross-site scripting risk as they can include JavaScript code. This Javascript executes within the context of the page and can therefore modify the page or the contents of the page. So when considering the experience of rendering the icons, DApps need to take into consideration how they’ll approach handling these concerns in order to prevent an image being used as an obfuscation technique to hide malicious modifications to the page or to other wallets.
```
obstropolos, line 352:
```suggestion
The use of SVG images introduces a cross-site scripting risk as they can include JavaScript code. This JavaScript executes within the context of the page and can modify the page or the contents of the page. So, when considering the experience of rendering the icons, DApps need to take into consideration how they’ll approach handling these concerns in order to prevent an image from being used as an obfuscation technique to hide malicious modifications to the page or to other wallets.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,385 @@
...
+
+## Privacy Considerations
+
+Any form of wallet discoverability must alwasys take in consideration wallet fingerprinting that can happen by malicious webpages or extensions that attempt to capture user information. Thus wallet providers can abstain from publishing announceWallet messages on every page load and wait for incoming promptWallet messages. Yet this open the possiblity for race conditions where wallet providers could be initialized after the promptWallet message was published therefore be undiscoverable. It is recommended that if wallet providers do offer this more "private connect" feature that is only enabled optionally by users rather than set by default.
```
obstropolos, line 356:
```suggestion
Any form of wallet discoverability must alwasys take in consideration wallet fingerprinting that can happen by malicious webpages or extensions that attempt to capture user information. Wallet providers can abstain from publishing `announceWallet` messages on every page load and wait for incoming `promptWallet` messages. Yet this opens the possibility for race conditions where wallet providers could be initialized after the `promptWallet` message was published and therefore be undiscoverable. It is recommended that wallet providers offer this more "private connect" feature that users only enable optionally, rather than set by default.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,385 @@
...
+
+## Backwards Compatibility
+
+It's important to note that existing blockchain ecosystems already have standards that overlap with the scope of this standard and backwards-compatibility must be considered for a smooth adoption by both wallet and application developers.
+
+For EIP155 (Ethereum) ecosystem there are already interfaces for discoverability of browser wallets through either legacy window.ethereum or EIP-6963 events. These existing mechanisms should be supported in parallel in order to receive incoming requests either through this new messaging interface or legacy ones.
```
obstropolos, line 362:
```suggestion
For the EIP155 (Ethereum) ecosystem there are already interfaces for the discoverability of browser wallets through either legacy `window.ethereum` or [EIP-6963] events. These existing mechanisms should be supported in parallel in order to receive incoming requests either through this new messaging interface or legacy ones.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,385 @@
...
+
+Similarly the Solana and BIP122 (Bitcoin) ecosystems have used similar patterns around window.solana and window.bitcoin respectively plus the wallet-standard events. Yet these can also be supported in parallel without conflict with this new messaging interface.
```
obstropolos, line 364:
```suggestion
Similarly the Solana and BIP122 (Bitcoin) ecosystems have used similar patterns around `window.solana` and `window.bitcoin` respectively plus the wallet-standard events. Yet these can also be supported in parallel without conflict with this new messaging interface.
```

obstropolos (commented):
A number of feedback across the spec - nothing too deep other than just a few consistency items / etc.

CAIPs/caip-282.md
pedrouid: Agreed! Just did a global find & replace

CAIPs/caip-282.md
```diff
@@ -0,0 +1,405 @@
...
+
+Wallet Provider: A user agent that manages accounts and facilitates transactions with a blockchain.
+
+Decentralized Application (dapp): A web page that relies upon one or many Web3 platform APIs which are exposed to the web page via the Wallet.
+
+Blockchain Library: A library or piece of software that assists a dapp to interact with a blockchain and interface with the Wallet.
+
+### Messaging APIs
+
+The browser exposes two APIs that can be used for messaging across different parts of the stack. Using `window.addEventListener` and `window.postMessage` enables communication between browser window, iframes and extensions.
```
Gudahtt, line 46: For extensions, it's worth clarifying that this only works for communicating with extension contentscripts.  `window.postMessage` can't be used to communicate directly with an extension UI/service worker/background process. For a page to communicate with an extension using `window.postMessage`, you'd need to send a message _to yourself_ and expect the extension to listen in and respond using a contentscript. Essentially the same as if you were communicating with another library in the same process.

Technically an extension contentscript is part of the extension, so this isn't wrong exactly, but it could be misleading to say this and not describe how. It may suffice even just to say "extension contentscript" rather than "extension".

Gudahtt, line 46: How do you imagine this would support cross-frame communication? Notably absent from this document is any mention of the `targetOrigin` parameter of `postMessage`, and the many security risks it has.

It someone read this document and just inserted a `targetOrigin` of `*` to make it work, the results could be disastrous; it would allow any page to snoop on all communication, it would allow malicious sites to reply with fake responses, and it would break poorly designed sites that cannot handle unexpected message events.

To me it seems critically important to provide some guidance about cross-frame communication if we're going to mention it at all.

CAIPs/caip-282.md
```diff
@@ -0,0 +1,405 @@
...
+  uuid: generateUUID(); // eg. "350670db-19fa-4704-a166-e52e178b59d2"
+  name: "Example Wallet",
+  icon: "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12P4//8/w38GIAXDIBKE0DHxgljNBAAO9TXL0Y4OHwAAAABJRU5ErkJggg==";
+  rdns: "com.example.wallet";
+}
+window.addEventListener("caip282:promptWallet", (event) => {
+  // when a prompt message was received then the wallet will announces again
+  window.postMessage("message", {
+    event: "caip282:announceWallet",
+    data,
```
shanejonas, line 300: why not just have `method, params` a la JSON-RPC. caip-27 nests JSON-RPC so nesting here doesn't sound too bad I guess.

CAIPs/caip-282.md
pedrouid: Might be unnecessary since we intend to break this CAIP into multiple CAIPs

CAIPs/caip-282.md
pedrouid: Agreed

pedrouid: Agreed... need to change this

CAIPs/caip-282.md
pedrouid: Just changed it 👍

pedrouid:
Just sharing here some developments on this CAIP... after discussing it with several teams who gave very valuable feedback

1. The interfaces have been changed to follow a JSON-RPC structure to make it more agnostic and easier to use with multiple transports
2. Enforcing window.postMessage makes this CAIP not very future-proof as Chrome Extensions are slowly migrating to externally_connectable 
3. This CAIP is more useful if it does NOT describe the actual transport and instead focuses on the discovery interfaces
4. The usage of window.postMessage without mentions of targetOrigin creates huge security concerns

**What are the next steps?**

a. Separate window.postMessage from discovery interface into two CAIPs
b. Create a new CAIP which uses the discovery interface with externally_connectable
c. Create a new CAIP which uses the discovery interface with window.dispatchEvent

**What is the end goal?**

This CAIP intends to provide the most agnostic and interoperable interface for wallet discovery and consequentally handshake and signing.

Since CAIP-25 and CAIP-27 already describe handshake and signing then CAIP-282 will focus purely on discovery

**Why create separate CAIPs for each transport?**

These additional CAIPs are important for Apps and Libraries to support multiple wallet transports... most importantly the following:
- window.postMessage for embedded wallets with iframes
- window.dispatchEvent for legacy browser extension support
- externally_connectable for newer browser extension support

The combination of all JSON-RPC interfaces and different transport specs will any wallet to be discoverable, connect and sign with a decentralized application for all CAIP compatible chains

pedrouid:
This PR now includes 3 proposals:
- CAIP-282 -> Browser Wallet Discovery Interface
- CAIP-294-> Browser Wallet Messaging for Extensions (window.dispatchEvent)
- CAIP-295 -> Browser Wallet Messaging for Iframes (window.postMessage)

Still missing another standard to be included in this PR:
- CAIP-XXX -> Browser Wallet Messaging for Extensions (externally_connectable)

CAIPs/caip-294.md
```diff
@@ -0,0 +1,431 @@
...
+        params: walletData,
+      }
+    }
+  ));
+});
+```
+
+#### Handshake
+
+After the wallet has been selected by the user then the Blockchain Library MUST publish a message to share its intent to establish a connection. This can be either done as a [CAIP-25][caip-25] request.
```
mcintyre94, line 132: We've been questioning this a bit with wallet-standard. Generally a wallet will allow a user to authorise some set of accounts for a dapp. It'll then make those available as accounts on the wallet object (or here the scopes object mentioned in caip-282). From the dapp's perspective, the user selecting that wallet (or the dapp selecting it for them) decides which accounts/features/chains/etc are available to it. But for the wallet, it only needs to care about a connection when it needs to change those scopes. That doesn't necessarily have to happen every time the user selects the wallet in an app.

mcintyre94, line 132: Nit: I think the second part of the either is missing

CAIPs/caip-294.md
```diff
@@ -0,0 +1,431 @@
...
+
+window.addEventListener(`caip294:${selected_uuid}`, (event) => {
+  if (event.detail.method === "wallet_createSession") {
+    sessionRequest = event.detail
+    // if incoming requests match the WalletData UUID
+    if (checkSupportedScopes(event.detail.params)) {
+        // prompt user to approve session
+    }
+  }
+});
```
mcintyre94, line 349: Could a different wallet 'hijack' these events? Eg if I'm a wallet, can I send a `wallet_prompt` message in the app, capture the response, find the current UUID of Metamask, and then listen to its events to eg. also display my own popup?

CAIPs/caip-295.md
```diff
@@ -0,0 +1,388 @@
...
+discussions-to: https://github.com/ChainAgnostic/CAIPs/issues/295
+status: Draft
+type: Standard
+created: 2024-06-26
+requires: 25, 27, 282
+---
+
+## Simple Summary
+
+CAIP-295 defines a standardized messaging transport for browser iframe wallets.
```
kewde, line 14: Is there a demand for supporting iframes from the dApp side?

I'm not sure which wallets currently support iframes. I know there are some limitations on `react-native-webview` (typically used in mobile wallet dApp browser) that doesn't support injecting JS into the iframes on iOS.

CAIPs/caip-282.md
```diff
@@ -0,0 +1,183 @@
...
+
+Wallet Provider: A user agent that manages accounts and facilitates transactions with a blockchain.
+
+Decentralized Application (dapp): A web page that relies upon one or many Web3 platform APIs which are exposed to the web page via the Wallet.
+
+Blockchain Library: A library or piece of software that assists a dapp to interact with a blockchain and interface with the Wallet.
+
+#### Discovery
+
+Both Wallet Providers and blockchain libraries must listen to incoming messages that might be published after their initialization. Additionally both Wallet Providers and blockchain libraries must publish a message to both announce themselves and their intent to connect, respectively.
```
kewde, line 46:
```suggestion
Both Wallet Providers and blockchain libraries MUST listen to incoming messages that might be published after their initialization. Additionally both Wallet Providers and blockchain libraries MUST publish a message to both announce themselves and their intent to connect, respectively.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,183 @@
...
+interface WalletPromptRequestParams {
+  chains?: string[]; // compatible with CAIP-2
+  authName?: string; // compatible with CAIP-275
+}
+
+// for "wallet_announce" method
+interface WalletAnnounceRequestParams {
+  uuid: string;
+  name: string;
+  icon: string;
```
kewde, line 61: Can we specify the allowed formats? 

```
export type WalletIcon = `data:image/${'svg+xml' | 'webp' | 'png' | 'gif'};base64,${string}`;
```

Is what wallet standard recommends, I think the data URIs should be the mandatory format.

CAIPs/caip-294.md
```diff
@@ -0,0 +1,431 @@
...
+  `caip294:${selected_uuid}`,
+  {
+    detail: signingResponse
+  }
+));
+```
+
+## Security Considerations
+
+TODO
```
kewde, line 407: I think it's worth noting that a malicious wallet may be eavesdropping on the traffic of the other wallets and racing to reply, essentially hijacking/mimicking the wallet.

As a wallet you could attach the object to the window in a permanent way, `window.mywallet` which would become irreplaceable and undeleteable, providing a stronger security guarantee that the wallet you're talking with is the same one. It, ofcourse, doesn't guarantee authenticity of the wallet but it's at least consistent.

```typescript
Object.defineProperty(window, 'mywallet', {
  value: provider,
  writable: false,
  configurable: false;
});
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,183 @@
...
+#### Discovery
+
+Both Wallet Providers and blockchain libraries must listen to incoming messages that might be published after their initialization. Additionally both Wallet Providers and blockchain libraries must publish a message to both announce themselves and their intent to connect, respectively.
+
+This discovery would use the following JSON-RPC request params:
+
+```typescript
+// for "wallet_prompt" method
+interface WalletPromptRequestParams {
+  chains?: string[]; // compatible with CAIP-2
```
kewde, line 53: The behavior for `chains` is undefined,  should a wallet only reply if it has a chain in this list?

CAIPs/caip-282.md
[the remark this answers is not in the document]
pedrouid: The wallet WILL announce itself in two instances:
1. when it loads
2. when it receives a wallet_prompt message

This prevents any async problems that might arise from wallet and library to load at different stages

This pattern is inspired by the work done by EIP-6963:
https://eips.ethereum.org/EIPS/eip-6963#window-events

CAIPs/caip-282.md
```diff
@@ -0,0 +1,183 @@
...
+    accounts?: string[];
+    rpcDocuments?: string[];
+    rpcEndpoints?: string[];
+  };
+}
+```
+
+#### UUIDs
+
+The generation of UUIDs is crucial for this messaging interface to work seamlessly for the users.
```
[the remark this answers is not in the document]
pedrouid: UUID is local and ephemeral

RDNS is global and permanent 

The two identifiers together allow wallets to be identifiable but also de-duplicated if necessary

Either one by itself would not meet those requirements

CAIPs/caip-282.md
```diff
@@ -0,0 +1,183 @@
...
+
+### Prevent SVG Javascript Execution
+
+The use of SVG images introduces a cross-site scripting risk as they can include JavaScript code. This JavaScript executes within the context of the page and can modify the page or the contents of the page. So, when considering the experience of rendering the icons, dapps need to take into consideration how they’ll approach handling these concerns in order to prevent an image from being used as an obfuscation technique to hide malicious modifications to the page or to other wallets.
+
+## Privacy Considerations
+
+Any form of wallet discoverability must alwasys take in consideration wallet fingerprinting that can happen by malicious webpages or extensions that attempt to capture user information. Wallet Providers can abstain from publishing "Announce" messages on every page load and wait for incoming "Prompt" messages. Yet this opens the possibility for race conditions where Wallet Providers could be initialized after the "Prompt" message was published and therefore be undiscoverable. It is recommended that Wallet Providers offer this more "private connect" feature that users only enable optionally, rather than set by default.
```
pedrouid, line 155:
```suggestion
Any form of wallet discoverability must always take into consideration wallet fingerprinting that can happen by malicious webpages or extensions that attempt to capture user information. Wallet Providers can abstain from publishing "Announce" messages on every page load and wait for incoming "Prompt" messages. Yet this opens the possibility for race conditions where Wallet Providers could be initialized after the "Prompt" message was published and therefore be undiscoverable. It is recommended that Wallet Providers offer this more "private connect" feature that users only enable optionally, rather than set by default.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,183 @@
...
```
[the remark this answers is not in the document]
pedrouid: Fixed 👍

CAIPs/caip-282.md
pedrouid: Well we can't guarantee that behavior ALWAYS

If the wallet announces itself before the library is able to send wallet_prompt message then it can't evaluate the chains

But if the wallet does receive the wallet_prompt and chains field is specified then it would make sense to not announce itself

CAIPs/caip-282.md
```diff
@@ -0,0 +1,183 @@
...
+  rdns: string;
+  scopes?: AuthorizationScopes;
+}
+```
+
+Whenever a new Wallet Provider is discovered the Blockchain Library would index them in order for the Decentralized Application to display them and prompt the user for selecting their wallet of choice for this connection.
+
+The parameters `name` and `icon` are used to display to the user to be easily recognizable while the `rdns` and `uuid` are only used internally for de-duping while they must always be unique, the `rdns` will always be the same but `uuid` is ephemeral per browser session.
+
+The only optional parameter is `scps` which is defined by CAIP-217 authorization specs that enables early discoverability and filtering of wallets based on RPC methods, notifications, documents and endpoints but also optional discovery of supported chains and even accounts.
```
pedrouid, line 71:
```suggestion
The only optional parameter is `scopes` which is defined by CAIP-217 authorization specs that enables early discoverability and filtering of wallets based on RPC methods, notifications, documents and endpoints but also optional discovery of supported chains and even accounts.
```

CAIPs/caip-282.md
```diff
@@ -0,0 +1,183 @@
...
```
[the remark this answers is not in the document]
pedrouid: Fixed 👍

CAIPs/caip-282.md
kewde: Should we put that in the specification?

CAIPs/caip-295.md
kdenhartog: I know this is common for session based wallets and believe gnosis safe uses this approach too. I think this is going to become more common as well where iframes are going to contain the DApp logic on web2 pages kind of like what we're seeing blinks do, so it makes sense to make this stuff work with iframes when possible.

kdenhartog (commented):
few comments, I need to do a couple of passes on this to make sure I'm fully digesting it but just wanted to get these comments in there early.

CAIPs/caip-282.md
adonesky1: Total nit: would `networks` be a more generic and flexible term here?

CAIPs/caip-294.md
```diff
@@ -0,0 +1,431 @@
...
+      method: "wallet_prompt",
+      params: {
+        // optionally the Blockchain Library can prompt wallets to announce matching only the chains
+        chains: []  // optional
+        //  if the Blockchain Library supports CAIP-275 then it can include a name
+        authName: "", // optional
+      },
+    }
+  }
+));
```
adonesky1, line 89: Just a note [I have a related CAIP](https://github.com/ChainAgnostic/CAIPs/pull/308) that is specifically handling the case of discovery when the wallet's API is delivered via `externally_connectable`. Since in this case[ the dapp will need the `extensionId`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/connect#extensionid) in order to establish a channel of communication with the wallet, the `extensionId` is a critical piece of data for us to announce in this event. Perhaps it makes sense to merge this in as an optional param to this CAIP with some of the language from my CAIP to explain its use.

CAIPs/caip-282.md
pedrouid: I'm not sure `networks` would be helpful since everything is a network when we are talking about the internet

Here we are referring to Blockchain Networks defined by CAIP-2 identifiers

CAIPs/caip-282.md
```diff
@@ -0,0 +1,183 @@
...
+
+Whenever a new Wallet Provider is discovered the Blockchain Library would index them in order for the Decentralized Application to display them and prompt the user for selecting their wallet of choice for this connection.
+
+The parameters `name` and `icon` are used to display to the user to be easily recognizable while the `rdns` and `uuid` are only used internally for de-duping while they must always be unique, the `rdns` will always be the same but `uuid` is ephemeral per browser session.
+
+The only optional parameter is `scopes` which is defined by CAIP-217 authorization specs that enables early discoverability and filtering of wallets based on RPC methods, notifications, documents and endpoints but also optional discovery of supported chains and even accounts.
+
+```typescript
+// Defined by CAIP-217
+interface AuthorizationScopes {
```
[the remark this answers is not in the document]
pedrouid: Yes but at the same time... is this something that should be defined on the CAIP-282 for the RPC spec or should it dependent on each transport spec such as CAIP-294, CAIP-295 or CAIP-296

CAIPs/caip-282.md
```diff
@@ -0,0 +1,183 @@
...
+
+#### UUIDs
+
+The generation of UUIDs is crucial for this messaging interface to work seamlessly for the users.
+
+A Wallet Provider MUST always generate UUIDs distinctly for each web page loaded, and they must not be re-used without a session being established between the application and the wallet with the user's consent.
+
+A UUID can be re-used as a `sessionId` if and only if the [CAIP-25][caip-25] procedure has been prompted to the user and the user has approved its permissions to allow the application to make future signing requests.
+
+Once established, the UUID is used as `sessionId` for the [CAIP-27][caip-27] payloads, which can verify that incoming messages are being routed through the appropriate channels.
```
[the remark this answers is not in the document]
pedrouid: Actually there have been discussions on the RPC working group to make this sessionId optional since it can be fixed for each wallet 

Plus it's not necessarily required for all transports

CAIPs/caip-294.md
```diff
@@ -0,0 +1,431 @@
...
+const signingRequest = {
+  id: 456,
+  jsonrpc: "2.0",
+  method: "wallet_requestMethod",
+  params: {
+    // UUID from WalletData is used as SessionId
+    sessionId: walletData.uuid,
+    scope: "chain:777",
+    request: {
+      method: "chain_signMessage",
```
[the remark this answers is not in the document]
pedrouid: It depends on your definition of remote... because the idea of CAIP-282 is that it standarizes communication for all transports

That might be window.dispatchEvent, window.postMessage, externally_connectable, etc

Therefore there is always a concept of communication between a webpage and an external party that is remote to the webpage itself

I would not consider remote to be exclusively server-side or onchain

CAIPs/caip-294.md
pedrouid: Should this new proposal be labelled as CAIP-296 which uses externally_connectable as a transport for CAIP-282 discovery??

https://github.com/ChainAgnostic/CAIPs/issues/296

CAIPs/caip-282.md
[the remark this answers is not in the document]
kdenhartog: Hmm, I'm not familiar with the delineation between the two or how we're splitting them. I hadn't caught those details when reviewing this. Do we expect to ever have a transport method that doesn't rely on browser JS ever to communicate between a page? If not, then it seems like a common point shared between the various transport methods would be useful and reinforced in the specs as well.

CAIPs/caip-282.md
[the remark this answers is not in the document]
kdenhartog: When I wrote this I was thinking more along the context of `WalletAnnoucementRequestParams.UUID` property. I'm not following why a sessionId would be fixed either so I'm likely missing some context here.

CAIPs/caip-294.md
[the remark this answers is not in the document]
kdenhartog: That makes sense, I wasn't thinking of them as exclusively as server-side/onchain either. Instead, I was thinking about it more within the context of the different software boundaries that exist and whether we should be leaking them/collapsing them and when.

For example, a page typically interacts with a remote client acting on behalf of the user (say for permission to sign or to retrieve their data) and needs some transport interface to perform that interaction. In most cases, we expect that the page will communicate client via this transport, but because we're utilizing RPCs that don't really align the permissioning model with this (since chains are different/privacy models different) we end up in certain scenarios where the page may choose to bypass interacting with the client and retrieve information directly from the chain instead.

In my mind, this creates a more complex interaction model where it's not clear when the page-client interaction model is expected if the page can access the information itself via the chain which will probably produce odd side effects in a cross chain model.

For this reason, I think it be useful for us to not leak chain specific RPCs to the page and instead should be establishing an encapsulation pattern where the page uses the interface which encapsulates blockchain specific RPC logic away from the page rather than directly calling the RPC methods.

This encapsulation will help with creating a clear interface of reusable patterns to define application logic between the client and page across different blockchain ecosystems such that we can operate under the heuristic of "if the page wants the clients data it should go to the client to get it". Now, I can think of cases where this won't universally apply so I'm not sure it's the cleanest heuristic, but it does seem to be a good practice so that we can establish patterns of reuse between the client and page and make it easier for pages to interact.

CAIPs/caip-282.md
```diff
@@ -0,0 +1,183 @@
...
+#### Discovery
+
+Both Wallet Providers and blockchain libraries MUST listen to incoming messages that might be published after their initialization. Additionally both Wallet Providers and blockchain libraries MUST publish a message to both announce themselves and their intent to connect, respectively.
+
+This discovery would use the following JSON-RPC request params:
+
+```typescript
+// for "wallet_prompt" method
+interface WalletPromptRequestParams {
+  chains?: string[]; // compatible with CAIP-2
```
shanejonas, line 53: in caip-25 this is now called `scopes`

CAIPs/caip-294.md
bumblefudge: i remember @kdenhartog discussing at length the pros and cons (more like, limitations) of prototype freezing in the context of JS/ECMAScript/browser security over the years when EIP-6963 was being debated... it might still be worth mentioning that other cross-chain standards use a freezing approach, but i think it's a legacy thing anyways since a communication channel outside the page is what wallets are already migrating to which sidesteps the issue.

CAIPs/caip-282.md
bumblefudge:
```suggestion
  scopes?: string[]; // compatible with CAIP-2
```
good catch, i think it's `chains` in pedro's prod implementation 😅

bumblefudge:
apologies, pedro, my june comments were sitting in github UI limbo waiting to be batched to a complete review 🙄

CAIPs/caip-294.md
adonesky1: merged/addressed here: https://github.com/ChainAgnostic/CAIPs/pull/282/commits/bc4942857a8e04593ed92f7dc66653577a1c4435

CAIPs/caip-282.md
bumblefudge: see next comment

CAIPs/caip-294.md
```diff
@@ -0,0 +1,508 @@
...
+  // Required properties
+  uuid: string;
+  name: string;
+  icon: string;
+  rdns: string;
+
+  // Optional properties
+  extensionId?: string;
+  scopes?: Caip217AuthorizationScopes;
+}
```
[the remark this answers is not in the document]
bumblefudge: wait isn't this your PR you're change-requesting? haha, i'm just gonna merge it myself

CAIPs/caip-282.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Security Considerations
+
+Regarding security considerations it's important to consider that the WalletData can be imitated and/or manipulated by cross-site scripting and additionally we must consider the following:
+
+### Wallet Imitation and Manipulation
+
+Application developers are expected to actively detect for misbehavior of properties or functions being modified in order to tamper with or modify other wallets. One way this can be easily achieved is to look for when the uuid property within two WalletData objects match. Applications and Libraries are expected to consider other potential methods that the WalletData objects are being tampered with and consider additional mitigation techniques to prevent this as well in order to protect the user.
+
+### Prevent SVG Javascript Execution
```
chris13524, line 149:
```suggestion
### Prevent SVG JavaScript Execution
```

pedrouid (commented):
LGTM 👌

CAIPs/caip-295.md
jxom: Yep, iframes are certainly a valid use-case. Perhaps the Wallet wishes to be embedded in an iframe on the Application, instead of the user installing a browser extension.

CAIPs/caip-282.md
bumblefudge: chains --> [scopes]() --> [references](https://github.com/ChainAgnostic/CAIPs/pull/327)
```suggestion
  references?: string[]; // compatible with CAIP-2
```

CAIPs/caip-282.md
bumblefudge:
```suggestion
  icon: string; //base64'd dataURI of type image/{svg+xml|webp|png|gif}
```

bumblefudge: i'm not sure where to stype the TS declaration offered above but i would ALSO add the comment to this line

bumblefudge (changes requested):
Mostly Approved but there are still a few nits I couldn't fix myself and a few threads that it felt wrong to "resolve"... @pedrouid another 10min should get this over the line for me and on to ligi or al or obstropolous for final approval

bumblefudge:
> LGTM 👌

I think you meant this for the upstream PR onto this branch?

CAIPs/caip-294.md
```diff
@@ -0,0 +1,526 @@
...
+      jsonrpc: "2.0"
+      method: "wallet_announce",
+      params: walletData,
+    }
+  }
+));
+
+
+
+// Wallet Providers starts listenning on init
```
adonesky1:
```suggestion
// Wallet Providers starts listening on init
```

obstropolos:
> Mostly Approved but there are still a few nits I couldn't fix myself and a few threads that it felt wrong to "resolve"... @pedrouid another 10min should get this over the line for me and on to ligi or al or obstropolous for final approval

@pedrouid bump

clx07 (approved):

Issue #290, opened 2024-06-20
[CAIP-285] - chain-specific resolution corner-case


bumblefudge:
A helpful reviewer left a pretty significant comment (that probably should've been an issue) on the PR merging CAIP-275, which maybe no one saw because the PR was merged before the comment:
https://github.com/ChainAgnostic/CAIPs/pull/275#issuecomment-2167321691

Is this a legitimate concern? Would it make more sense to consider an "implicit chainId" (1, or 0) be injected any time no chainId is present?  I think the semantics of chainId 0 are starting to "be a thing", between our recent [namespace PR](https://github.com/ChainAgnostic/namespaces/pull/107/files) making it a thing in CAIP-2 systems, and [EIP-7702](https://eips.ethereum.org/EIPS/eip-7702#behavior) specifying 0 as the `chainId` for offchain cases... although maybe I misunderstand OP and they are actually suggesting chainId `1` be the default when no other id is set?

FedericoAmura:
Hi Bumblefudge!
It is a valid concern, which we have been rethinking this last days as we advanced with development and communication with partners
We are making an update on the CAIP with some changes to make this more clear, including:
- removing the name to address resolution call. This call was the biggest culprit of confusion and is going to be replaced with the `chain` and `address` field in the `authenticator`. This also makes the necessary information totally self contained and easier to use, improving both DX and UX
- explaining how the `chain` field must be treated. In this case, it would be as you say, understanding no value as 0, offline or EOA. And whenever that value is defined (CAIP-2 format) it can be said it is a smart wallet on that specific chain

We will open a PR shortly

PR #292, opened 2024-06-25
CAIP-275: remove extra request and expand on chain property usage

FedericoAmura:
Following comments on  #290 (and [this comment](https://github.com/ChainAgnostic/CAIPs/pull/275#issuecomment-2167321691)) we expand some more on how to treat the `chain` property in authentication config. Also we remove the need for an extra request that can cause issues or degrade UX when using different wallets for authentication and other things

### DIFF
--- a/CAIPs/caip-275.md
+++ b/CAIPs/caip-275.md
@@ -83,7 +83,7 @@ Application will filter the authentication flows that are not supported and then
 An authentication flow may be unsupported due to the platform it is running mismatching the one it requires, e.g. "browser", "mobile", etc. Or because it requires an specific wallet but it is not installed, e.g. requires "io.metamask" but there is no browser extension wallet.
 When an authentication flow requires user action, such as scanning a QR code, or authorizing the connection from the wallet, the application must wait for the user to complete or cancel the action.
 
-### Authentication flows definition
+## Authentication flows definition
 
 The Authentication flows definition JSON MUST conform to the following [Draft 7 JSON Schema][]:
 
@@ -173,11 +173,25 @@ To resolve that configuration:
 3. Second authentication flow indicates that it has to run on `mobile`, and using [Mobile Wallet Protocol][] (indicated with the key `mwp`) application has to find another app that responds to the universal link `https://universal-link.wallet.com/`.
 4. Last authentication flow does not impose a platform to run on because it does not specify any, so it will be valid everywhere. It tries to set up a [Wallet Connect][] connection to the URI `http://www.wallet.com/auth`. When this authentication flow is triggered, the URI will be opened in a new browser window passing `address`, `domain` and `wcUri` as query params with the resolved address, the domain name and [Wallet Connect][] URI respectively. As this flow requires user action (connecting the wallet at the URI) this step will never be skipped as long as the application supports [Wallet Connect][].
 
-### Other examples
+## Standard Authentication flows
 
 To further clarify different authentication flow configurations we can consider the following cases.
 
-#### Use the wallet extension installed at the browser, without requiring a specific one
+### Use a wallet extension installed at the browser
+
+```json
+{
+  "platform": "browser",
+  "connection": "extension",
+  "URI": "io.metamask"
+}
+```
+
+When using this configuration, the application MUST check available [EIP-6963][] wallets via its signaling channel and trigger the one that matches the URI field with its rdns.
+In these cases, the application should not ignore and replace the URI field with any other wallet found in the browser, as it could lead to a security issue where the user connects a wallet that is not the one they want to authenticate with.
+Therefore, this configuration will only be valid if the browser has that wallet installed (in this case, Metamask) and be skipped when it is not.
+
+### Use the wallet extension installed at the browser, without requiring a specific one
 
 ```json
 {
@@ -189,11 +203,12 @@ To further clarify different authentication flow configurations we can consider 
 
 As this authentication flow calls for the `injected` extension, without specifying which one, it should use the default wallet provider at the browser.
 For EVM chains, this would be `window.ethereum`.
+Having `connection` set to `extension`, the `URI` field is by default `injected`, so some users might want to skip this field and save some bytes.
 
-However, this configuration is not recommended as having multiple wallets in the browser generates a race condition, and can result in false positives where there is a wallet extension, but it does not have the account linked to the domain.
-We recommend using [EIP-6963][] by specifying a wallet rnds in the URI field if possible, to avoid this issue.
+However, this configuration is not recommended as having multiple wallets in the browser generates a race condition, and can result in false positives where the wallet used does not have the account linked to the domain.
+We recommend using [EIP-6963][] by specifying a wallet rnds in the `URI` field if possible, to avoid this issue.
 
-#### Wallet connect to any wallet, letting the user decide
+### Wallet connect to any wallet, letting the user decide upon connection
 
 ```json
 {
@@ -202,14 +217,23 @@ We recommend using [EIP-6963][] by specifying a wallet rnds in the URI field if 
 ```
 
 Using this configuration, the application should provide the QR code and [Wallet Connect][] URI for the user to connect with his wallet from any place.
-Such configuration can be useful when the user has their wallet in their mobile device and also on another platform such as a web wallet or desktop application that can receive the QR code or [Wallet Connect][] URI in any way, and they want to decide which one to use at connection time.
+Such configuration can be useful when the user has their wallet in their mobile device or on another platform, such as a web wallet or desktop application that can receive the [EIP-1328] URI and present it to the end-user as a "[Wallet Connect]"-style QR code or otherwise get informed consent from the user to make the relay connection; this has the added benefit of allowing a manual choice among multiple wallets at connection time.
 
 This configuration should never be skipped as it does not impose any platform and requires the user action of connecting from the wallet making this a great last case in the `authFlows` array and behaving as the default when no other authentication flow is useful.
-The only requirement for this configuration to work is that the application supports [Wallet Connect][].
+The only requirement for this configuration to work is that the application supports [ERC-1328]-conformant relay connections such as [Wallet Connect][].
 
 ## Self-hosting authentication flows
 
 In case the user does not trust any authentication flow provider to store their info, they can easily host it themselves using a public serverless function and saving its URL in the ENS `authenticator` record.
+The requirements for the serverless functions are:
+- Being accessible from the internet.
+- Being able to return the authentication flows JSON object when requested.
+- Provide the user with a way to update the authentication flows in case they want to change them.
+- Be able to handle the user's domain name as a parameter to return the correct authentication flows when that changes it.
+
+The user can use any service that provides serverless functions, such as Vercel, AWS Lambda, Google Cloud Functions, etc.
+When this is done for personal reasons, the user should be aware that the serverless function will be public and anyone can access it.
+Therefore, it is recommended to use a service that provides, at least, rate limiting to avoid abuse.
 
 ## Direct authentication flows resolution
 
@@ -239,32 +263,34 @@ Web3 applications and login modal providers can implement the Login With Name fl
 ## Connection properties
 
 How the dApp actually requests signatures and talks to the signer will vary depending on the authenticator and the chain being used.
-One option for session management is [Wallet Connect][], where the "authenticated session" returned is actually a WalletConnect session.
+One option for session management is to create a session with the [CAIP-25][] protocol, where the "authenticated session" token returned is actually a JSON-RPC session token, such as that used by the [Wallet Connect] JSON-RPC relay network.
 Applications may also choose to directly integrate signer SDKs, providing a more streamlined signing flow.
 
 For example, here is how an application would integrate with the Login With Name login process:
 
-1. Install [Wagmi][] and [Login With Name WAGMI connector][] packages.
-2. Import the `loginWithName` connector from the package.
-3. Configure the [Wagmi][] config with the `loginWithName` connector.
-   This package provides ENS as a name resolver which can be configured to use `mainnet` or `sepolia`.
-4. Wrap the app with `WagmiProvider`, pass it the configuration with the loginWithName connector.
-5. Upon connection request, the connector will request application for user name. Application can show user an input for it, if not obtained in other way.
+1. Install connection/authentication libraries, such as [Wagmi][] and [Login With Name WAGMI connector][] packages.
+2. Import the connector method from the package (`loginWithName` in the case of `wagmi` and our sample implementation).
+3. Configure the macro library (in this case [Wagmi][]) to use the specific connector (`loginWithName`).
+   Our sample implementation package provides `ENS` as a name resolver which can be configured to use `mainnet` or `sepolia`.
+4. Configure the app to use the previously configured library to establish the connection.
+   Following the example would be wrapping the app with `WagmiProvider`, and passing it the configuration with the `loginWithName` connector.
+5. Upon connection request, the connector will request application for user name.
+   Application can try cached or automatically resolved suggestions first and offer a user confirmation of the suggestion(s), or if none can be resolved without user intervention, prompt the user for a user name directly.
 6. Connector will now resolve user name into an address and authentication flows and trying to reach the wallet specified and its accounts.
-7. After signing in with the wallet. Application can use standard [Wagmi][] hooks like `useAccount`, `useConnect` etc.
+7. After signing in with the wallet. Application can use the hooks and abstractions provided by the library, such as `useAccount`, `useConnect` etc in the case of `wagmi`.
 
 ## Name Resolver systems
 
 For the purposes of this document, we've detailed a flow based on ENS domains.
 But this standard is extensible to any name resolution system.
 For example, to use Solana domains, you could just replace the ENS name resolver component with a Solana name resolver component.
-The name resolver can also mix several sources, so resolving using several chains is only a matter of combining the specific name resolver for each chain and then combining them with the necessary logic.
+The name resolver can also mix several sources, so resolving from several chains (even non-EVM chains) is only a matter of combining the specific name resolver for each chain and then combining them with the necessary logic.
 
-Reference implementation included in [Login With Name WAGMI connector][] include two name resolvers to showcase different implementations.
+Sample implementation included in [Login With Name WAGMI connector][] include two name resolvers to showcase different implementations.
 
-- ENS, compatible with any chain that uses it and likely the one that will be the most common.
+- ENS, which has wide compatibility with many chains and integrated in many user-facing interfaces across web3.
 - Domain Wallets, an online web wallet system compatible with many chains and that already integrates a domain naming system to identify each wallet.
-- An application level mixer name resolver, that combines the two previous ones.
+- An application-level multi-resolver, that handles the logic of the two specific resolvers above.
 
 It has to be noted too, that the name resolver can be any service that resolves names and their associated `authenticator` text record to an address and authentication flows respectively.
 A name resolver can take any type of name and as long as it can resolve it, it does not have to be tied to any chain.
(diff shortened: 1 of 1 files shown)


CAIPs/caip-275.md
```diff
@@ -186,11 +187,13 @@ To further clarify different authentication flow configurations we can consider
 }
 ```
 
-As this authentication flow calls for the `injected` extension, without specifying which one, it should use EIP-6963 to check if a matching extension provider is available, and use that if it is available. If one is not found, we resolve to the default one.
-For ethereum addresses, this would be `window.ethereum`.
-However, this configuration is not recommended as having multiple wallets generates a race condition and can provide false positives where there is a wallet extension but it does not have the account linked to the domain. We recommend EIP-6963 if possible.
+As this authentication flow calls for the `injected` extension, without specifying which one, it should use the default wallet provider at the browser.
+For EVM chains, this would be `window.ethereum`.
```
bumblefudge, line 191: I'm still a little confused why this flow skips over EIP 6963 signaling (which could check for `com.wallet.rdns` by name if multiple extensions are installed) and defaulting to legacy EIP 1193 (which has that famous race condition to inject the provider, and could open a different wallet than intended if another wallet overwrote window.ethereum).

bumblefudge: it's not really a blocker here, i can just open a separate issue

CAIPs/caip-275.md
```diff
@@ -186,11 +187,13 @@ To further clarify different authentication flow configurations we can consider
...
 
-#### Wallet connect to any wallet
+However, this configuration is not recommended as having multiple wallets in the browser generates a race condition, and can result in false positives where there is a wallet extension, but it does not have the account linked to the domain.
+We recommend using [EIP-6963][] by specifying a wallet rnds in the URI field if possible, to avoid this issue.
```
bumblefudge, line 194: Maybe it would be smoother to write this section as a 

SHOULD check rdns against 6963 signaling channel first to avoid false-positives or wallet impersonation via protocol pollution

followed by

MUST check window.ethereum if 6963 signal not detected.

This way your own SDK could do it right with 1193 as fallback and lazy implementations or implementations biased towards certain wallets/formfactors would have the option to skimp, but a stronger warning not to?

CAIPs/caip-275.md
```diff
@@ -199,91 +202,82 @@ However, this configuration is not recommended as having multiple wallets genera
...
+Such configuration can be useful when the user has their wallet in their mobile device and also on another platform such as a web wallet or desktop application that can receive the QR code or [Wallet Connect][] URI in any way, and they want to decide which one to use at connection time.
 
-This configuration will never be skipped as it does not impose any platform and requires the user action of connecting from the wallet making this a great last case in the `authFlows` array and behaving as the default when no other authentication flow is useful.
+This configuration should never be skipped as it does not impose any platform and requires the user action of connecting from the wallet making this a great last case in the `authFlows` array and behaving as the default when no other authentication flow is useful.
+The only requirement for this configuration to work is that the application supports [Wallet Connect][].
```
bumblefudge, line 208:
```suggestion
The only requirement for this configuration to work is that the application supports [ERC-1328](https://eips.ethereum.org/EIPS/eip-1328)-conformant relay connections such as [Wallet Connect][].
```
Probably worth generalizing here, since there are actually alternate providers/copycats that use the same QR format but a different network.

CAIPs/caip-275.md
```diff
@@ -199,91 +202,82 @@ However, this configuration is not recommended as having multiple wallets genera
 ```
 
 Using this configuration, the application should provide the QR code and [Wallet Connect][] URI for the user to connect with his wallet from any place.
-Such configuration can be useful when the user has their wallet in their mobile device and also on another platform such as a web wallet or desktop application that can receive the QR code or [Wallet Connect][] URI in any way and they want to decide which one to use at connection time.
+Such configuration can be useful when the user has their wallet in their mobile device and also on another platform such as a web wallet or desktop application that can receive the QR code or [Wallet Connect][] URI in any way, and they want to decide which one to use at connection time.
```
bumblefudge, line 205:
```suggestion
Such configuration can be useful when the user has their wallet in their mobile device and also on another platform such as a web wallet or desktop application that can receive the [EIP-1328] URI and present it to the end-user as a "[Wallet Connect]"-style QR code or otherwise get informed consent from the user to make the relay connection; this has the added benefit of allowing a manual choice among multiple wallets at connection time.
```

CAIPs/caip-275.md
```diff
@@ -374,6 +361,7 @@ Also and a working demo with further details on how this CAIP works is hosted at
 - [Draft 7 JSON Schema][] - Draft 7 JSON Schema
 - [Login With Name WAGMI connector][] - Login With Name WAGMI connector
 - [Mobile Wallet Protocol][] - Mobile Wallet specification
+- [Wagmi][] - Wagmi SDK and docs
 - [Wallet Connect][] - Wallet Connect SDK and docs
 
 [CAIP-2]: https://chainagnostic.org/CAIPs/caip-2
```
bumblefudge, line 367:
```suggestion
[CAIP-2]: https://chainagnostic.org/CAIPs/caip-2
[EIP-1328]: https://eips.ethereum.org/EIPS/eip-1328
```

CAIPs/caip-275.md
```diff
@@ -199,91 +202,82 @@ However, this configuration is not recommended as having multiple wallets genera
...
 
-This configuration will never be skipped as it does not impose any platform and requires the user action of connecting from the wallet making this a great last case in the `authFlows` array and behaving as the default when no other authentication flow is useful.
+This configuration should never be skipped as it does not impose any platform and requires the user action of connecting from the wallet making this a great last case in the `authFlows` array and behaving as the default when no other authentication flow is useful.
+The only requirement for this configuration to work is that the application supports [Wallet Connect][].
 
-## Self hosting authentication flows
+## Self-hosting authentication flows
 
 In case the user does not trust any authentication flow provider to store their info, they can easily host it themselves using a public serverless function and saving its URL in the ENS `authenticator` record.
```
bumblefudge, line 212: serverless like service worker URI? or a static gitpages website?  I would probably provide an appendix (for each, if both work) because this is a little hard to picture and people deciding whether or not to implement this would probably need to know how foolproof an automation of this self-hosting option is before commiting 😄

CAIPs/caip-275.md
```diff
@@ -199,91 +202,82 @@ However, this configuration is not recommended as having multiple wallets genera
...
 
 ## Direct authentication flows resolution
```
bumblefudge, line 214: I'm a little confused on the organization here-- maybe we could move usecases to one `##`-level section and map each `###`-level usecase inside of it explicitly to a `###`-level flow specification inside of a `##`-level Flows section? Not urgent but as this doc gets longer making it more readable/skimmable will help

CAIPs/caip-275.md
```diff
@@ -199,91 +202,82 @@ However, this configuration is not recommended as having multiple wallets genera
...
-- Application will resolve only its issued names, therefore the authentication flows or their location are likely already known. In such cases the application will likely use its own backend to resolve `authenticator` instead of ENS or another Crypto Domain system NFT system. Responding with the stringified version of the saves the user of the extra request as it does not provide any useful information.
+- The user wants a truly decentralized solution and does not care about the gas cost of storing lots of data in blockchain; then they can write a stringified version of their authentication flows definition in the `authenticator` text record.
+- Application will resolve only its issued names, therefore the authentication flows or their location are likely already known. In such cases the application can use its own backend to resolve `authenticator` instead of ENS or another Crypto Domain system NFT system. Responding with the stringified version of the saves the user of the extra request as it does not provide any useful information.
 
 # Login With Name Flow
```
bumblefudge, line 221: similarly, i think it might be clearer to lead with the macro-level flow first and then detail individual flows second and use-cases third?

CAIPs/caip-275.md
```diff
@@ -199,91 +202,82 @@ However, this configuration is not recommended as having multiple wallets genera
...
+5. If no authentication flow can be processed by the application, then display this situation accordingly to the user.
+6. Wallet, when needed, will trigger the user for authentication and connection approval before establishing the connection with the wallet.
+7. After wallet approves connection, application can match the domain name, the user's wallet address, and an authenticated session to the ones requested previously to validate user is not connecting with another wallet.
+8. When everything succeeds, connection should be established between application and the wallet found and triggered under the name provided by the user.
 
 ## Connection properties
 
 How the dApp actually requests signatures and talks to the signer will vary depending on the authenticator and the chain being used.
-One option for session management is WalletConnect, where the "authenticated session" returned is actually a WalletConnect session.
+One option for session management is [Wallet Connect][], where the "authenticated session" returned is actually a WalletConnect session.
```
bumblefudge, line 242:
```suggestion
One option for session management is to create a session with the [CAIP-25][] protocol, where the "authenticated session" token returned is actually a JSON-RPC session token, such as that used by the [Wallet Connect] JSON-RPC relay network.
```

There's other CAIP-25 implementations in the works so important to generalize the terminology here!

CAIPs/caip-275.md
```diff
@@ -199,91 +202,82 @@ However, this configuration is not recommended as having multiple wallets genera
...
 
 For example, here is how an application would integrate with the Login With Name login process:
 
-1. Install wagmi and [Login With Name WAGMI connector][] packages
-2. Import the loginWithName connector from the package
-3. Configure the wagmi config with the loginWithName connector.
-   This package provides ENS as a name resolver which can be configured to use mainnet or sepolia
-4. Wrap the app with WagmiProvider, pass it the configuration with the loginWithName connector
-5. Upon connection request, the connector will request application for user name. Application can show user an input for it if not obtained in other way.
+1. Install [Wagmi][] and [Login With Name WAGMI connector][] packages.
```
bumblefudge, line 247:
```suggestion
1. Install connection/authentication libraries, such as the [Wagmi][] and [Login With Name WAGMI connector][] packages.
```
Even tho it's a non-normative example, still good to generalize so that the spec still makes sense years from now, if, say, wagmi gets renamed, splits in two, etc etc.

CAIPs/caip-275.md
```diff
@@ -199,91 +202,82 @@ However, this configuration is not recommended as having multiple wallets genera
...
+2. Import the `loginWithName` connector from the package.
```
bumblefudge, line 248:
```suggestion
2. Import the connector method from the package (`loginWithName` in the case of `wagmi`).
```

CAIPs/caip-275.md
```diff
@@ -199,91 +202,82 @@ However, this configuration is not recommended as having multiple wallets genera
...
+3. Configure the [Wagmi][] config with the `loginWithName` connector.
```
bumblefudge, line 249:
```suggestion
3. Configure the macro library (in this case [Wagmi][]) to use the specific connector (`loginWithName`).
```

CAIPs/caip-275.md
```diff
@@ -199,91 +202,82 @@ However, this configuration is not recommended as having multiple wallets genera
...
+   This package provides ENS as a name resolver which can be configured to use `mainnet` or `sepolia`.
+4. Wrap the app with `WagmiProvider`, pass it the configuration with the loginWithName connector.
```
bumblefudge, line 251: lol this one is hard for me to generalize off the top of my head but you get the idea

CAIPs/caip-275.md
```diff
@@ -199,91 +202,82 @@ However, this configuration is not recommended as having multiple wallets genera
...
+5. Upon connection request, the connector will request application for user name. Application can show user an input for it, if not obtained in other way.
```
bumblefudge, line 252:
```suggestion
5. Upon connection request, the connector will request application for user name. Application can try cached or automatically resolved suggestions first and offer a user confirmation of the suggestion(s), or if none can be resolved without user interventioned, prompt the user for a user name directly.
```
better to be super explicit and phrase as an algorithm

CAIPs/caip-275.md
```diff
@@ -199,91 +202,82 @@ However, this configuration is not recommended as having multiple wallets genera
...
 6. Connector will now resolve user name into an address and authentication flows and trying to reach the wallet specified and its accounts.
-7. After signing in with the wallet. Application can use standard wagmi hooks like useAccount, useConnect etc.
+7. After signing in with the wallet. Application can use standard [Wagmi][] hooks like `useAccount`, `useConnect` etc.
```
bumblefudge, line 254:
```suggestion
7. After signing in with the wallet. Application can use the hooks and abstractions provided by the library, such as `useAccount`, `useConnect` etc in the case of `wagmi`.
```

CAIPs/caip-275.md
```diff
@@ -199,91 +202,82 @@ However, this configuration is not recommended as having multiple wallets genera
...
 
 ## Name Resolver systems
 
 For the purposes of this document, we've detailed a flow based on ENS domains.
 But this standard is extensible to any name resolution system.
 For example, to use Solana domains, you could just replace the ENS name resolver component with a Solana name resolver component.
-The name resolver can also mix several sources, so resolving using several chains is only a matter of combining the specific name resolver for each chain and then combining them with the necessary logic
+The name resolver can also mix several sources, so resolving using several chains is only a matter of combining the specific name resolver for each chain and then combining them with the necessary logic.
```
bumblefudge, line 261:
```suggestion
The name resolver can also mix several sources, so resolving from several chains (even non-EVM chains) is only a matter of combining the specific name resolver for each chain and then combining them with the necessary logic.
```

CAIPs/caip-275.md
```diff
@@ -199,91 +202,82 @@ However, this configuration is not recommended as having multiple wallets genera
...
 
 Reference implementation included in [Login With Name WAGMI connector][] include two name resolvers to showcase different implementations.
 
-- ENS, compatible with any chain that uses it and likely the one that will be the most common
-- Domain Wallets, an online web wallet system compatible with many chains and that already integrates a domain naming system to identify each wallet
-- An application level mixer name resolver, that combines the two previous ones
+- ENS, compatible with any chain that uses it and likely the one that will be the most common.
```
bumblefudge, line 265:
```suggestion
- ENS, which has wide compatibility with many chains and integrated in many user-facing interfaces across web3.
```
maybe better not to explicit call out market-share, it makes competitors nervous

CAIPs/caip-275.md
```diff
@@ -199,91 +202,82 @@ However, this configuration is not recommended as having multiple wallets genera
...
+- Domain Wallets, an online web wallet system compatible with many chains and that already integrates a domain naming system to identify each wallet.
+- An application level mixer name resolver, that combines the two previous ones.
```
bumblefudge, line 267:
```suggestion
- An application-level multi-resolver, that handles the logic of the two specific resolvers above.
```

CAIPs/caip-275.md
```diff
@@ -350,9 +337,9 @@ export class ENS implements NameResolver {
 
 # Rationale
```
bumblefudge, line 338: it might be good to add a kind of high-level rationale, like "this doesn't compete with other authentication method, it is just an opt-in discovery system that name-controllers can use and that applications can trust to automate one friction point."
it might also be the right section to explain how future authentication methods (#teamWearables) get added to the mix-- followup CAIPs? are you providing a template for those? 😄

CAIPs/caip-275.md
```diff
@@ -361,8 +348,8 @@ Existing NFTs and applications will continue to function normally.
 
 ## Reference Implementation
```
bumblefudge, line 349: Not to be a stickler, but the two main differences between a sample implementation and a reference implementation is that 
1.) ongoing maintenance - it has to still work for a long time, and you have to remove or rename this section if you stop maintaining it later
2.) all optional extensions supported -- i.e., you have to update (or at least accept PRs) if #teamWearables writes a CAIP a year from now adding logic for a wearables flow.

one dodge if you don't want to commit to that before going final with this CAIP is calling it a "sample implementation" 😅

FedericoAmura:
Hi @bumblefudge !
Were you able to review the updated version? Do you have any more feedback?

CAIPs/caip-275.md
bumblefudge: 5 weeks passed and now I don't get my point either

bumblefudge:
Sorry, I missed that last commit somehow. Reviewing now

bumblefudge (approved):
LGTM

GregTheGreek (approved):

CAIPs/caip-275.md
```diff
@@ -94,11 +94,11 @@ The Authentication flows definition JSON MUST conform to the following [Draft 7
   "properties": {
     "address": {
       "type": "string",
-      "description": "The address requested by the user. This can be from any chain."
+      "description": "The wallet address requested by the user. This can be from any chain. And can be an externally owned account or a smart wallet"
```
obstropolos, line 97:
```suggestion
      "description": "The wallet address requested by the user. This can be from any chain and can be an externally owned account or smart wallet"
```

CAIPs/caip-275.md
```diff
@@ -94,11 +94,11 @@ The Authentication flows definition JSON MUST conform to the following [Draft 7
...
     },
     "chain": {
       "type": "string",
-      "description": "The blockchain id as specified in CAIP-2. This is optional and can be used to filter the authentication flows by chain. If not provided, it should support all chains that support the address."
+      "description": "The blockchain id as specified in CAIP-2. This is only required when the wallet is a smart account in a specific chain. If not provided, it is assumed the wallet is an EOA that is not on any chain."
```
obstropolos, line 101:
```suggestion
      "description": "The blockchain id as specified in CAIP-2. This is only required when the wallet is a smart account on a specific chain. If not provided, it is assumed the wallet is an EOA that is not on any chain."
```

CAIPs/caip-275.md
```diff
@@ -138,7 +138,9 @@ Possible values for each field are:
   - `wc`, to trigger a connection with the wallet using [Wallet Connect][]
   - `mwp`, to discover and communicate with the wallet usign [Mobile Wallet Protocol][]
 
-Those values are not exhaustive. Meaning they can be extended as new options arise or become popular. For example, a new platform such as `wearable` or a new communication protocol could be defined
+Those values are not exhaustive. Meaning they can be extended as new options arise or become popular.
```
obstropolos, line 141:
```suggestion
Those values are not exhaustive, which means they can be extended as new options arise or become popular.
```

CAIPs/caip-275.md
```diff
@@ -166,17 +168,30 @@ To further clarify how authentication flows should be used, let's consider the f
...
+3. Second authentication flow indicates that it has to run on `mobile`, and using [Mobile Wallet Protocol][] (indicated with the key `mwp`) application has to find another app that responds to the universal link `https://universal-link.wallet.com/`.
+4. Last authentication flow does not impose a platform to run on because it does not specify any, so it will be valid everywhere. It tries to set up a [Wallet Connect][] connection to the URI `http://www.wallet.com/auth`. When this authentication flow is triggered, the URI will be opened in a new browser window passing `address`, `domain` and `wcUri` as query params with the resolved address, the domain name and [Wallet Connect][] URI respectively. As this flow requires user action (connecting the wallet at the URI) this step will never be skipped as long as the application supports [Wallet Connect][].
 
-### Other examples
+## Standard Authentication flows
```
obstropolos, line 176:
```suggestion
## Standard authentication flows
```

obstropolos: Just for consistency - if you want to fully title-capitalize, do that instead - but one or the other.

CAIPs/caip-275.md
```diff
@@ -350,19 +363,25 @@ export class ENS implements NameResolver {
 
 # Rationale
 
-Specifying the authenticator URL as a domain name NFT text record allows applications to easily discover and integrate with compatible login methods in a standard way
-Having a chain-agnostic standard enables interoperability between different crypto domain providers and authentication methods
-Providing clear wallet implementer steps and code samples makes it easy for developers to adopt this standard
+The Domain Wallet Authentication standard is designed to provide a user-friendly way to authenticate with web3 applications using easy to remember names, reducing friction for end-users, specially those that are not familiar with the complexities of blockchain addresses.
```
obstropolos, line 366:
```suggestion
The Domain Wallet Authentication standard is designed to provide a user-friendly way to authenticate with web3 applications using easy-to-remember names, reducing friction for end-users - especially those that are not familiar with the complexities of blockchain addresses.
```

CAIPs/caip-275.md
```diff
@@ -350,19 +363,25 @@ export class ENS implements NameResolver {
...
+By linking a domain name with authentication methods or providers, users can easily log in to web3 applications without having to remember their wallet provider or address.
+Specifying the `authenticator` configuration as a domain name NFT text record allows applications to easily discover and integrate with compatible login methods in a standard way.
+Having a chain-agnostic standard enables interoperability between different crypto domain providers and authentication methods.
+Providing clear wallet implementer steps and code samples makes it easy for developers to adopt this standard.
+
+The standard is designed to not compete with other authentication methods and as an opt-in discovery system that name-controllers can use and applications can trust to automate one friction point in the web3 login process.
+It is designed to be flexible, offering options to both consumers at connection time and developers at configuration time.
+And also to be extensible, allowing for new platforms, connections, and name resolvers to be added as they become available and publicly discussed and adopted by the community in followup CAIPs.
```
obstropolos, line 374:
```suggestion
Additionally, it is designed to be extensible, allowing for new platforms, connections, and name resolvers to be added as they become available and publicly discussed and adopted by the community in followup CAIPs.
```

obstropolos (commented):
Might be some additional notes here but for now just some slight errata

FedericoAmura:
> Might be some additional notes here but for now just some slight errata

Applied suggested changes. Thanks!!!

obstropolos (approved):

octavio12345300 (approved):

PR #300, opened 2024-06-28
CAIP-300: Wallet Connect JSON-RPC Method

lukaisailovic:
This CAIP defines a JSON-RPC method to request a batch of RPC methods to be resolved when connecting a wallet in a single roundtrip.

### DIFF
--- a/caip-300.md
+++ b/caip-300.md
@@ -1,0 +1,223 @@
+---
+caip: 300
+title: Wallet Connect JSON-RPC Method
+author: Lukas Rosario (@lukasrosario), Conner Swenberg (@ilikesymmetry), Pedro Gomes (@pedrouid), Luka Isailovic (@lukaisailovic)
+discussions-to: https://github.com/ChainAgnostic/CAIPs/discussions/300
+status: Draft
+type: Standard
+created: 2023-06-28
+requires: 2, 10,
+---
+
+## Simple Summary
+
+This CAIP defines a JSON-RPC method to request a batch of RPC methods to be resolved when connecting a wallet in a single roundtrip.
+
+## Abstract
+
+This proposal enables a one-click experience to fulfill several RPC requests in parallel during the connection approval when it's established when the user is requested to connect its wallet. It gives the ability for a wallet to not only establish a session, but also authenticate the user, expose capabilities or features, define some onchain permissions, etc.
+
+## Motivation
+
+Currently connecting a wallet establishes a session which will be used for resolving future RPC requests which are scoped in a session with [CAIP-25][caip-25] scopes which results in an experience where user is redirected multiple times to the wallet for approving different signatures.
+
+However there are several scenarios where you are required to resolve RPC requests that require user approval but are contextually required to happen when you first connect your wallet. This essentially ends up creating a friction point for users on their first few seconds when they connect their wallet and are redirected to the wallet two or three times in a row.
+
+One of the most common scenarios is the requirement to authenticate a user right after the session is established which means the user is redirected first to approve a [CAIP-25][caip-25] session and immediatelly is redirected to sign an authentication request with [CAIP-122][caip-122] message to be signed. This specific use-case is addressed by the introduction of `wallet_authenticate` as standardized by [CAIP-222][caip-222].
+
+Yet there are many other requests that can benefit from this pattern but it's not feasible to create several replicas of specialized batches of requests on a single round-trip similarly to [CAIP-222][caip-222]. Also these batches are sometimes not chain-agnostic and require ecosystem specific requests which are specified for blockchains like Ethereum, Solana, Cosmos, Polkadot, etc.
+
+Therefore this proposal creates a generalized approach to batching requests to be resolved in a single roundtrip which would be fullfilled by redirecting the user only once when they connect their wallet and more importantly, it will support RPCs that are useful for ecosystem specifically.
+
+## Specification
+
+This JSON-RPC method can be requested to a wallet provider without prior knowledge of any blockchain accounts, chains, methods or other features.
+
+### Request
+
+The application would interface with a wallet to make request as follows:
+
+```jsonc
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "method": "wallet_connect",
+  "params": {
+      "requests?": {
+        "method": string,
+        "params": unknown
+      }[],
+  }
+}
+```
+
+The JSON-RPC method is labelled as `wallet_connect` and expects the following parameters:
+
+- requests - is an OPTIONAL array of rpc requests that would be batched togehter when requesting the user for approval
+  - method - the RPC method being requested
+  - params - will include params specific to the RPC method
+
+The methods that are allowed to be batched together in the `wallet_connect` request is going to be going to dependant on each ecosystem that must be defined under the [Namespaces][namespaces] repo to define a CAIP-300 profile
+
+### Response
+
+The wallet will prompt the user with a dedicated UI to display the app requesting the authentication and allow the user to select which blockchain account to sign with.
+
+#### Success
+
+If approved, the wallet will return a list of signed, valid CACAOs for each account authorized on the networks requested by the `chains` property.
+
+```jsonc
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "result": { "result": unknown } || { error: { message: string, code: number } }[]
+}
+```
+
+The JSON-RPC response will include an array of requests which are going to be ordered in the same order as provided in the request.
+
+#### Failure
+
+Request will fail if rejected by the user or if parameters fail validation.
+
+The following Error responses MUST be used:
+
+- User Rejected Request
+  - code = 7000
+  - message = "User Rejected Connect"
+- Invalid Request Params
+  - code = 7001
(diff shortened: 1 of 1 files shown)


caip-300.md
```diff
@@ -0,0 +1,223 @@
...
+
+The application would interface with a wallet to make request as follows:
+
+```jsonc
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "method": "wallet_connect",
+  "params": {
+      "requests?": {
```
ganchoradkov, line 46: shouldn't there be least one request be mandatory? What would a wallet approve in an empty request?

pedrouid: Check the "Fallback Behavior" section

caip-300.md
```diff
@@ -0,0 +1,223 @@
...
+          "params": {...}
+      },
+      {
+          "method": "wallet_getCapabilities", // ERC-5792 request
+          "params": {...}
+      },
+      {
+          "method": "wallet_grantPermissions", // ERC-7715 request
+          "params": {...}
+      },
```
ganchoradkov, line 158: Does this standard allows for `personal_sign` to be sent via `wallet_connect`? As at the time of sending, the user address is unknown to the dapp

KannuSingh:
How will the standard handle the request if one of the rpc method requested via `wallet_connect` is not supported by the wallet and how in the case where account selected by user for approving the `wallet_connect` request don't support one of the request in the requested batch?

caip-300.md
pedrouid: Not really... you need to use wallet_authenticate

pedrouid:
Then the wallet must reject it with 7001 code for invalid method

Or partially reject it with an error message in the batch

jxom:
Can we consider adding `wallet_disconnect` to this proposal? Few reasons:

- Just like `wallet_connect` has been a missing RPC method for years, so has a way to programatically "disconnect" or "remove a host" from the wallet without the user having to do this in the wallet itself. 
- Adding a `wallet_disconnect` RPC method will further provide strengthened symmetry of this proposal.
- The absense of programmatic disconnect has been a painpoint in Ethereum libraries which interface Dapps to Wallets for years (have had to "polyfill" / "fake" the behavior).
- EIP-1193 provides `"connect"` and `"disconnect"` events, so having both `wallet_connect` and `wallet_disconnect` makes sense.

bumblefudge:
@lukaisailovic i'm seeing a commit adding jake as author but not one adding a _disconnect method 😅 

holler at me when there's an approval from jake so we can get this merged and visible on the website 😎

pedrouid:
@jxom there is only one concern with introducing wallet_disconnect here... the different payloads have different purposes

So is the intention to batch them together as a batch "disconnection"... for example you have wallet_authenticate and wallet_createSession... the former is providing a SIWE signature while the latter is enabling a live session with the wallet... are you disconnecting both??

Also regarding ERC-7715 would the wallet_disconnect perform a wallet_revokePermissions??

Definitely agree that the absence of `disconnect` as a pain-point for Ethereum but I want to make sure we understand the expected behavior for CAIP-300 since it's batching multiple requests

caip-300.md
```diff
@@ -0,0 +1,223 @@
...
+}
+```
+
+## Rationale
+
+An RPC method was choosen for batching because it enabled a smoother upgrade path for existing dapps and wallets to support it without much breaking changes. While the alternative would be to batch RPC requests at the transport layer this is not possible today without a breaking change.
+
+Therefore batching inside a single RPC request makes it easier to re-use the existing developer tooling and will enable both dapps and wallet to progressively support this new "connection" standard for wallets.
+
+The rationale for making this as generic as possible is to enable profiles to be defined for different ecosystems such as Bitcoin, Ethereum, Solana, Cosmos, Polkadot, etc. The alternative would be to create specific fields for "session", "authentication", "capabilities", "permissions", etc. But this would introduce a bias from one chain to another and not favor chain-agnosticisim.
```
bumblefudge, line 128:
```suggestion
The rationale for making this as generic as possible is to enable profiles to be defined for different ecosystems such as Bitcoin, Ethereum, Solana, Cosmos, Polkadot, etc. The alternative would be to create specific fields for "session", "authentication", "capabilities", "permissions", etc. But this would introduce a bias from one chain to another and harm chain-agnosticism.
```

bumblefudge:
i've lost the plot a bit, what are next steps here, @pedrouid ? this looks complete but I vaguely remember this being blocked on some of the non-DOM connection flows or something? Totally fine to switch it to draft if you're waiting for other implementers or open research questions to finish it.

RioAlfarizi321:
Screenshot_2024-10-27-15-13-09-12_dba69a5e82e939c3ddef13f99a115ca3

octavio12345300 (approved):

caip-300.md
```diff
@@ -0,0 +1,223 @@
...
+  "id": 1,
+  "jsonrpc": "2.0",
+  "method": "wallet_connect",
+  "params": {
+      "requests?": {
+        "method": string,
+        "params": unknown
+      }[],
+  }
+}
```
pedrouid, line 51:
```suggestion
  "params": {
    "<method_name>" : { ...params }
  }
}
```

caip-300.md
```diff
@@ -0,0 +1,223 @@
...
+```
+
+The JSON-RPC method is labelled as `wallet_connect` and expects the following parameters:
+
+- requests - is an OPTIONAL array of rpc requests that would be batched togehter when requesting the user for approval
+  - method - the RPC method being requested
+  - params - will include params specific to the RPC method
```
pedrouid, line 58:
```suggestion
The JSON-RPC method is labelled as `wallet_connect` and expects a list of RPC parameters indexed by the RPC method name.
```

caip-300.md
```diff
@@ -0,0 +1,223 @@
...
+
+#### Success
+
+If approved, the wallet will return a list of signed, valid CACAOs for each account authorized on the networks requested by the `chains` property.
+
+```jsonc
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "result": { "result": unknown } || { error: { message: string, code: number } }[]
```
pedrouid, line 74:
```suggestion
  "result": {
    "<method_name>": { "result": unknown } || { error: { message: string, code: number } }
  }
```

caip-300.md
```diff
@@ -0,0 +1,223 @@
...
+}
+```
+
+The JSON-RPC response will include an array of requests which are going to be ordered in the same order as provided in the request.
```
pedrouid, line 78:
```suggestion
The JSON-RPC response will include a list of RPC results or errors indexed by the corresponding RPC method name
```

caip-300.md
```diff
@@ -0,0 +1,223 @@
...
+```jsonc
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "result": [
+    {
+      "result": ["0xa...", "0xb..."] // fallback to ERC-1102 request
+    }
+  ]
+}
```
pedrouid, line 119:
```suggestion
  "result": {
      "eth_requestAccounts": ["0xa...", "0xb..."] // fallback to ERC-1102 request
   }
}
```

caip-300.md
```diff
@@ -0,0 +1,223 @@
...
+          "params": {...}
+      },
+      {
+          "method": "wallet_grantPermissions", // ERC-7715 request
+          "params": {...}
+      },
+
+    ]
+  }
+}
```
pedrouid, line 162:
```suggestion
  "params": {
    "wallet_authenticate": {...}, // CAIP-222 request
    "wallet_createSession": {...}, // CAIP-25 request
    "wallet_getCapabilities": {...}, // ERC-5792 request
    "wallet_grantPermissions": {...}, // ERC-7715 request
  }
}
```

caip-300.md
```diff
@@ -0,0 +1,223 @@
...
+        "result": {...} // CAIP-25 response
+    },
+    {
+        "result": {...} // ERC-5792 response
+    },
+    {
+        "result": {...} // ERC-7715 response
+    },
+
+  ]
```
pedrouid, line 185:
```suggestion
  "result": {
    "wallet_authenticate": { result: {...} } // CAIP-222 response
    "wallet_createSession": { result: {...} } // CAIP-25 response
    "wallet_getCapabilities": { result: {...} } // ERC-5792 response
    "wallet_grantPermissions": { result: {...} } // ERC-7715 response
  }
```

octavio12345300 (approved):

octavio12345300 (approved):

glitch-txs (commented):
Some general comments:

- The CAIP could use a chain specification to support requests for different chains.

- It would be nice to add support for transaction requests. Although Smart Sessions allow the user to trigger transactions on the go, a faster approach (for some scenarios and specific contexts) would be for the dapp to request a series of transactions to be executed immediately. In **some cases**, this simplification could even eliminate the need for a connection altogether.

octavio12345300 (approved):

caip-300.md
chris13524: Where does it specify that `personal_sign` is not allowed? This seems reasonable

octavio12345300 (changes requested):
list

PR #301, opened 2024-07-01
[Caip 25] add scopedProperties; update examples to include `eip155:0`, `scopedProperties`, and `sessionProperties`

bumblefudge:


jiexi:
how we do handle CAIP ChainId scopeString collisions? Should this be allowed, or should the `scopes` description be updated to also specify that `scopes` should not specify scopeStrings already defined outside of the `scopeObject` itself?

```
requiredScopes: {
    'eip155': {
        scopes: ['eip155:1'],
        methods: [
            'eth_sendTransaction',
            'eth_accounts',
            'eth_blockNumber',
            'eth_getBalance',
            'personal_sign',
        ],
        notifications: ['accountsChanged', 'chainChanged'],
    },
    'eip155:1': {
        methods: ['eth_blockNumber', 'eth_chainId'],
        notifications: ['accountsChanged', 'chainChanged'],
    },
},
```

bumblefudge:
> how we do handle CAIP ChainId scopeString collisions? Should this be allowed, or should the `scopes` description be updated to also specify that `scopes` should not specify scopeStrings already defined outside of the `scopeObject` itself?
> 
> ```
> requiredScopes: {
>     'eip155': {
>         scopes: ['eip155:1'],
>         methods: [
>             'eth_sendTransaction',
>             'eth_accounts',
>             'eth_blockNumber',
>             'eth_getBalance',
>             'personal_sign',
>         ],
>         notifications: ['accountsChanged', 'chainChanged'],
>     },
>     'eip155:1': {
>         methods: ['eth_blockNumber', 'eth_chainId'],
>         notifications: ['accountsChanged', 'chainChanged'],
>     },
> },
> ```

Is this a feature or a bug? I can certainly see the footgun (and maybe a BCP/non-normative advice CAIP is worth writing that includes this exact cornercase) but I can also think of reasons you might want to do it this way, particularly with `scopedProperties`. For example, imagine (it's hard, i know) that wallet-preference for RPC endpoints is rolled out imperfectly in the coming years, and people want to exploit this redundancy as a way to attempt rpcEndpoint suggestion with ERCs adopted piecemeal in the dapp ecosystem.  They could do this by creating a redundant `eip155:1` object, as you wrote above, and tacking on this guy:
```json
...
   scopedProperties: {
     "eip155:1": {
       "rpcEndpoint": "https://use-this-rpc-please.com"
     } 
```
This way, they can try the `eip155:1` `sessionScope` first, and  if it fails/timeouts/no-ops/etc, they can fall back on the dapp's RPC using the main/baseline `sessionScope`.

bumblefudge:
All that said, I don't have strong opinions, if the footgun outweighs the constructive usage (or if there strong security arguments), I'm happy to update the spec as per the consensus of the WG.

CAIPs/caip-25.md
```diff
@@ -123,43 +96,46 @@ Example:
...
 
-Respondent SHOULD ignore and drop from its response any properties not defined in this document or in another CAIP document extending this protocol which the respondent has implemented in its entirety; 
+Respondent SHOULD ignore and drop from its response any properties not defined in this document or in another CAIP document extending this protocol which the respondent has implemented in its entirety;
 similarly, the `requiredScopes`, `optionalScopes`, and `sessionScopes` arrays returned by the respondent SHOULD contain only valid [CAIP-217][] objects, and properties not defined in [CAIP-217][] SHOULD also be dropped from each of those objects.
+An exception can safely be made for `scopedProperties`, but caution is recommended for such extensions.
 
-Requesting applications are expected to persist all of these returned properties in the session object identified by the `sessionId`. 
+Requesting applications are expected to persist all of these returned properties in the session object identified by the `sessionId`.
```
hmalik88, line 113: Why do they need to persist? I'm not sure why it would be important to a dapp to persist connection information for a chain?

bumblefudge: `persist` is probably the wrong word here, the dapp isn't expected to persist the authorization/scope objects longer than or outside the "session" (or the sessionId) is kept alive. it's more like maintain state, to do deltas if the wallet calls something to add a session (this was in the more idempotent model of calling periodically to keep alive and potentially add or subtract capabilities).  but this only get included in the diff because I removed a trailing space when linting the new stuff 😅 

let's work on that in a separate PR once we've got the "sessionId not needed" RPCs written and do a big PR to update ALL language about session state and persistence?

hmalik88 (approved):

CAIPs/caip-25.md
bumblefudge: i'll opening a tracking issue and come back to this after the CAIP-285 stuff lands, as discussed on today's sync

CAIPs/caip-25.md
```diff
@@ -185,30 +162,39 @@ An example of a successful response follows:
...
         "methods": ["wallet_getPermissions", "wallet_creds_store", "wallet_creds_verify", "wallet_creds_issue", "wallet_creds_present"],
         "notifications": []
       },
       "cosmos": {
         ...
       }
     },      
+    "scopedProperties": {
+      "eip155:42161": {
+        "rpcDocuments": "https://example.com/wallet_extension.json",
```
adonesky1, line 175: is this getting moved from CAIP-217? If not I think its a bit confusing to use it as an example? @bumblefudge

bumblefudge: oooh, good catch, that was unintentional!

PR #303, opened 2024-07-06
Preliminary proposal for a wallet scope

vandan:
This proposal is meant to extend CAIP-217 `scopeStrings` to accept "wallet:" prefixes.

### DIFF
--- a/CAIPs/caip-217.md
+++ b/CAIPs/caip-217.md
@@ -63,7 +63,7 @@ The syntax of a `scopeObject` is as follows:
 
 Where:
 
-- {`scopeString`} (conditional) = EITHER a namespace identifier string registered in the CASA [namespaces][] registry to authorize multiple chains with identical properties OR a single, valid [CAIP-2][] identifier, i.e., a specific `chain_id` within a namespace.
+- {`scopeString`} (conditional) = EITHER a namespace identifier string registered in the CASA [namespaces][] registry to authorize multiple chains with identical properties OR a single, valid [CAIP-2][] identifier, i.e., a specific `chain_id` within a namespace OR a [CAIP-X][] `wallet_scope`.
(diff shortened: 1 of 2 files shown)


bumblefudge:
wouldn't it be simpler to just define `wallet` as a namespace in the [chainagnostic/namespaces](https://github.com/chainagnostic/namespaces/ ) repo? I'd recommend making a README.md and a caip2.md both in the initial PR, and maybe even state explicitly that there CANNOT and SHOULD NOT be any caip19.md, for example...

vandan:
> wouldn't it be simpler to just define `wallet` as a namespace in the [chainagnostic/namespaces](https://github.com/chainagnostic/namespaces/) repo? I'd recommend making a README.md and a caip2.md both in the initial PR, and maybe even state explicitly that there CANNOT and SHOULD NOT be any caip19.md, for example...

@bumblefudge That's actually where we started (`namespaces` repo). Also, CAIP-2 has fairly extensive language specifying that it's meant to define namespaces for blockchains/networks or ecosystems so I thought an alternative to CAIP-2 specifically for wallet scope would be less disruptive. 
I can submit another PR we've prepared for the `namespace` approach, but you may also come to the conclusion that attempting to fit wallet scope into CAIP-2's `namespace` + ":" + `reference` definition gets a little convoluted.

hmalik88:
Looks good, we should also update the language in CAIP-27 to reflect the fact that a request for the wallet namespaced methods should only be paired with a `wallet_scope` scope string.

bumblefudge:
Closed in favor of [namespaces#119](https://github.com/ChainAgnostic/namespaces/pull/119) as per RPC WG consensus

bumblefudge:
Closed in favor of https://github.com/ChainAgnostic/namespaces/pull/119 as per RPC WG consensus

PR #314, opened 2024-07-25
Use 'eth_sendTransaction' for CAIP-27 example. 'personal_sign' is not…

adonesky1:
'personal_sign' is an offchain signing method (more appropriate to the new proposed Wallet namespace) and therefore a bad example for `CAIP-27`


bumblefudge (approved):
I totally agree!

In fact, a second PR adding an additional example with `personal_sign` sent, targeted to `wallet:eip155` might be a useful addition, here and/or in a profile for the namespaces repo, i.e. in the `wallet/caip25.md` profile 😎 

1 review from hassan or two from non-codeowners and we merge

hmalik88 (approved):

PR #317, opened 2024-07-29
caip-200: fix markdown

apoorvlathey:
While building EIP.tools, I found that the CAIP-200's markdown for the metadata was invalid, so fixing that one character here.


bumblefudge (approved):

bumblefudge:
good catch, thanks!

PR #319, opened 2024-08-01
Standardize CAIP-2 Scoped Notification Support

adonesky1:
## Simple Summary

CAIP-319 defines a JSON-RPC method for a wallet to send notifications to a
caller in a specified context defined by a valid [scopeObject][CAIP-217] and,
optionally, tagged with a [sessionId][CAIP-171] for maintaining session
continuity.

## Abstract

This proposal aims to define a standard method for wallets to send notifications
to callers regarding events or state changes related to a specific,
previously-authorized target chain (such as nodes of a specific blockchain or
consensus community within a protocol). It requires a valid
[scopeObject][CAIP-217]. It MAY be tagged with a [sessionId][CAIP-171] if the
[CAIP-25][] session in which it is authorized is keyed by a sessionId (see
[CAIP-316][] for more details). These two properties MAY be inherited from a
persistent session created by [CAIP-25][], but could also be used as part of
other session management mechanisms.

## Motivation

The motivation for this proposal comes from the need for standardized,
chain-specific notifications from wallets to applications in a concurrent
multi-chain connection where methods and notifications with the same name may
exist across chains or namespaces.

### DIFF
--- a/CAIPs/caip-319.md
+++ b/CAIPs/caip-319.md
@@ -1,5 +1,5 @@
 ---
-caip: X
+caip: 319
 title: Wallet Notification JSON-RPC Method
 author: Alex Donesky (@adonesky1)
 discussions-to: https://github.com/ChainAgnostic/CAIPs/pull/X
@@ -11,8 +11,8 @@ requires: 2, 25, 171, 217
 
 ## Simple Summary
 
-CAIP-X defines a JSON-RPC method for a wallet to send notifications to a caller
-in a specified context defined by a valid [scopeObject][CAIP-217] and,
+CAIP-319 defines a JSON-RPC method for a wallet to send notifications to a
+caller in a specified context defined by a valid [scopeObject][CAIP-217] and,
 optionally, tagged with a [sessionId][CAIP-171] for maintaining session
 continuity.
 
@@ -87,8 +87,8 @@ Additionally, it MAY include an **optional parameter**:
 
 1. The application MUST check the scope against the identified session object
    before processing the notification.
-2. The application SHOULD verify that the notification.method is one it expects to
-   receive for that specific scope.
+2. The application SHOULD verify that the notification.method is one it expects
+   to receive for that specific scope.
 3. The application MAY apply other logic or validation to the notification data.
 4. The application MAY choose to ignore notifications it doesn't recognize or
    considers invalid.
--- a/CAIPs/caip-319.md
+++ b/CAIPs/caip-319.md
@@ -12,7 +12,7 @@ requires: 2, 25, 171, 217
 ## Simple Summary
 
 CAIP-319 defines a JSON-RPC method for a wallet to send notifications to a
-caller in a specified context defined by a valid [scopeObject][CAIP-217] and,
+caller in a context authorized by or in a valid [scopeObject][CAIP-217] and,
 optionally, tagged with a [sessionId][CAIP-171] for maintaining session
 continuity.
 
@@ -73,10 +73,9 @@ follows:
 The JSON-RPC method is labeled as `wallet_notify` and expects two required
 parameters:
 
-- **scope** - a valid `scopeObject` previously authorized to the caller and
-  persisted in the session identified by `sessionId`
+- **scope** - a valid [CAIP-2][] chainId previously authorized to the caller within a [CAIP-25][] session
 - **request** - an object containing the fields:
-  - **method** - JSON-RPC notification method name
+  - **method** - JSON-RPC notification method name previously authorized to the caller within a [CAIP-25][] session
   - **params** - JSON-RPC notification parameters
 
 Additionally, it MAY include an **optional parameter**:
@@ -100,7 +99,6 @@ wallet or user agent SHOULD NOT wait for a response before continuing its
 operations.
 
 ## Links
-
 [CAIP-2]: https://chainagnostic.org/CAIPs/caip-2
 [CAIP-25]: https://chainagnostic.org/CAIPs/caip-25
 [CAIP-171]: https://chainagnostic.org/CAIPs/caip-171
--- a/CAIPs/caip-319.md
+++ b/CAIPs/caip-319.md
@@ -74,7 +74,7 @@ The JSON-RPC method is labeled as `wallet_notify` and expects two required
 parameters:
 
 - **scope** - a valid [CAIP-2][] chainId previously authorized to the caller within a [CAIP-25][] session
-- **request** - an object containing the fields:
+- **notification** - an object containing the fields:
   - **method** - JSON-RPC notification method name previously authorized to the caller within a [CAIP-25][] session
   - **params** - JSON-RPC notification parameters
 
--- a/CAIPs/caip-319.md
+++ b/CAIPs/caip-319.md
@@ -14,13 +14,13 @@ requires: 2, 25, 171, 217
 CAIP-319 defines a JSON-RPC method for a wallet to send notifications to a
 caller in a context authorized by or in a valid [scopeObject][CAIP-217] and,
 optionally, tagged with a [sessionId][CAIP-171] for maintaining session
-continuity.
+continuity if applicable.
 
 ## Abstract
 
 This proposal aims to define a standard method for wallets to send notifications
 to callers regarding events or state changes related to a specific,
-previously-authorized target chain (such as nodes of a specific blockchain or
+previously-authorized target network (such as nodes of a specific blockchain or
 consensus community within a protocol). It requires a valid
 [scopeObject][CAIP-217]. It MAY be tagged with a [sessionId][CAIP-171] if the
 [CAIP-25][] session in which it is authorized is keyed by a sessionId (see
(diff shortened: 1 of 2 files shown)


jiexi (approved):

bumblefudge:
Makes perfect sense to me.  My only question is how this interacts with the namespace-wide (or "carbon copy" compact notation, to be more precise) authorizations.  Imagine, for example, a session created by the following CAIP-25 response:

```json
{
...
  sessionScopes: {
    "bip122:1" { ... },
    "eip155": {
      "chains": ["1","42069"],
      ...
...
}
```

1. can the wallet send an update to "eip155" (and have it apply to both "1" and "42069")?
2. can the wallet send an update to "eip155:1" ?

It's probably a simple answer in both cases, and I am less concerned with what the answer is, than in the spec being explicit in advance about both 😄 

Other than that, LGTM?

adonesky1:
> can the wallet send an update to "eip155" (and have it apply to both "1" and "42069")?

I wouldn't think so. I would assume notifications, like CAIP-27 invokeMethod requests, are scoped to the full CAIP-2 chainId. I can't think of a use case for namespace wide notifications 🤔 but if you can then we could consider ways to include that.

CAIPs/caip-319.md
```diff
@@ -0,0 +1,112 @@
...
+status: Draft
+type: Standard
+created: 2024-08-01
+requires: 2, 25, 171, 217
+---
+
+## Simple Summary
+
+CAIP-319 defines a JSON-RPC method for a wallet to send notifications to a
+caller in a specified context defined by a valid [scopeObject][CAIP-217] and,
```
bumblefudge, line 15:
```suggestion
caller in a specified context authorized by or in a valid [scopeObject][CAIP-217] and,
```

This is a slightly obtuse phrasing, but I think it captures what I mean about the case where CAIP-25 authorized
```json
{
...
  sessionScopes: {
    "bip122:1" { ... },
    "eip155": {
      "scopes": ["1","42069"],
      ...
    }
...
}
```

and a later call to CAIP-319 or CAIP-27 targets `eip155:1`.  the technicality is that `eip155:1` is NOT a valid scopeObject in this case, even though it's a valid target authorized IN a valid scopeObject.

adonesky1: Yep. Personally, per my comment [here](https://github.com/ChainAgnostic/CAIPs/pull/319#issuecomment-2265503607) I would think we should change the wording in CAIP-27 to exclude the possibility of calling CAIP-27 (or CAIP-319) with anything other than a **fully formed CAIP-2 chainId**  for the `scope`

adonesky1: https://github.com/ChainAgnostic/CAIPs/pull/319/commits/930ed61828f32ecb2eb23888db6cb9fd775f9735 shows what I mean. I think we should make this change in CAIP-27 as well

adonesky1: Its inaccurate to characterize the CAIP-27 `scope` param as a `scopeObject`... https://github.com/ChainAgnostic/CAIPs/blob/6922e7446cd90add44feb899417164b8f0a7a1e2/CAIPs/caip-27.md?plain=1#L85

Its a CAIP-2 chainId?

bumblefudge: whoops, it currently should read "the [`scopeString`](https://chainagnostic.org/CAIPs/caip-217#definition) of a valid `scopeObject`, which _could_ be a CAIP-2 or a namespace, that's what was meant by the current text. and I agree we should make a normative change, I'm just kind of stuck on the best change to make:

1.) any "authorized" CAIP-2 is legal, even if it refers to one [of potentially many] `scopes[]` members _in_ a namespace-wide, "carbon copy" scopeObject.  This puts the onus on the wallet to make sure that CAIP-2 has been authorized, which is probably fine sense every wallet just thinks of a namespace-wide authorization with multiple scopes as separate per-chain authorization objects anyways under the hood.

or

2.) be more explicit and say that ONLY a `scopeString` that IS ALSO a CAIP-2 AND has been authorized explicitly, removing the possibility of [case #2 here](https://github.com/ChainAgnostic/CAIPs/pull/319#issuecomment-2265126837)

adonesky1: > 2.) be more explicit and say that ONLY a scopeString that IS ALSO a CAIP-2 AND has been authorized explicitly, removing the possibility of https://github.com/ChainAgnostic/CAIPs/pull/319#issuecomment-2265126837

This doesn't make sense to me. 

->>>

> 1.) any "authorized" CAIP-2 is legal, even if it refers to one [of potentially many] scopes[] members in a namespace-wide, "carbon copy" scopeObject, i.e. https://github.com/ChainAgnostic/CAIPs/pull/319#issuecomment-2265126837. This puts the onus on the wallet to make sure that a given CAIP-2 has been authorized, which is probably fine since every wallet just thinks of a namespace-wide authorization with multiple scopes as separate per-chain authorization objects anyways under the hood.

Definitely this. I don't see any reason for wallets (or the protocol generically) to treat or think of 

> scopes[] members in a namespace-wide, "carbon copy" scopeObject,

As different from individually enumerated scopes in any way at all. 

And then furthermore, per your comment [here](https://github.com/ChainAgnostic/CAIPs/pull/320#discussion_r1711286113), is there any outstanding reason we can't just constrain CAIP-27 (and this CAIP) to be only **full CAIP-2 chainIds** not `scopeStrings`?

bumblefudge: Fair enough! I'll open that PR before today's meeting.

CAIPs/caip-319.md
```diff
@@ -0,0 +1,110 @@
...
+created: 2024-08-01
+requires: 2, 25, 171, 217
+---
+
+## Simple Summary
+
+CAIP-319 defines a JSON-RPC method for a wallet to send notifications to a
+caller in a context authorized by or in a valid [scopeObject][CAIP-217] and,
+optionally, tagged with a [sessionId][CAIP-171] for maintaining session
+continuity.
```
bumblefudge, line 17:
```suggestion
continuity if applicable.
```

CAIPs/caip-319.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Abstract
+
+This proposal aims to define a standard method for wallets to send notifications
+to callers regarding events or state changes related to a specific,
+previously-authorized target chain (such as nodes of a specific blockchain or
```
bumblefudge, line 23:
```suggestion
previously-authorized target network (such as nodes of a specific blockchain or
```

CAIPs/caip-319.md
```diff
@@ -0,0 +1,110 @@
...
+[CAIP-25][] session in which it is authorized is keyed by a sessionId (see
+[CAIP-316][] for more details). These two properties MAY be inherited from a
+persistent session created by [CAIP-25][], but could also be used as part of
+other session management mechanisms.
+
+## Motivation
+
+The motivation for this proposal comes from the need for standardized,
+chain-specific notifications from wallets to applications in a concurrent
+multi-chain connection where methods and notifications with the same name may
```
bumblefudge, line 35:
```suggestion
multi-chain connection where methods and notifications with the same namespace may
```

CAIPs/caip-319.md
```diff
@@ -0,0 +1,110 @@
...
+1. The application MUST check the scope against the identified session object
+   before processing the notification.
+2. The application SHOULD verify that the notification.method is one it expects
+   to receive for that specific scope.
+3. The application MAY apply other logic or validation to the notification data.
+4. The application MAY choose to ignore notifications it doesn't recognize or
+   considers invalid.
+
+### Response
+~~~~
```
bumblefudge, line 96:
```suggestion

```

CAIPs/caip-319.md
```diff
@@ -0,0 +1,110 @@
...
+As this is a notification, no response is expected from the application. The
+wallet or user agent SHOULD NOT wait for a response before continuing its
+operations.
```
bumblefudge, line 99:
```suggestion
As this syntax simply provides a wrapper to a standard [JSON-RPC notification][]
and functions as one itself, no response is expected from the application. The
wallet or user agent SHOULD NOT wait for a response before continuing its
operations.
```

CAIPs/caip-319.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Links
+[CAIP-2]: https://chainagnostic.org/CAIPs/caip-2
+[CAIP-25]: https://chainagnostic.org/CAIPs/caip-25
+[CAIP-171]: https://chainagnostic.org/CAIPs/caip-171
+[CAIP-217]: https://chainagnostic.org/CAIPs/caip-217
+[CAIP-316]: https://chainagnostic.org/CAIPs/caip-316
```
bumblefudge, line 106:
```suggestion
[CAIP-316]: https://chainagnostic.org/CAIPs/caip-316
[JSON-RPC notification]: https://www.jsonrpc.org/specification#notification
```

bumblefudge (approved):
let's ship it and move onto the CAIP-27 update

CAIPs/caip-319.md
```diff
@@ -0,0 +1,111 @@
...
+    }
+  }
+}
+```
+
+The JSON-RPC method is labeled as `wallet_notify` and expects two required
+parameters:
+
+- **scope** - a valid [CAIP-2][] chainId previously authorized to the caller within a [CAIP-25][] session
+- **request** - an object containing the fields:
```
hmalik88, line 77:
```suggestion
- **notification** - an object containing the fields:
```

hmalik88 (approved):

PR #324, opened 2024-08-13
This EIP-2256 has not been used for more than 6 months caip-2.md

Ocheretovich:
This EIP-2256 has not been used for more than 6 months caip-2.md


obstropolos:
This is a reference contained within the CAIP, please don't break it.

PR #327, opened 2024-09-05
Refine 'scopes' naming conventions

adonesky1:
~Seems like calling this property `scopes` rather than `chainIds` causes unnecessary confusion. The array can only ever contain full CAIP-2 chainIds.~

A proposed type & name change for CAIP-217:

1. Why not only allow `references` for the given `namespace` when a `scopeString`  is a `namespace`?

2. In my opinion the name `scope` should be reserved for the entire `CAIP-217` record:
In Typescript Terms:
`Scope = Record<ScopeString, ScopeObject>`


bumblefudge (approved):
I agree that `references` is better than `scopes`, so approving, but I'm wondering if something even more tangible or obvious would be even better... like "referents" or "targets" or something? i'll have a think on it.

note: also found some typos in the current text 🤦

CAIPs/caip-217.md
```diff
@@ -64,10 +62,10 @@ The syntax of a `scopeObject` is as follows:
...
-  - This property MUST NOT be present if the scope is an entire [namespace][namespaces] in which no `chainId`s are defined or no [CAIP-2] profile has been published.
-  - This property SHOULD be present if the scope is an entire [namespace][namespaces] in which `chainId`s are defined. An empty `scopes` array MUST NOT be interpreted as authorizing an entire namespace in which a finite list of [CAIP-2] value could be set, but rather, as applying equal to zero members of that finite list until 1 or more of them are added to `scopes`.
+  - This property MUST NOT be present if the scope is an entire [namespace][namespaces] in which no `references` are defined or no [CAIP-2] profile has been published.
+  - This property SHOULD be present if the scope is an entire [namespace][namespaces] in which `chainId`s are defined. An empty `references` array MUST NOT be interpreted as authorizing an entire namespace in which a finite list of [CAIP-2] value could be set, but rather, as applying equal to zero members of that finite list until 1 or more of them are added to `references`.
```
bumblefudge, line 68:
```suggestion
  - This property SHOULD be present if the scope is an entire [namespace][namespaces] in which `chainId`s are defined. An empty `references` array MUST NOT be interpreted as authorizing an entire namespace in which a finite list of [CAIP-2] values could be set, but rather, as applying equally to zero members of that finite list until 1 or more of them are added to `references`.
```

bumblefudge: @hmalik88

hmalik88 (approved):

PR #344, opened 2025-01-07
CAIP-320 wallet switch chain JSON-RPC

tuna1207:
This CAIP defines a JSON-RPC method for a chain agnostic wallet to switch the current active chain of the wallet using the id specified in CAIP-2.

### DIFF
--- a/CAIPs/caip-320.md
+++ b/CAIPs/caip-320.md
@@ -4,7 +4,7 @@
 caip: 320
 title: Wallet Switch Chain JSON-RPC Method
 author: Anh Tu Nguyen (@tuna1207), Chaitanya Potti (@chaitanyapotti), Tai Nguyen (@tanguyenvn)
-discussions-to: <URL(s); if multiple, list separated by , without " or []>
+discussions-to: https://github.com/ChainAgnostic/CAIPs/discussions/343
 status: Draft
 type: Standard
 created: 2025-01-06
--- a/CAIPs/caip-320.md
+++ b/CAIPs/caip-320.md
@@ -1,0 +1,149 @@
+---
+# Every document starts with a front matter in YAML enclosed by triple dashes.
+# See https://jekyllrb.com/docs/front-matter/ to learn more about this concept.
+caip: 320
+title: Wallet Switch Chain JSON-RPC Method
+author: Anh Tu Nguyen (@tuna1207), Chaitanya Potti (@chaitanyapotti), Tai Nguyen (@tanguyenvn)
+discussions-to: <URL(s); if multiple, list separated by , without " or []>
+status: Draft
+type: Standard
+created: 2025-01-06
+updated: 2025-01-06
+requires: 2, 25, 171, 316
+---
+
+<!--You can leave these HTML comments in your merged EIP and delete the visible duplicate text guides, they will not appear and may be helpful to refer to if you edit it again. This is the suggested template for new EIPs. Note that an EIP number will be assigned by an editor. When opening a pull request to submit your EIP, please use an abbreviated title in the filename, `eip-draft_title_abbrev.md`. The title should be 44 characters or less.-->
+
+## Simple Summary
+
+<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the CAIP.-->
+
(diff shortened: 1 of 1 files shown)


bumblefudge:
Hey, this is an interesting addition to the repertoire here! I think I understand the use-case fairly well, but just for my understanding and to get this closer to mergeable, could you maybe sketch out for me an end-to-end example?

Ideally, a detailed example, i.e., involving a multi-namespace dapp, and a multi-namespace wallet creating a session, and a realistic example, i.e. involving two namespaces that already use an "active chain" concept and have native RPCs to switch active chains, maybe eip155 and solana? and show why wallet and/or dapps want this, versus using each namespace's native RPCs via CAIP-27 over a CAIP-25 connection?

CAIPs/caip-320.md
```diff
@@ -0,0 +1,149 @@
+---
+# Every document starts with a front matter in YAML enclosed by triple dashes.
+# See https://jekyllrb.com/docs/front-matter/ to learn more about this concept.
+caip: 320
```
bumblefudge, line 4:
```suggestion
caip: 344
```
Nit: the PR that adds the CAIP gives its number to the CAIP :D

CAIPs/caip-320.md
```diff
@@ -0,0 +1,149 @@
...
+requires: 2, 25, 171, 316
+---
+
+<!--You can leave these HTML comments in your merged EIP and delete the visible duplicate text guides, they will not appear and may be helpful to refer to if you edit it again. This is the suggested template for new EIPs. Note that an EIP number will be assigned by an editor. When opening a pull request to submit your EIP, please use an abbreviated title in the filename, `eip-draft_title_abbrev.md`. The title should be 44 characters or less.-->
+
+## Simple Summary
+
+<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the CAIP.-->
+
+CAIP-320 introduces the wallet_switchChain method for a chain agnostic wallet to switch the current active chain of the wallet using the id specified in CAIP-2.
```
bumblefudge, line 21:
```suggestion
CAIP-344 introduces the wallet_switchChain method for a chain agnostic wallet to switch the current active chain of the wallet using the id specified in CAIP-2.
```

CAIPs/caip-320.md
```diff
@@ -0,0 +1,149 @@
...
+```
+
+The JSON-RPC method is labeled as `wallet_switchChain` and expects two parameters, **two of them required**:
+
+- **sessionId** (conditional) - [CAIP-171][] `SessionId` disambiguates an open session in a multi-session actor; it is required in some sessions, such as [CAIP-25][] sessions created by a response containing one, and SHOULD be omitted in other sessions, such as [CAIP-25] sessions created by a response not containing one (see [CAIP-316]).
+- **chainId** (required) - a valid [CAIP-2][] network identifier, previously authorized by or within a `scopeObject` in the active session
+
+### Validation
+
+1. A respondent SHOULD check the `scope` against active session's `scopeObject`s before executing or responding to such a request, and SHOULD invalidate a request for a scope not previously authorized.
```
bumblefudge, line 74: but how does it disambiguate if there are two different scopes that it could apply, e.g. `eip155` and `eip155:1`? Or are you assuming a use-case where the scopes `eip155:1` and `eip155:10` have been authorized, and those are the only two values which can be valid for `chainId`?

CAIPs/caip-320.md
```diff
@@ -0,0 +1,149 @@
...
+  }
+}
+```
+
+##### Trusted Failure Codes
+
+More informative error messages MAY be sent in trusted-counterparty circumstances.
+The core error messages over trusted connections are as follows:
+
+The valid error messages codes are the following:
```
bumblefudge, line 117: Once I'm clearer on the answers above, I _may_ have to return to this section, it's still a little unclear to me if I should be evaluating these error codes as a _subset_ of CAIP-25 errors applicable/valid for this method, or as an _alternate set_ of errors that would be applicable if this were called _outside_ of a CAIP-25 connection!

RGraham1931 (approved):

RGraham1931 (changes requested):
That's amazing thank you

bumblefudge:
@tuna1207 ?

PR #345, opened 2025-02-17
CAIP-345: Wallet service URL property

chris13524:
This proposal defines the wallet service property, `caip345`, for use in CAIP-25. Wallets set this to indicate that certain methods, instead of being handled by the CAIP-25 session, will instead be sent to a JSON-RPC HTTP endpoint. Compatible apps that support this will be able to call these RPC methods, without interactivity with the actual wallet application or user.

### DIFF
--- a/CAIPs/caip-345.md
+++ b/CAIPs/caip-345.md
@@ -6,7 +6,7 @@ discussions-to: https://github.com/ChainAgnostic/CAIPs/discussions/345
 status: Draft
 type: Standard
 created: 2025-02-17
-updated: 2025-02-17
+updated: 2025-06-24
 requires: 25
 ---
 
@@ -20,7 +20,7 @@ This proposal defines the wallet service property, `caip345`, for use in CAIP-25
 
 ## Motivation
 <!--The motivation is critical for CAIP. It should clearly explain why the state of the art is inadequate to address the problem that the CAIP solves. CAIP submissions without sufficient motivation may be rejected outright.-->
-It is sub-optimal UX to redirect the user to their wallet in order to handle RPC requests that are non-actionable to them. This is especially relevant for protocols such as WalletConnect which are used in especially distributed environments such as mobile wallets or custodial solutions. In these contexts, actioning a wallet RPC request can involve significant effort.
+It is sub-optimal UX to redirect the user to their wallet in order to handle RPC requests that are non-actionable to them. This is especially relevant for protocols such as WalletConnect which is used in more distributed environments such as mobile wallets or custodial solutions. In these contexts, actioning a wallet RPC request can involve significant effort.
 
 Examples of non-actionable wallet requests include:
 - [EIP-5792](https://eips.ethereum.org/EIPS/eip-5792#wallet_getcapabilities) `wallet_getCapabilities`
@@ -87,9 +87,9 @@ type Caip25Properties = {
 };
 ```
 
-If a method is listed in the `caip345` property, then the same method MUST be listed as of the CAIP-25 session for the same scopes. Wallets MUST implement fallback handling for all wallet service methods, in-case the app does not implement this CAIP.
+If a method is included in the `caip345` property, then the same method MUST be included in the CAIP-25 session for the same scopes. Wallets MUST implement fallback handling for all wallet service methods, in case the app does not implement this CAIP.
 
-Below is an example support for `wallet_getAssets` which is only supported on `eip155` scope:
+Below is an unverifed, non-normative, example for `wallet_getAssets` which is only supported in the `eip155` scope:
 
 ```json
 "scopedProperties": {
@@ -102,7 +102,7 @@ Below is an example support for `wallet_getAssets` which is only supported on `e
 }
 ```
 
-Below is an example for ERC-7836, which is also only valid in the `eip155` scope:
+Below is an unverifed, non-normative, example for ERC-7836, which is also only valid in the `eip155` scope:
 
 ```json
 "scopedProperties": {
@@ -126,21 +126,21 @@ There must still be 1 canonical wallet service URL for a given method (if availa
 
 ### Different wallet service depending on account
 
-There was consideration for being able to specify different wallet service URLs for different accounts. However, this would not make sense in the context of CAIP-25 because there is no mechanism to scope the methods themselves to particular accounts. If we provided a mechanism to scope the methods to accounts in this CAIP, the app may still try to send the method requests for non-listed accounts directly to the wallet.
+There was consideration for being able to specify different wallet service URLs for different accounts. However, this would not make sense in the context of CAIP-25 because there is no mechanism to scope the methods to particular accounts. If this CAIP provided a way to scope the methods to particular accounts, the app may still try to send the method requests for non-listed accounts directly to the wallet.
 
 ### Map methods to wallet services, instead of methods in array
 
-There was consideraiton for defining the standard to have a unique wallet service URL for every single method. However, this would cause excessive bandwidth consumption if the same URL were to be used for multiple methods which we think is the more likely case.
+There was consideraiton for defining the standard to have a separate wallet service URL for every single method. However, this would cause excessive bandwidth consumption if the same URL were to be used for multiple methods which we think is the more likely case.
 
 ### Not supporting custom headers
 
-Specifying custom headers to use in the wallet service request is not supported. This is because in browsers, custom headers will create pre-flight `OPTIONS` requests, increasing bandwidth and server load.
+Specifying custom headers to use in the wallet service request is not supported. This is because in browsers, custom headers will create pre-flight `OPTIONS` requests, increasing bandwidth and server load. Instead, the necessary parameters should be passed via query params.
 
-### Apps using `connect-src` in Content-Security-Policy
+### Web apps using `connect-src` in Content-Security-Policy
 
-Many apps specify `connect-src` in their CSP which prevents the application and its libraries from connecting to URLs that aren't pre-specified. Since each wallet may use their own wallet service hosted on unique origins, it's not possible nor advisable to dynamically set the `connect-src` value necessary to allow a connection to flow properly.
+Many web apps specify `connect-src` in their CSP which prevents the app and its libraries from connecting to URLs that aren't pre-specified. Since each wallet may use their own wallet service hosted on unique origins, it's not possible/advisable to dynamically set the `connect-src` value as-necessary.
 
-To circumvent this, it is RECOMMENDED that such apps consume a minimal proxy service (such as a server function) in order to forward the wallet RPC request to the destination wallet service. This proxy service is known by the app, and has a fixed origin URL, which allows putting in `connect-src`. The service would be implemented by the app, or by a third-party.
+However to support this use case, such apps can consume a minimal proxy service (such as a server function) which will forward the wallet RPC request to the destination wallet service. This proxy service is known by the app, and has a fixed origin URL, which allows placing it in `connect-src`. The proxy service could be implemented by the app, or by a third-party.
 
 The mechanism by which this proxy service is implemented or consumed is outside the scope of this CAIP.
 
@@ -172,6 +172,8 @@ Similarly, the wallet service would have visibility into the wallet RPC requests
 ## Backwards Compatibility
 <!--All CAIPs that introduce backwards incompatibilities must include a section describing these incompatibilities and their severity. The CAIP must explain how the author proposes to deal with these incompatibilities. CAIP submissions without a sufficient backwards compatibility treatise may be rejected outright.-->
 This new property is fully backwards-compatible with CAIP-25.
+
+As also mentioned above, wallets MUST implement fallback handling for all wallet service methods, in case the app does not implement this CAIP.
 
 ## References
 <!--Links to external resources that help understanding the CAIP better. This can e.g. be links to existing implementations. See CONTRIBUTING.md#style-guide . -->
--- a/CAIPs/caip-345.md
+++ b/CAIPs/caip-345.md
@@ -130,7 +130,7 @@ There was consideration for being able to specify different wallet service URLs 
 
 ### Map methods to wallet services, instead of methods in array
 
-There was consideraiton for defining the standard to have a separate wallet service URL for every single method. However, this would cause excessive bandwidth consumption if the same URL were to be used for multiple methods which we think is the more likely case.
+There was consideration for defining the standard to have a separate wallet service URL for every single method. However, this would cause excessive bandwidth consumption if the same URL were to be used for multiple methods which we think is the more likely case.
 
 ### Not supporting custom headers
 
--- a/CAIPs/caip-345.md
+++ b/CAIPs/caip-345.md
@@ -146,6 +146,7 @@ The mechanism by which this proxy service is implemented or consumed is outside 
 
 ## Test Cases
 <!--Please add diverse test cases here if applicable. Any normative definition of an interface requires test cases to be implementable. -->
+
 Valid wallet service:
 ```json
 {
--- a/CAIPs/caip-345.md
+++ b/CAIPs/caip-345.md
@@ -89,7 +89,7 @@ type Caip25Properties = {
 
 If a method is included in the `caip345` property, then the same method MUST be included in the CAIP-25 session for the same scopes. Wallets MUST implement fallback handling for all wallet service methods, in case the app does not implement this CAIP.
 
-Below is an unverifed, non-normative, example for `wallet_getAssets` which is only supported in the `eip155` scope:
+Below is an unverified, non-normative, example for `wallet_getAssets` which is only supported in the `eip155` scope:
 
 ```json
 "scopedProperties": {
--- a/CAIPs/caip-345.md
+++ b/CAIPs/caip-345.md
@@ -23,7 +23,7 @@ This proposal defines the wallet service property, `caip345`, for use in CAIP-25
 It is sub-optimal UX to redirect the user to their wallet in order to handle RPC requests that are non-actionable to them. This is especially relevant for protocols such as WalletConnect which is used in more distributed environments such as mobile wallets or custodial solutions. In these contexts, actioning a wallet RPC request can involve significant effort.
 
 Examples of non-actionable wallet requests include:
-- [EIP-5792](https://eips.ethereum.org/EIPS/eip-5792#wallet_getcapabilities) `wallet_getCapabilities`
+- [EIP-5792 `wallet_getCapabilities`](https://eips.ethereum.org/EIPS/eip-5792#wallet_getcapabilities)
 - [EIP-5792 `wallet_getCallsStatus`](https://eips.ethereum.org/EIPS/eip-5792#wallet_getcallsstatus-rpc-specification)
 - [ERC-7836](https://github.com/ethereum/ERCs/pull/758) `wallet_prepareCalls` and `wallet_sendPreparedCalls`
 - [ERC-7811](https://eips.ethereum.org/EIPS/eip-7811) `wallet_getAssets`
--- a/CAIPs/caip-345.md
+++ b/CAIPs/caip-345.md
@@ -163,7 +163,7 @@ Invalid wallet service:
 
 ## Security Considerations
 <!--Please add an explicit list of intra-actor assumptions and known risk factors if applicable. Any normative definition of an interface requires these to be implementable; assumptions and risks should be at both individual interaction/use-case scale and systemically, should the interface specified gain ecosystem-namespace adoption. -->
-The wallet service would bear the security resonsibility of responding to these wallet RPC requests. Wallets should make informed decisions about which providers they use for this.
+The wallet service would bear the security responsibility of responding to these wallet RPC requests. Wallets should make informed decisions about which providers they use for this.
 
 ## Privacy Considerations
 <!--Please add an explicit list of intra-actor assumptions and known risk factors if applicable. Any normative definition of an interface requires these to be implementable; assumptions and risks should be at both individual interaction/use-case scale and systemically, should the interface specified gain ecosystem-namespace adoption. -->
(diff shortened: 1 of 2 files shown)


CAIPs/caip-345.md
```diff
@@ -0,0 +1,104 @@
...
+requires: 25
+---
+
+## Simple Summary
+<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the CAIP.-->
+This CAIP defines a mechanism for CAIP-25 wallets to advertize support for a "wallet service" which can handle requests for specific wallet RPC methods instead of apps sending these requests directly to the wallet app.
+
+## Abstract
+<!--A short (~200 word) description of the technical issue being addressed.-->
+A short (~200 word) description of the technical issue being addressed.
```
chris13524, line 19: TODO write this

CAIPs/caip-345.md
```diff
@@ -0,0 +1,104 @@
...
+};
+```
+
+An easy way for apps to find the wallet service URL for a given method would be:
+
+```javascript
+walletService.find(s => s.methods.includes(method)).url
+```
+
+The `walletService` key can be used in both `sessionProperties` and `scopedProperties`, depending on what namespaces the method(s) in question are supported on.
```
chris13524, line 57: What about conflicting scenarios here?

CAIPs/caip-345.md
```diff
@@ -0,0 +1,104 @@
...
+## Motivation
+<!--The motivation is critical for CAIP. It should clearly explain why the state of the art is inadequate to address the problem that the CAIP solves. CAIP submissions without sufficient motivation may be rejected outright.-->
+In protocols such as WalletConnect it is bad UX to redirect the user to a wallet to handle requests that are non-actionable to them. Examples of this include:
+- `wallet_getCapabilities` - is this real? point to EIP
+- `wallet_getCallsStatus` - point to EIP
+- `wallet_getAssets` - point to EIP
+
+By defining a way for wallets to send requests to an external URL, the requests can be satisfied without needing the wallet app to be open.
+
+## Specification
```
chris13524, line 30: TODO specify wallet service endpoint design. E.g. use POST and be JSON-RPC compatible

Also are these responses cacheable or not? I guess not since POST.

CAIPs/caip-345.md
```diff
@@ -0,0 +1,104 @@
...
+
+Apps SHOULD use the wallet service when available for a method, instead of calling the wallet directly.
+
+Wallets SHOULD implement fallback handling for all wallet service methods.
+
+Wallets MUST NOT include the same method multiple times or with different (conflicting) wallet service URLs. Apps SHOULD NOT attempt to recover from multiple or conflicting wallet service URLs, but MAY use the first URL available for the method as a convenience for implementation.
+
+```ts
+type WalletService = {
+    url: string,
```
chris13524, line 41: TODO mention the case where wallet wants authentication on this service and should include a token in this URL. What about an RPC to refresh the token?

CAIPs/caip-345.md
```diff
@@ -0,0 +1,104 @@
...
+walletService.find(s => s.methods.includes(method)).url
+```
+
+The `walletService` key can be used in both `sessionProperties` and `scopedProperties`, depending on what namespaces the method(s) in question are supported on.
+
+Below is an example support for `wallet_getAssets` which is only supported on `eip155` namespaces:
+
+```json
+"scopedProperties": {
+    "eip155": {
```
jakubuid, line 63: What's the example of `wallet_getAssets`, which is supported on eip155 and solana?

```
"scopedProperties": {
    "eip155": {
	  "walletService": [{
            	"url": "<wallet service URL>",
            	"methods": ["wallet_getAssets"]
        }],
    },
  "solana": {
	  "walletService": [{
            	"url": "<wallet service URL>",
            	"methods": ["wallet_getAssets"]
        }],
    }
}
```

bumblefudge:
I'm thinking maybe this should be "Draft PR" until most of those TODOs get filled out? Interesting stuff, overall...

CAIPs/caip-345.md
```diff
@@ -0,0 +1,104 @@
...
+This CAIP defines a mechanism for CAIP-25 wallets to advertize support for a "wallet service" which can handle requests for specific wallet RPC methods instead of apps sending these requests directly to the wallet app.
+
+## Abstract
+<!--A short (~200 word) description of the technical issue being addressed.-->
+A short (~200 word) description of the technical issue being addressed.
+
+## Motivation
+<!--The motivation is critical for CAIP. It should clearly explain why the state of the art is inadequate to address the problem that the CAIP solves. CAIP submissions without sufficient motivation may be rejected outright.-->
+In protocols such as WalletConnect it is bad UX to redirect the user to a wallet to handle requests that are non-actionable to them. Examples of this include:
+- `wallet_getCapabilities` - is this real? point to EIP
```
bumblefudge, line 24:
```suggestion
- `wallet_getCapabilities` - Defined in [EIP-5792](https://eips.ethereum.org/EIPS/eip-5792#wallet_getcapabilities)
```

CAIPs/caip-345.md
```diff
@@ -0,0 +1,104 @@
...
+- `wallet_getCallsStatus` - point to EIP
```
bumblefudge, line 25:
```suggestion
- `wallet_getCallsStatus` - Defined in [EIP-5792](https://eips.ethereum.org/EIPS/eip-5792#wallet_getcallsstatus-rpc-specification)
```

CAIPs/caip-345.md
```diff
@@ -0,0 +1,104 @@
...
+- `wallet_getAssets` - point to EIP
```
bumblefudge, line 26:
```suggestion
- `wallet_getAssets` - Defined in [EIP-7811](https://eip.tools/eip/7811)
```

Note that this is linked to eip.tools instead of eip.ethereum.org, because the PR hasn't been merged yet.  We'll have to update it later, or else point to the open-PR's specific head commit, at time of publishing here if that's still the case.

CAIPs/caip-345.md
bumblefudge: I would think you'd want to scope it to individual chains, right, with a slightly different URL/token/query-params/etc per-chain?

CAIPs/caip-345.md
bumblefudge: I assume you mean this one from an open PR? https://eip.tools/eip/7811

CAIPs/caip-345.md
```diff
@@ -0,0 +1,104 @@
...
+- `wallet_getAssets` - Defined in [EIP-7811](https://eip.tools/eip/7811)
+
+By defining a way for wallets to send requests to an external URL, the requests can be satisfied without needing the wallet app to be open.
+
+## Specification
+<!--The technical specification should describe the standard in detail. The specification should be detailed enough to allow competing, interoperable implementations. -->
+
+Apps SHOULD use the wallet service when available for a method, instead of calling the wallet directly.
+
+Wallets SHOULD implement fallback handling for all wallet service methods.
```
chris13524, line 35: What is the relationship between methods listed in `walletService` and methods available according to CAIP-25? If a method is listed in `walletService` must it also be part of the CAIP-25 session?

Not requiring them to be present in the session could allow wallets to indicate support for a method ONLY on the wallet service, avoiding the wallet potentially needing to forward methods to the wallet service (if apps don't implement the wallet service but do implement the method).

CAIPs/caip-345.md
```diff
@@ -0,0 +1,104 @@
...
+
+Wallets MUST NOT include the same method multiple times or with different (conflicting) wallet service URLs. Apps SHOULD NOT attempt to recover from multiple or conflicting wallet service URLs, but MAY use the first URL available for the method as a convenience for implementation.
+
+```ts
+type WalletService = {
+    url: string,
```
chris13524, line 41: Should this be `uri` instead to account for other non-HTTP protocols? For example a wallet could indicate to use p2p or a light client of some kind in order to handle the method. Furthermore, we could allow listing a wallet service with multiple URLs and the client could select the first one it is able to support/use.

CAIPs/caip-345.md
jakubuid: > For example if a provider pre-validates a method is valid before checking if it should be forwarded to the walletService.

It's not only providers that validate the methods; it's also the dapp/wallet that does that. E.g. when the method is received and is not listed in CAIP-25 it most likely would be defined as invalid.

CAIPs/caip-345.md
jakubuid: Why `sessionProperties`? Can't we limit this to `scopedProperties` only? Where `eip155` stands for the EVM chains. If there's a need to define support per chain, `scopedProperties` also enables that.

bumblefudge: well, the way CAIP-25 was written, `sessionProperties` were meant to be things about the app<>wallet connection that had nothing to do with specific VMs/L1s, like "expiry" (of the connection) for example.  which this kinda feels like to me, actually, unless it's a different URL for each chain?

bumblefudge: so yeah, kinda depends on the API design, all options seem justifiable (as does picking one and excluding the others _for this specific CAIP_)

CAIPs/caip-345.md
```diff
@@ -0,0 +1,104 @@
...
+
+The `walletService` key can be used in both `sessionProperties` and `scopedProperties`, depending on what namespaces the method(s) in question are supported on.
+
+Below is an example support for `wallet_getAssets` which is only supported on `eip155` namespaces:
+
+```json
+"scopedProperties": {
+    "eip155": {
+        "walletService": [{
+            "url": "<wallet service URL>",
```
llbartekll, line 65: what about adding headers?

CAIPs/caip-345.md
bumblefudge: sounds complicated but future-proof.
`uri: [string]` sounds highly decentralized and would save us having to ducktype it as an array five years from now, although you end up having to catch the case of strings not-in-an-array-of-1 because humans 😅

CAIPs/caip-345.md
chris13524: What is an example use case?

llbartekll: Auth token, bundler name for userop status?

chris13524: Auth token can be added as query param IMO. Avoids pre-flight requests on web apps.

chris13524:
TODO outline proxy strategy for dapp CSPs

bumblefudge:
bump @chris13524 is this back-burner or abandoned?

CAIPs/caip-345.md
chris13524: > unless it's a different URL for each chain?

It very much could be. But this entirely depends on the implementation of the wallet service. The thinking was that in a worst-case scenario folks could provide a unique endpoint per-method/chain covering things like authentication or any other properties they may need to fulfill the method (theoretically even providing a hardcoded response as query param). Or in other cases there could be far less diversity. We aren't sure ourselves yet how we will end up doing this, but we'll figure it out.

> sessionProperties were meant to be things about the app<>wallet connection that had nothing to do with specific VMs/L1s, like "expiry"

I didn't realize there was more prescribed semantics to `sessionProperties` other than "it applies to all scopes".

I could see a theoretical `wallet_getCapabilities` if it was a CAIP being valid for all scopes i.e. the session.

If we remove this prop from `sessionProperties` then it would resolve the potential conflict, but it could increase bandwidth/verbosity for methods that are valid for the entire session. We could always prescribe a requirement in this CAIP that would avoid conflicts, or specify a deterministic resolution strategy e.g. scoped first.

CAIPs/caip-345.md
chris13524: Added this rationale in 58cbfec0

CAIPs/caip-345.md
chris13524: In 58cbfec0 I prescribed that wallets must comply with CAIP-25 and provide the same methods as part of the session, as well as requiring that they handle backwards-compat.

CAIPs/caip-345.md
chris13524: Examples added in 58cbfec0

CAIPs/caip-345.md
chris13524: Decided to not generalize this for now since it makes it harder to understand this CAIP and increases implementation complexity. Leaving single URL. If we want more advanced request types in the future, we can make a new CAIP & extend in backwards-compatible way (e.g. via a v2 property).

CAIPs/caip-345.md
chris13524: Use case mentioned in [58cbfec](https://github.com/ChainAgnostic/CAIPs/pull/345/commits/58cbfec03f9ead0cc051e10c5f74e45cfe523430)

CAIPs/caip-345.md
chris13524: Specified in [58cbfec](https://github.com/ChainAgnostic/CAIPs/pull/345/commits/58cbfec03f9ead0cc051e10c5f74e45cfe523430)

CAIPs/caip-345.md
```diff
@@ -0,0 +1,186 @@
...
+Invalid wallet service:
+```json
+{
+    "url": "https://wallet-service.example.com/rpc"
+}
+```
+
+## Security Considerations
+<!--Please add an explicit list of intra-actor assumptions and known risk factors if applicable. Any normative definition of an interface requires these to be implementable; assumptions and risks should be at both individual interaction/use-case scale and systemically, should the interface specified gain ecosystem-namespace adoption. -->
+The wallet service would bear the security resonsibility of responding to these wallet RPC requests. Wallets should make informed decisions about which providers they use for this.
```
obstropolos:
```suggestion
The wallet service would bear the security responsibility of responding to these wallet RPC requests. Wallets should make informed decisions about which providers they use for this.
```

CAIPs/caip-345.md
```diff
@@ -0,0 +1,186 @@
...
+// scoped or session properties
+type Caip25Properties = {
+    caip345: WalletService[],
+    [key: string]: any, // other properties
+};
+```
+
+If a method is included in the `caip345` property, then the same method MUST be included in the CAIP-25 session for the same scopes. Wallets MUST implement fallback handling for all wallet service methods, in case the app does not implement this CAIP.
+
+Below is an unverifed, non-normative, example for `wallet_getAssets` which is only supported in the `eip155` scope:
```
obstropolos:
```suggestion
Below is an unverified, non-normative, example for `wallet_getAssets` which is only supported in the `eip155` scope:
```

CAIPs/caip-345.md
```diff
@@ -0,0 +1,186 @@
...
+
+There must still be 1 canonical wallet service URL for a given method (if available at all).
+
+### Different wallet service depending on account
+
+There was consideration for being able to specify different wallet service URLs for different accounts. However, this would not make sense in the context of CAIP-25 because there is no mechanism to scope the methods to particular accounts. If this CAIP provided a way to scope the methods to particular accounts, the app may still try to send the method requests for non-listed accounts directly to the wallet.
+
+### Map methods to wallet services, instead of methods in array
+
+There was consideraiton for defining the standard to have a separate wallet service URL for every single method. However, this would cause excessive bandwidth consumption if the same URL were to be used for multiple methods which we think is the more likely case.
```
obstropolos:
```suggestion
There was consideration for defining the standard to have a separate wallet service URL for every single method. However, this would cause excessive bandwidth consumption if the same URL were to be used for multiple methods which we think is the more likely case.
```

obstropolos (approved):
Looks fine to me, approving and making the small changes before merging. Fine as a draft.

CAIPs/caip-345.md
```diff
@@ -0,0 +1,186 @@
...
+This proposal defines the wallet service property, `caip345`, for use in CAIP-25. Wallets set this to indicate that certain methods, instead of being handled by the CAIP-25 session, will instead be sent to a JSON-RPC HTTP endpoint. Compatible apps that support this will be able to call these RPC methods, without interactivity with the actual wallet application or user.
+
+## Motivation
+<!--The motivation is critical for CAIP. It should clearly explain why the state of the art is inadequate to address the problem that the CAIP solves. CAIP submissions without sufficient motivation may be rejected outright.-->
+It is sub-optimal UX to redirect the user to their wallet in order to handle RPC requests that are non-actionable to them. This is especially relevant for protocols such as WalletConnect which is used in more distributed environments such as mobile wallets or custodial solutions. In these contexts, actioning a wallet RPC request can involve significant effort.
+
+Examples of non-actionable wallet requests include:
+- [EIP-5792](https://eips.ethereum.org/EIPS/eip-5792#wallet_getcapabilities) `wallet_getCapabilities`
```
bumblefudge:
```suggestion
- [EIP-5792 `wallet_getCapabilities`](https://eips.ethereum.org/EIPS/eip-5792#wallet_getcapabilities)
```

CAIPs/caip-345.md
```diff
@@ -0,0 +1,186 @@
...
+- [EIP-5792](https://eips.ethereum.org/EIPS/eip-5792#wallet_getcallsstatus-rpc-specification) `wallet_getCallsStatus`
```
bumblefudge:
```suggestion
- [EIP-5792 `wallet_getCallsStatus`](https://eips.ethereum.org/EIPS/eip-5792#wallet_getcallsstatus-rpc-specification)
```

CAIPs/caip-345.md
```diff
@@ -0,0 +1,186 @@
...
+
+By defining a way for wallets to send requests to an out-of-band endpoint, the requests can be satisfied without needing the wallet app to be open.
+
+## Specification
+
+### Wallet Service
+
+A "wallet service" is a JSON-RPC-compatible HTTP endpoint that can be used to satisfy certain wallet RPC methods. This service may be developed, hosted, and maintained by the same organization developing the wallet app, or by a third-party. It is up to the wallet to determine what server should be responsible for handling wallet RPCs.
+
+The wallet service can be specified as a URL, and a list of methods for which to use the URL. `methods` SHOULD NOT be empty. The endpoint MUST be JSON-RPC compatible and support `POST` requests.
```
bumblefudge:
```suggestion
The wallet service can be specified as a URL, and a list of JSON-RPC methods for which to use the URL. `methods` SHOULD NOT be empty. The endpoint MUST be JSON-RPC compatible and support `POST` requests.
```

CAIPs/caip-345.md
```diff
@@ -0,0 +1,186 @@
...
+
+```ts
+// scoped or session properties
+type Caip25Properties = {
+    caip345: WalletService[],
+    [key: string]: any, // other properties
+};
+```
+
+If a method is included in the `caip345` property, then the same method MUST be included in the CAIP-25 session for the same scopes. Wallets MUST implement fallback handling for all wallet service methods, in case the app does not implement this CAIP.
```
bumblefudge, line 90: @chris13524 Wait, I'm confused here. 

1. what should the wallet do if a method IS listed in caip345 but not in the caip25 request?
1. what should the dapp do if the method is listed in both but not authorized in the returned caip25 object?
1. fallback handling for what, exactly? how to handle the RPC method at that endpoint, or for communicating the endpoint to the dapp if they can't use caip-25 to do it?

(all three of these could be addressed in a future PR if you want this merged as draft ASAP)

CAIPs/caip-345.md
```diff
@@ -0,0 +1,186 @@
...
+
+Many web apps specify `connect-src` in their CSP which prevents the app and its libraries from connecting to URLs that aren't pre-specified. Since each wallet may use their own wallet service hosted on unique origins, it's not possible/advisable to dynamically set the `connect-src` value as-necessary.
+
+However to support this use case, such apps can consume a minimal proxy service (such as a server function) which will forward the wallet RPC request to the destination wallet service. This proxy service is known by the app, and has a fixed origin URL, which allows placing it in `connect-src`. The proxy service could be implemented by the app, or by a third-party.
+
+The mechanism by which this proxy service is implemented or consumed is outside the scope of this CAIP.
+
+## Test Cases
+<!--Please add diverse test cases here if applicable. Any normative definition of an interface requires test cases to be implementable. -->
+Valid wallet service:
```
bumblefudge, line 150:
```suggestion

Valid wallet service:
```

CAIPs/caip-345.md
```diff
@@ -0,0 +1,187 @@
...
+
+Valid wallet service:
+```json
+{
+    "methods": ["wallet_prepareCalls", "wallet_sendPreparedCalls"],
+    "url": "https://wallet-service.example.com/rpc"
+}
+```
+
+Invalid wallet service:
```
bumblefudge, line 158:
```suggestion
Invalid wallet services:

```json
{
    "methods": ["wallet_prepareCalls", "wallet_sendPreparedCalls"],
    "url": ["https://wallet-service.example.com/rpc1", "https://wallet-service.example.com/rpc2"]
}
```

```

File: CAIPs/caip-345.md [hunk 21]
```diff
@@ -0,0 +1,187 @@
...
+
+## Privacy Considerations
+<!--Please add an explicit list of intra-actor assumptions and known risk factors if applicable. Any normative definition of an interface requires these to be implementable; assumptions and risks should be at both individual interaction/use-case scale and systemically, should the interface specified gain ecosystem-namespace adoption. -->
+Similarly, the wallet service would have visibility into the wallet RPC requests being sent to it.
+
+## Backwards Compatibility
+<!--All CAIPs that introduce backwards incompatibilities must include a section describing these incompatibilities and their severity. The CAIP must explain how the author proposes to deal with these incompatibilities. CAIP submissions without a sufficient backwards compatibility treatise may be rejected outright.-->
+This new property is fully backwards-compatible with CAIP-25.
+
+As also mentioned above, wallets MUST implement fallback handling for all wallet service methods, in case the app does not implement this CAIP.
```
bumblefudge, line 177: ... but does implement CAIP-25?

PR #341, opened 2024-12-12
CAIP-341 - Extension ID Target Type Specification

ffmcgee725:
Defining extension id as a valid target type for wallet data.

PR #348, opened 2025-04-10
Browser extension provider communication

jiexi:
This CAIP discusses the motivation, specification, and rationale for a proposal aimed at improving how web extension wallets interact with websites. It outlines the current method of injecting JavaScript provider APIs into websites, its advantages, and its numerous disadvantages, such as security concerns, performance issues, and the risk of breaking websites. An alternative strategy is proposed that specifies a standard communication specification over a new transport layer which enables websites to be able to embed their own provider as a library, addressing the disadvantages of injecting providers into websites and improving web extension interoperability as a whole.


jiexi:
WIP. I still need to address comments in the old PR [here](https://github.com/jiexi/CAIPs/pull/1)

caip-browser_extension_provider_communication.md
```diff
@@ -0,0 +1,139 @@
...
+
+Web extension inter-operability is the key to enabling generalized provider implementations. Generalized provider implementations allows for convenient adoption by Dapps, leading to more wide spread adoption of the standard as a result.
+
+## Test Cases
+<!--Please add test cases here if applicable.-->
+
+## Security Considerations
+<!--Please add an explicit list of intra-actor assumptions and known risk factors if applicable. Any normative definition of an interface requires these to be implementable; assumptions and risks should be at both individual interaction/use-case scale and systemically, should the interface specified gain ecosystem-namespace adoption. -->
+`externally_connectable` has seen a decade of usage via extensions on Chrome. It has a strictly better security when compared to postMessage over contentscript.
```
bumblefudge, line 121:
```suggestion
`externally_connectable` has seen a decade of usage via extensions on Chrome. It has strictly better security properties when compared to `postMessage` over content scripts.
```

bumblefudge:
@jiexi Should this be merged into #341 ? It seems like they're complementary, to some degree; if you want to keep them as two distinct documents, maybe this one should be informational and that one should be normative (might require moving some stuff over there)

caip-browser_extension_provider_communication.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Rationale
+<!--The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages. The rationale may also provide evidence of consensus within the community, and should discuss important objections or concerns raised during discussion.-->
+While web extensions could realize the benefits of `externally_connectable` without using it with a standardized interface, this would mean that every web extension would have to ship a provider library that worked specifically for their interface and that Dapps would need to import them. It wouldn't be feasible for Dapps to import every single web extension's specific provider implementation, making this approach less viable.
+
+Web extension inter-operability is the key to enabling generalized provider implementations. Generalized provider implementations allows for convenient adoption by Dapps, leading to more wide spread adoption of the standard as a result.
+
```
bumblefudge, line 119: reminder: maybe mention other uses of the transport being possible?

caip-browser_extension_provider_communication.md
```diff
@@ -0,0 +1,145 @@
+---
+# Every document starts with a front matter in YAML enclosed by triple dashes.
+# See https://jekyllrb.com/docs/front-matter/ to learn more about this concept.
+caip: <to be assigned>
+title: Web extension provider communication
+author: Mark Stacey (@Gudahtt), Jiexi Luan (@jiexi)
+discussions-to: <URL>
+status: Draft
+type: Standard
```
bumblefudge, line 9:
```suggestion
type: Informational
```
It might be better to think of this as the BCP explaining, contextualizing, and providing an "if you don't care about the rest of the transports/options" entrance for CAIP-282.  Otherwise it's kind of confusing that they're both normative and cover the same API 😄

adonesky1: So partly I agree, but this specifies an envelope/message format that is expected to wrap the CAIP messages over the `externally_connectable` transport. Its possible it should be folded into CAIP-341 instead?

pedrouid:
Shouldnt this be closed in favor of the recently merged #282

PR #350, opened 2025-04-18
New CAIP profile: interoperable address binary specification

0xteddybear:


CAIPs/caip-000.md
```diff
@@ -0,0 +1,121 @@
...
+---
+
+## Simple Summary
+A CAIP profile for every chain namespace to describe how to serialize their addresses and chain references into the Interoperable Address format described in [ERC-7930].
+
+## Abstract
+<!--A short (~200 word) description of the technical issue being addressed.-->
+Every chain defines their own address types & also a way to identify networks within that chain's ecosystem.
+[ERC-7930] is an address format for (address, chain) pairs supporting any chain, and also including metadata on the chain & addresses' type so no information needs to be communicated out of band.
+This CAIP namespace aims to be a living & mutable repository of serialization methods for all chains to uniformly conform to Interoperable Addresses.
```
bumblefudge:
```suggestion
This CAIP aims to append a living & mutable repository of chain-specific address serialization methods to the CASA system in the form of its namespace-specific profiles, enabling all current and future chains to uniformly conform to Interoperable Addresses.
```

CAIPs/caip-000.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Motivation
+Standards like CAIP-2 and CAIP-10 are simple text representations of addresses & chain namespaces/references, but they do not address:
+- Binary representation of addresses and/or chain namespace+reference: Relevant & desired for on-chain usage.
+- Canonicity: CAIP-10, being a text format which leaves serialization to actual on-chain addresses undefined, makes no guarantees on a blockchain account (_target address_ in ERC-7930 parlance) having only one CAIP-10 representation, which makes them less useful as dictionary keys.
```
bumblefudge:
```suggestion
- Canonicity: CAIP-10, being a generic text format, leaves the canonization and serialization of ASCII address to on-chain addresses to each namespace's profile; unless a given namespace's profile specifies such logic, the standard itself makes no universal guarantees on a blockchain account (_target address_ in ERC-7930 parlance) having only one CAIP-10 representation, which makes them less useful as dictionary keys.
```
For the record, this canonicity issue HAS come up, but CAIP-10 was already final at the time and it was deemed unnecessary for EIP155 because EIP55 is pretty definitive and all known implementations canonicalize to checksum-case for deduplication purposes. See the [Syntax section of the EIP155 profile of CAIP-10](https://namespaces.chainagnostic.org/eip155/caip10#syntax) for more detail.

CAIPs/caip-000.md
```diff
@@ -0,0 +1,121 @@
...
+- Completeness: both formats have limits on the length of their identifiers, which are reasonable for human-readability but often identifiers have to drop meaningful information to conform to those length requirements. While that information might be easy to look up in the context of wallet software, doing so within a smart contract would not be possible.
+- Succinctness: Text formats have an inherent overhead over binary ones when measured for information efficiency in a machine context.
```
bumblefudge:
```suggestion
- Succinctness: Text formats necessarily have incur encoding "overhead" compared to binary ones,  i.e. relative informational inefficiency.
```

CAIPs/caip-350.md
```diff
@@ -0,0 +1,121 @@
...
+- Living standards are not a usual thing in the Ethereum ecosystem: the only other Living ERC is [EIP-1]
+- It would be desireable to finalize the definition of the format itself while allowing the specification of every chain's serialization to be defined as needed.
+- It would have placed all the editorial (and presumably most of the specification) work on the [ERC-7930] authors, while the CASA paradigm allows better distribution of work by leaving the definition of the profile for every chain namespace to the parties of its ecosystem interested in using Interoperable Addresses.
+
+## Test Cases
+Not applicable since this does not define a normative interface, instead being a meta-specification of normative interfaces.
+
+## Security Considerations
+It is possible that a CAIP-000 profile for a given chain namespace is not able to guarantee requirements of canonicity which are a desired property of Interoperable Addresses due to characteristics of the chain namespace, and systems relying on that canonicity may exhibit unexpected behavior as a result. When this is the case, it should be noted in the `Extra Considerations` section.
```
bumblefudge, line 104:
```suggestion
It is possible that a CAIP-350 profile for a given chain namespace is not able to guarantee requirements of canonicity which are a desired property of Interoperable Addresses due to characteristics of the chain namespace, and systems relying on that canonicity may exhibit unexpected behavior as a result. When this is the case, it should be noted in the `Extra Considerations` section.
```

i wish we had written a section like this for CAIP-10 before it went to final 🙈

jiexi:
Thoughts on splitting this into separate CAIPs, one for the binary representation of CAIP-2 and one for CAIP-10? If it's one CAIP as it is now, then the only way for me to reference this is "the CAIP-350 binary version of CAIP-2" or "the CAIP-350 binary version of CAIP-10". I think that may look awkward in code at minimum.

u59149403:
We already have CAIP for binary addresses: CAIP-50.

Also, I don't see any point in binary ids at all. We can just use string id (i. e. CAIP-10) as binary. We just need to agree on case canonicalization. See my thoughts on case canonicalization here: https://github.com/ChainAgnostic/CAIPs/issues/351

u59149403:
Also, why we need text representation of CAIP-350? We already have CAIP-10. If you want to add checksum, just add checksum to CAIP-10 address.

Also, why checksum is so long? The only reason why we need checksum is for cases when people type address manually. (This is why bank cards have 1 digit checksum.) For such cases checksum should be as short as possible

CAIPs/caip-350.md
```diff
@@ -0,0 +1,120 @@
...
+requires: 2
+---
+
+## Simple Summary
+A CAIP profile for every chain namespace to describe how to serialize their addresses and chain references into the Interoperable Address format described in [ERC-7930].
+
+## Abstract
+Every chain defines their own address types & also a way to identify networks within that chain's ecosystem.
+[ERC-7930] is an address format for (address, chain) pairs supporting any chain, and also including metadata on the chain & addresses' type so no information needs to be communicated out of band.
+This CAIP namespace aims to be a living & mutable repository of serialization methods for all chains to uniformly conform to Interoperable Addresses.
```
u59149403, line 18: In CAIP context, the word "namespace" has different meaning. It means namespaces such as EIP-155, BIP-122, etc. See https://namespaces.chainagnostic.org/

CAIPs/caip-350.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Motivation
+Standards like CAIP-2 and CAIP-10 are simple text representations of addresses & chain namespaces/references, but they do not address:
+- Binary representation of addresses and/or chain namespace+reference: Relevant & desired for on-chain usage.
+- Canonicity: CAIP-10, being a generic text format, leaves the canonization and serialization of ASCII address to on-chain addresses to each namespace's profile; unless a given namespace's profile specifies such logic, the standard itself makes no universal guarantees on a blockchain account (_target address_ in ERC-7930 parlance) having only one CAIP-10 representation, which makes them less useful as dictionary keys.
```
u59149403, line 23: This means we should fix CAIP-10 instead of introducing new standard

u59149403:
Also, a lot of CAIP-10 implementations for particular namespaces (i. e. CAIP-10 column at https://namespaces.chainagnostic.org/ ) have "Draft" status. (I didn't check, I think they all have "Draft" status.) So, we can still fix them and add canonicalization. Also, CAIP-50 has "Draft" status, too, so we can change it, too.

(Well, I think we are allowed to change all CAIPs, even final ones, because, as well as I understand, they are not adopted widely anyway yet.)

So, again: don't create yet another binary representation. Just contribute to already existing CAIP-50. Of course, we should add column, corresponding to CAIP-50, to https://namespaces.chainagnostic.org/ .

And, of course, don't create new text representation. We just need to use CAIP-10. CAIP-50 should get its own column at https://namespaces.chainagnostic.org/ , and it should describe conversion between CAIP-10 and CAIP-50.

And if you want checksum, just add it to CAIP-10. Yes, it has final status, but I personally think that we still can change it. Or, alternatively, just specify in ERC-7930 that address should have form `CAIP-10 # checksum`

u59149403:
Also: 4 byte checksum is too big to be typed manually (I already said this), but in the same time too small to be cryptographically secure. I. e. 4 bytes is not enough to be sure that addresses equal based just on equality of checksums alone. I. e. malicious agent can easily craft address with same checksum.

So, again: consider using 1 byte checksum or even 4 bit checksum. (Bank cards use 1 digit checksum.)

And if your standard is not designed for users to type addresses manually, then don't add checksum at all

CAIPs/caip-350.md
0xteddybear: you're right, I meant 'profile'. Will edit this, thanks!

0xteddybear:
> Thoughts on splitting this into separate CAIPs, one for the binary representation of CAIP-2 and one for CAIP-10? 

I see the point of this CAIP to be the algorithms for converting between the two representations and the namespace-specific ones, and from that point of view it makes sense to have specifications for the input and output of the algorithms live in the same place as the algorithms themselves.

For context, I imagined this CAIP as the one-stop-shop for developers trying to integrate with ERC-7930, and tried to make the life of said hypothetical developer as easy as possible. Although the developer is not hypothetical anymore, wdyt of this @0xjagger?

> If it's one CAIP as it is now, then the only way for me to reference this is "the CAIP-350 binary version of CAIP-2" or "the CAIP-350 binary version of CAIP-10". I think that may look awkward in code at minimum.

Fair point. I'll try to come up with better names and define them normatively in the CAIP profile, would that work?

> We already have CAIP for binary addresses: CAIP-50.

I gave CAIP-50 a look, and it struck me as:
- stagnant: no updates since 2020.
- incomplete: attempted to define namespaces' implementation inside the CAIP profile itself, instead of leaving it up to the different namespaces, and did not address the length limitations of CAIP-2 references, which can be a problem with standards such as ERC-7785

But otherwise it was a good inspiration. It would be a good idea to reference it as such.
The line on whether to re-write a standard or start a new one had to be drawn somewhere, and in my case I drew it somewhere in between ERC-7828 and CAIP-50 :grin:

> Also, I don't see any point in binary ids at all. We can just use string id (i. e. CAIP-10) as binary. We just need to agree on case canonicalization.

That would be very inefficient for on-chain actors, having to use *at least* double the memory/storage.

> Also, why we need text representation of CAIP-350?

- Avoiding the length limitations imposed on CAIP-2 (and to a lesser extent CAIP-10) outlined above
- Defining a grammar that's friendly to hierarchical names (e.g allowing for dots on the chain reference), which could be a stepping stone towards something line ERC-7828

> Also, why checksum is so long?

Mainly I wanted to make the dreams of EIP-55 authors come true: https://eips.ethereum.org/EIPS/eip-55
('This is a ~50x improvement over ICAP, but not as good as a 4-byte check code.', in Rationale section)

> The only reason why we need checksum is for cases when people type address manually

That is not entirely true, see my comment on the 7828 PR https://github.com/ethereum/ERCs/pull/735#discussion_r2067394145

thanks for your feedback @u59149403 @jiexi !, will make sure to tag you both when I implement your suggestions :grin:

u59149403:
@0xteddybear

> * Avoiding the length limitations imposed on CAIP-2 (and to a lesser extent CAIP-10) outlined above
> * Defining a grammar that's friendly to hierarchical names (e.g allowing for dots on the chain reference), which could be a stepping stone towards something line ERC-7828

Both these things should be fixed in CAIP-2.
- I think it is okay to make maximal length for CAIP-2 bigger, despite CAIP-2 is final. Because it is not implemented widely anyway
- We can surely add hierarchical names to CAIP-2 using namespace-specific versions of CAIP-2. Most of them are not final, and we can add new ones

CAIPs/caip-350.md
bumblefudge: i'm not sure it's fixable in general, given how divergent addressing and account models are across namespaces; each namespace needs to be profiled down and specify its own canonicity behaviors, in my opinion.

bumblefudge:
> Also, a lot of CAIP-10 implementations for particular namespaces (i. e. CAIP-10 column at https://namespaces.chainagnostic.org/ ) have "Draft" status. (I didn't check, I think they all have "Draft" status.) So, we can still fix them and add canonicalization. Also, CAIP-50 has "Draft" status, too, so we can change it, too.

Namespace profiles are never final by design! They are also technically non-normative, fwiw, in IETF terms.

CAIP-50 was never officially marked as final because there are no known production deployments. Which brings me to...

> (Well, I think we are allowed to change all CAIPs, even final ones, because, as well as I understand, they are not adopted widely anyway yet.)

CAIP-2 _IS_ final, because there are many prod deployments that would break if changed.

I would rather CAIP-2 were extended by a new (ASCII-only) CAIP, personally, because introducing hierarchical strings of CAIP-2s would break some implementations that do not handle this case.  

I have been tracking this "L3" issue for a while, as well as the "onchain alternatives to ethereum-lists/chains" issue, and there also exists a more "nuclear option", which would be creating a distinct namespace if the chainId system assumptions and guarantees change drastically enough.  `eip155` is literally the namespace of EVM chains that you connect to directly once you have an RPC endpoint and a chainId-- `evm` or `eth` could be an overlapping namespace (with most of the same chains in it) that assumes different addressing for chains, and allows you to recursively dereference ephemeral or secret chains from onchain data.

Same with the length requirement-- a longer alternative to CAIP-2 would be a breaking upgrade/"major version" for most existing implementations, so I would rather it were a new CAIP#, and implementations that don't make that upgrade can still be in conformance with a final standard.

@pedrouid @vandan may have thoughts on "breaking CAIP-2" 😄

0xJagger:
> For context, I imagined this CAIP as the one-stop-shop for developers trying to integrate with ERC-7930, and tried to make the life of said hypothetical developer as easy as possible. Although the developer is not hypothetical anymore, wdyt of this @0xJagger?

Yes, we're considering developing an interop wallet that supports both CAIP-10 and CAIP-2. I believe having a single CAIP that includes both algorithms will be more developer-friendly.

From the perspective of a new namespace looking to adopt the standard, I think it's simpler to implement one unified specification that includes both algorithms

0xteddybear:
@jiexi pushed a new commit referencing 'Interoperable Addresses' and 'Interoperable Names', and also updated ERC-7930 to introduce the two concepts. I believe it should be easier to refer to those as: 

    Interoperable
    ├── Name
    │   ├── Chain
    │   └── Address
    └── Address
        ├── Chain
        └── Address

CAIPs/caip-350.md
0xteddybear: update: this standard already defines the 'Extra Considerations' sections to deal with this, but added the possibility of address/chain collisions as a 'Security Consideration' of ERC-7930

CAIPs/caip-350.md
0xteddybear: addressed!

CAIPs/caip-350.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Abstract
+Every chain defines their own address types & also a way to identify networks within that chain's ecosystem.
+[ERC-7930] is an address format for (address, chain) pairs supporting any chain, and also including metadata on the chain & addresses' type so no information needs to be communicated out of band.
+This CAIP profile aims to be a living & mutable repository of serialization methods for all chains to uniformly conform to Interoperable Addresses.
+
+## Motivation
+Standards like CAIP-2 and CAIP-10 are simple text representations of addresses & chain namespaces/references, but they do not address:
+- Binary representation of addresses and/or chain namespace+reference: Relevant & desired for on-chain usage.
+- Canonicity: CAIP-10, being a generic text format, leaves the canonization and serialization of ASCII address to on-chain addresses to each namespace's profile; unless a given namespace's profile specifies such logic, the standard itself makes no universal guarantees on a blockchain account (_target address_ in ERC-7930 parlance) having only one CAIP-10 representation, which makes them less useful as dictionary keys.
```
bumblefudge, line 23:
```suggestion
- Canonicity: CAIP-10, being a generic text format, leaves the canonization and serialization of ASCII address to on-chain addresses to each namespace's profile; unless a given namespace's profile specifies such logic, the standard itself makes no universal guarantees on a blockchain account (_target address_ in ERC-7930 parlance) having only one CAIP-10 representation; for namespaces where such canonicity is neither inherent or specified by the CASA profile, duplicate entries can be created (e.g. when used as dictionary keys).
```

CAIPs/caip-350.md
```diff
@@ -0,0 +1,120 @@
...
+- Completeness: both formats have limits on the length of their identifiers, which are reasonable for human-readability but often identifiers have to drop meaningful information to conform to those length requirements. While that information might be easy to look up in the context of wallet software, doing so within a smart contract would not be possible.
+- Succinctness: Text formats necessarily have to incur encoding overhead compared to binary ones, causing relative informational inefficiency.
+
+## Specification
+The purpose of every supporting profile is be to be able to deterministically and unambiguously convert between:
+- Customary address text formats of the ecosystem (which may also be described in its [CAIP-10] profile) and the Interoperable Address text representation defined in [ERC-7930]
```
bumblefudge, line 29: I think it would be more precise to refer to all these as "ASCII" formats than "text", since AFAIK all of them are the latter and the former could be confused with, e.g. "seed phrases" or more human-readable/speakable formats.

CAIPs/caip-350.md
```diff
@@ -0,0 +1,120 @@
...
+- Binary representation of addresses and/or chain namespace+reference: Relevant & desired for on-chain usage.
+- Canonicity: CAIP-10, being a generic text format, leaves the canonization and serialization of ASCII address to on-chain addresses to each namespace's profile; unless a given namespace's profile specifies such logic, the standard itself makes no universal guarantees on a blockchain account (_target address_ in ERC-7930 parlance) having only one CAIP-10 representation, which makes them less useful as dictionary keys.
+- Completeness: both formats have limits on the length of their identifiers, which are reasonable for human-readability but often identifiers have to drop meaningful information to conform to those length requirements. While that information might be easy to look up in the context of wallet software, doing so within a smart contract would not be possible.
+- Succinctness: Text formats necessarily have to incur encoding overhead compared to binary ones, causing relative informational inefficiency.
+
+## Specification
+The purpose of every supporting profile is be to be able to deterministically and unambiguously convert between:
```
bumblefudge, line 28:
```suggestion
The purpose of each namespace's profile is to specify deterministic and unambiguous conversions between the following format-pairs:
```

CAIPs/caip-350.md
```diff
@@ -0,0 +1,120 @@
...
+- Customary address text formats of the ecosystem (which may also be described in its [CAIP-10] profile) and the Interoperable Address text representation defined in [ERC-7930]
+- Customary chain reference text formats of the ecosystem (which may also be described in [CAIP-2]) and the text representation defined in [ERC-7930]
+- [ERC-7930]'s Interoperable Address -> [ERC-7930]'s Interoperable Name and back of the ecosystem's chain references
```
bumblefudge, line 31:
```suggestion
- From [ERC-7930]'s Interoperable Address -> [ERC-7930]'s Interoperable Name to the ecosystem's chain references and back
```

bumblefudge: the others are just pairs, but this specifies "to and back" -- is there a different in the expectation of losslessness?

CAIPs/caip-350.md
```diff
@@ -0,0 +1,120 @@
...
+- [ERC-7930]'s Interoperable Address -> [ERC-7930]'s Interoperable Name and back of the ecosystem's addresses *for all address formats of said ecosystem*
```
bumblefudge, line 32:
```suggestion
- [ERC-7930]'s Interoperable Address -> [ERC-7930]'s Interoperable Name to *all address formats of said ecosystem* and back
```

CAIPs/caip-350.md
```diff
@@ -0,0 +1,120 @@
...
+
+And must also define the binary id of the namespace itself, akin to [ERC-7930]'s `ChainType`
+
+For this, every namespace MUST define the following sections in their definition of their CAIP-350 profile:
```
bumblefudge, line 36:
```suggestion
For this, every namespace MUST define specify all of the above in a CAIP-350 profile to maximize interoperability and review. A [template for these profiles](https://github.com/ChainAgnostic/namespaces/blob/main/_template/caip350.md) is defined in the Namespaces registry.
```

And move the template to the URL above by opening a separate PR [over in namespaces](https://github.com/ChainAgnostic/namespaces/pulls)

bumblefudge (approved):
Approved modulo the most recent change requests!

CAIPs/caip-350.md
0xteddybear: I'm not sure the definition of text formats (such as CAIP-10, and EIP-55 to a lesser extent since it mentions ascii/utf8 but I'm not sure how normative it intends to be about it) specify the text encoding to be used, only that the text is meant to contain such and such latin/punctuation _characters_). Since they are meant to be displayed to humans, the underlying encoding could be considered an implementation detail, and describing them as ASCII could potentially be incorrect.

I do think the document could benefit from clearly differentiating between resolved (potentially ERC-7828 Interoperable Names) and un-resolved (ERC-7930 Interoperable Names) names. Do you think there are better words for those concepts though? I can't seem to be able to come up with something both clear and concise

CAIPs/caip-350.md
0xteddybear: no, I wrote it like that because ...I thought it was cleaner natural language? but losslessness expectations are the same. Rewrote to match the items above

bumblefudge:
> So I propose to mandate that set of possible CAIP-350 text representations for accounts is subset of set of possible CAIP-10s. I. e. every valid text CAIP-350 address should be valid CAIP-10. Similarly, every valid CAIP-350 text representation of chain should be valid CAIP-2. This will allow us to have canonicity and in the same time to not introduce yet another totally different standard.

I like this idea of backwards-compat with CAIP-2 and CAIP-10 a lot for the _textual_ versions, if this is possible (am I missing a corner-case where it isn't?).  If CAIP-10 allows multiple representations of a given address in a given namespace, A.) a profile can constrain them, and B.) CAIP-350 can MANDATE usage of such profiles, or otherwise mandate a subset of possible CAIP-10s (ideally a single canonical one) be used.

CAIPs/caip-350.md
```diff
@@ -0,0 +1,65 @@
...
+author: Defi Wonderland (@defi-wonderland), Teddy (@0xteddybear), Joxes (@Joxess), Racu (@0xRacoon), Skeletor Spaceman (@0xskeletor-spaceman), TiTi (@0xtiti), Gori (@0xGorilla), Ardy (@0xArdy), Onizuka (@onizuka-wl)
+discussions-to: https://ethereum-magicians.org/t/erc-7930-interoperable-addresses/23365
+status: Draft
+type: Standard
+created: 2025-04-18
+requires: 2
+---
+
+## Simple Summary
+A CAIP profile for every chain namespace to describe how to serialize their addresses and chain references into the Interoperable Address format described in [ERC-7930].
```
bumblefudge, line 13:
```suggestion
This standard, when profiled down for each namespace, describes a unified and general purpose method for serializing chain-specified addresses into the Interoperable Address format specified canonically in [ERC-7930].
```

CAIPs/caip-350.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Abstract
+Every chain defines their own address types & also a way to identify networks within that chain's ecosystem.
```
bumblefudge, line 16:
```suggestion
Every namespaces defines their own address types & also a way to identify networks within that namespace's ecosystem.
```

CAIPs/caip-350.md
```diff
@@ -0,0 +1,65 @@
...
+[ERC-7930] is an address format for (address, chain) pairs supporting any chain, and also including metadata on the chain & addresses' type so no information needs to be communicated out of band.
```
bumblefudge, line 17:
```suggestion
[ERC-7930] is an address format for (address, chain) pairs supporting any chain, and also including metadata on the chain & addresses' type in a self-describing way, reducing the need to pass out-of-band metadata or context.
```

CAIPs/caip-350.md
```diff
@@ -0,0 +1,65 @@
...
+This CAIP aims to append a living & mutable repository of chain-specific address serialization methods to the CASA system in the form of its namespace-specific profiles, enabling all current and future chains to uniformly conform to Interoperable Addresses.
+
+## Motivation
+Standards like CAIP-2 and CAIP-10 are simple text representations of addresses & chain namespaces/references, but they do not address:
+- Binary representation of addresses and/or chain namespace+reference: Relevant & desired for on-chain usage.
+- Canonicity: CAIP-10, being a generic text format, leaves the canonization and serialization of text address to on-chain addresses to each namespace's profile; unless a given namespace's profile specifies such logic, the standard itself makes no universal guarantees on a blockchain account (_target address_ in ERC-7930 parlance) having only one CAIP-10 representation; for namespaces where such canonicity is neither inherent or specified by the CASA profile, duplicate entries can be created (e.g. when used as dictionary keys).
```
bumblefudge, line 23:
```suggestion
- Canonicity: CAIP-10, being a generic text format, leaves the canonization and serialization of text address to on-chain addresses to each namespace's profile. Unless a given namespace's profile specifies such logic, the standard itself makes no universal guarantees on a blockchain account (_target address_ in ERC-7930 parlance) having only one CAIP-10 representation. For namespaces where such canonicity is neither inherent nor specified by the CASA profile, duplicate entries can be created (e.g. when used as dictionary keys).
```

CAIPs/caip-350.md
```diff
@@ -0,0 +1,65 @@
...
+For this, every namespace MUST define specify all of the above in a CAIP-350 profile to maximize interoperability and review. A [template for these profiles](https://github.com/ChainAgnostic/namespaces/blob/main/_template/caip350.md) is defined in the Namespaces registry.
+
+## Rationale
+The main alternative to this standard would've been to define all formats & conversions in [ERC-7930], turning it into a Living ERC, which was not ideal since:
+- Living standards are not a usual thing in the Ethereum ecosystem: the only other Living ERC is EIP-1
+- It would be desireable to finalize the definition of the format itself while allowing the specification of every chain's serialization to be defined as needed.
+- It would have placed all the editorial (and presumably most of the specification) work on the [ERC-7930] authors, while the CASA paradigm allows better distribution of work by leaving the definition of the profile for every chain namespace to the parties of its ecosystem interested in using Interoperable Addresses.
+
+## Test Cases
+Not applicable since this does not define a normative interface, instead being a meta-specification of normative interfaces.
```
bumblefudge, line 45:
```suggestion
Not applicable since this does not define a normative interface, instead being a meta-specification of normative interfaces.
Test cases SHOULD be included in each namespace profile.
```

CAIPs/caip-350.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Security Considerations
+It is possible that a CAIP-350 profile for a given chain namespace is not able to guarantee requirements of canonicity which are a desired property of Interoperable Addresses due to characteristics of the chain namespace, and systems relying on that canonicity may exhibit unexpected behavior as a result. When this is the case, it should be noted in the `Extra Considerations` section.
+
+## Backwards Compatibility
+This standard actively seeks to be exhaustive in its backwards compatibility with CAIP-2 and CAIP-10.
```
bumblefudge, line 51: I'm still not clear on how this works with Completeness, and whether a CAIP-350 implementation is allowed to produce or required to accept a CAIP-2 or CAIP-10 in cases where these truncate the information preserved in binary or in the novel format!

bumblefudge (changes requested):
G2G in "Draft" status, modulo my suggestions, although I'd still like to see a "corner-case" example in at least one profile where address and/or chainId is truncated by CAIP-2, this part is still a little fuzzy for me.  Other minor things:

1. Could you put a space after each Markdown header? Most markdown linters do this by default.
2. In the profiles and profile template, could you add a test vector section (could just copy-paste the examples, but include all variants, and ideally a truncation corner-case example)

bumblefudge:
@obstropolos @ligi does either of you have time to give this a glance before next editorial meeting? I think OP is eager to get CAIP number reserved in "Draft" status so that they can refer to it from inside the EIP. I'm fine for the draft to be a little looseygoosey on the truncation corner-case, as long as it gets fixed eventually (e.g. in at least one profile to serve as template for future namespace profiles)

0xteddybear:
> I like this idea of backwards-compat with CAIP-2 and CAIP-10 a lot for the textual versions, if this is possible (am I missing a corner-case where it isn't?). If CAIP-10 allows multiple representations of a given address in a given namespace, A.) a profile can constrain them, and B.) CAIP-350 can MANDATE usage of such profiles, or otherwise mandate a subset of possible CAIP-10s (ideally a single canonical one) be used.

CAIP-10 address compatibility is possible if the namespaces define their 10 and 350 profiles in a consistent way, and that is very desireable, but there's no way to normatively require it in this standard as far as I can think of.

The CAIP-2 part, however, is a bit more complex (see below)

> More fundamentally, though, it's not up to me-- if CAIP-350 is valid/implementable, it's up to OP whether backwards compatibility is worth it or not

Full disclosure on this: I'm choosing to *not* preserve backwards compatibility with CAIP-2, in order to provide first-class support for 256-bit EVM chainids, so they can be the output of a hash function, so the Ethereum ecosystem can deprecate [https://github.com/ethereum-lists/chains] and replace it with a decentralized solution outlined in [ERC-7785](https://eips.ethereum.org/EIPS/eip-7785) (and other ecosystems can do similar things!).

The alternatives I can think of to maintain backwards compatibility with CAIP-2 are either technically complex in a way that can cause problems for users and developers or isn't well abstracted from them (by adding ambiguities on the chain spec, e.g. how to tell apart a genuinely short chainid from one that has been truncated to comply with CAIP-2), or don't hold the required property of the previous paragraph.

CAIPs/caip-350.md
```diff
@@ -0,0 +1,81 @@
+---
+caip: 350
+title: Binary serialization of blockchain IDs and addresses
```
obstropolos:
```suggestion
title: Binary Serialization of Blockchain IDs and Addresses
```

CAIPs/caip-350.md
```diff
@@ -0,0 +1,81 @@
...
+requires: 2
+---
+
+## Simple Summary
+
+This standard, when profiled down for each namespace, describes a unified and general purpose method for serializing chain-specified addresses into the Interoperable Address format specified in [ERC-7930].
+
+## Abstract
+
+Every namespace defines their own address types & also a way to identify networks within that namespace's ecosystem.
```
obstropolos:
```suggestion
Every namespace defines their own address types and also a way to identify networks within that namespace's ecosystem.
```

CAIPs/caip-350.md
```diff
@@ -0,0 +1,81 @@
...
+[ERC-7930] is an address format for (address, chain) pairs supporting any chain, and also including metadata on the chain & addresses' type in a self-describing way, reducing the need to communicate metadata or context out-of-band.
+This CAIP aims to append a living & mutable repository of chain-specific address serialization methods to the CASA system in the form of its namespace-specific profiles, enabling all current and future chains to uniformly conform to Interoperable Addresses.
+
+## Motivation
+
+Standards like CAIP-2 and CAIP-10 are simple text representations of addresses & chain namespaces/references, but they do not address:
```
obstropolos:
```suggestion
Standards like CAIP-2 and CAIP-10 are simple text representations of addresses & chain namespaces and references, but they do not address:
```

CAIPs/caip-350.md
```diff
@@ -0,0 +1,81 @@
...
+- Binary representation of addresses and/or chain namespace+reference: Relevant & desired for on-chain usage.
+- Canonicity: CAIP-10, being a generic text format, leaves the canonization and serialization of text address to on-chain addresses to each namespace's profile. Unless a given namespace's profile specifies such logic, the standard itself makes no universal guarantees on a blockchain account (_target address_ in ERC-7930 parlance) having only one CAIP-10 representation. For namespaces where such canonicity is neither inherent nor specified by the CASA profile, duplicate entries can be created (e.g. when used as dictionary keys).
+- Completeness: both formats have limits on the length of their identifiers, which are reasonable for human-readability but often identifiers have to drop meaningful information to conform to those length requirements. While that information might be easy to look up in the context of wallet software, doing so within a smart contract would not be possible.
```
obstropolos:
```suggestion
- Completeness: Both formats have limits on the length of their identifiers, which are reasonable for human-readability, but often identifiers have to drop meaningful information to conform to those length requirements. While that information might be easy to look up in the context of wallet software, doing so within a smart contract would not be possible.
```

obstropolos (commented):
This looks fine, just some tiny things - plus moving a tiny reference in to display.

CAIPs/caip-350.md
```diff
@@ -0,0 +1,81 @@
...
+
+The purpose of each namespace's profile is to specify deterministic and unambiguous conversions between the following format-pairs:
+- Customary address text formats of the ecosystem (which may also be described in its [CAIP-10] profile) and the Interoperable Names text representation defined in [ERC-7930]
+- Customary chain reference text formats of the ecosystem (which may also be described in [CAIP-2]) and the text representation defined in [ERC-7930]
+- [ERC-7930]'s Interoperable Address and [ERC-7930]'s Interoperable Name of the ecosystem's chain references
+- [ERC-7930]'s Interoperable Address and [ERC-7930]'s Interoperable Name of the ecosystem's addresses *for all address formats of said ecosystem*
+
+And must also define the binary id of the namespace itself, akin to [ERC-7930]'s `ChainType`
+
+For this, every namespace MUST define specify all of the above in a CAIP-350 profile to maximize interoperability and review. A [template for these profiles](https://github.com/ChainAgnostic/namespaces/blob/main/_template/caip350.md) is defined in the Namespaces registry.
```
bumblefudge:
```suggestion
For this, every namespace MUST specify all of the above in a CAIP-350 profile to maximize interoperability and review. A [template for these profiles](https://github.com/ChainAgnostic/namespaces/blob/main/_template/caip350.md) is defined in the Namespaces registry.
```

CAIPs/caip-350.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Security Considerations
+
+It is possible that a CAIP-350 profile for a given chain namespace is not able to guarantee requirements of canonicity which are a desired property of Interoperable Addresses due to characteristics of the chain namespace, and systems relying on that canonicity may exhibit unexpected behavior as a result. When this is the case, it should be noted in the `Extra Considerations` section.
+
+## Backwards Compatibility
+
+This standard actively seeks to be exhaustive in its backwards compatibility with CAIP-2 and CAIP-10.
+
+Namespaces are responsible over clarifying which conversions to and from the aforementioned standards are:
```
bumblefudge:
```suggestion
Namespace profiles SHOULD clarify which conversions to and from the aforementioned standards are:
```

CAIPs/caip-350.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Security Considerations
+
+It is possible that a CAIP-350 profile for a given chain namespace is not able to guarantee requirements of canonicity which are a desired property of Interoperable Addresses due to characteristics of the chain namespace, and systems relying on that canonicity may exhibit unexpected behavior as a result. When this is the case, it should be noted in the `Extra Considerations` section.
+
+## Backwards Compatibility
+
+This standard actively seeks to be exhaustive in its backwards compatibility with CAIP-2 and CAIP-10.
+
+Namespace profiles SHOULD clarify which conversions to and from the aforementioned standards are:
```
bumblefudge, line 67:
```suggestion
Namespace profiles SHOULD clarify which conversions to and from the aforementioned standards are:

```

alialobidm:
>

Issue #354, opened 2025-04-29
Withdraw CAIP-20


u59149403:
I propose to withdraw CAIP-20.

We should refer to native currencies of chains (I mean word "chain" in CAIP-2 sense) just by using word "native".

I. e. we should refer to ether as `eip155:1/native`.

Why? The main reason is this: some native currencies simply don't have SLIP-44. For example, let's consider `eip155:199` (BitTorrent chain). Its native currency BTT is not listed in SLIP-44. SLIP-44 contains ByteTrade BTT (`slip44:34952`, https://bytetrade.io ), but it is completely different from BitTorrent BTT ( https://bt.io/btt ). (In fact, link https://bytetrade.io is broken now, so it seems that this project is dead.)

Moreover, it seems BitTorrent BTT simply will never be added to SLIP-44. SLIP-44 says: "Coin types will be added only if there is a wallet implementing BIP-0044 for desired coin". There is no such wallet for BTT, so it will never be added. In fact, page https://app.bt.io/wallet , titled "BitTorrent Chain Wallet", doesn't offer any desktop wallet, but instead simply offers to connect MetaMask.

So, currently CAIP-19 cannot refer to BitTorrent BTT, and, assuming CAIPs will not change, will never be able.

But my proposal allows to refer to it simply as `eip155:199/native`.

Okay, now let me provide other arguments.

As well as I understand, every chain (in CAIP-2 sense) have exactly one so-called native currency. (I'm not aware of any chains, which have no native currency or have multiply native currencies.)

As well as I understand, SLIP-44 lists native currencies. Thus, every entry in SLIP-44 also indirectly describes chain (in CAIP-2 sense) of that currency. But some SLIP-44 currencies are native currencies of multiple chains. For example, ether (`slip44:60`) is native currency for Ethereum (`eip155:1`) and Base (`eip155:8453`). But still every SLIP-44 native currency has some designated main chain. For example, main chain for ether (`slip44:60`) is Ethereum (`eip155:1`).

Okay, so, in short: SLIP-44 lists native currencies, and every chain (in CAIP-2 sense) has exactly one native currency. Thus, CAIP-2 id is enough to identify that currency. Thus SLIP-44 id is not needed, it is okay to just use word "native".

[...]

PR #356, opened 2025-05-21
CAIP-XXX: Chain-Agnostic Token Payment URI

pedrouid:
Proposed Schema

caip:<CHAIN>/pay?recipient=<RECIPIENT>&token=<TOKEN>&amount=<AMOUNT>

### DIFF
--- a/CAIPs/caip-xxx.md
+++ b/CAIPs/caip-xxx.md
@@ -1,0 +1,126 @@
+---
+
+# CAIP-XX: Chain-Agnostic Token Payment URI
+
+**Title**: Chain-Agnostic Token Payment URI
+**Author**: Pedro Gomes
+**Status**: Draft
+**Type**: Standard
+**Created**: 2025-05-20
+
+---
+
+## Abstract
+
+This proposal defines a URI scheme for token-based payments in a chain-agnostic format. The format encodes the target blockchain, recipient account, token asset, and optional amount using URL-safe syntax and Chain Agnostic Improvement Proposals (CAIPs):
+
+* **CAIP-2** for chain identification
+* **CAIP-19** for token identification (excluding chain prefix)
+* **CAIP-10** for account identification (excluding chain prefix)
+
+---
+
+## Motivation
+
+Digital payments in Web3 are fragmented across wallets, chains, and token standards. Current methods lack consistency and interoperability, especially for use in deep links and QR codes within mobile apps, websites, ecommerce checkouts, or social media posts.
+
+A standardized URI string provides a simple, readable, and composable format for triggering token transfers across different chains using a unified scheme. This enables:
+
+* One-click payments from mobile wallets via deep linking
+* QR code generation for point-of-sale payments
+* Standardized checkout links for ecommerce platforms
+* Tipping or payment links embedded in social media profiles or messages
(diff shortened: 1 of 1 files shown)


oed:
What's the purpose of this? Wouldn't the same problem be solved with erc7715?

CAIPs/caip-xxx.md
```diff
@@ -0,0 +1,126 @@
+---
+
+# CAIP-XX: Chain-Agnostic Token Payment URI
+
+**Title**: Chain-Agnostic Token Payment URI
```
bumblefudge, line 5:
```suggestion
---
caip: 356
title: ...
```
YAML got botched somehow (were you pasting from a Notion?)

CAIPs/caip-xxx.md
```diff
@@ -0,0 +1,126 @@
...
+
+### URI Format
+
+```
+caip:<CHAIN>/pay?recipient=<RECIPIENT>&token=<TOKEN>&amount=<AMOUNT>
+```
+
+* `CHAIN` is **required** and follows CAIP-2
+* `recipient` is **required**, follows CAIP-10 but omits chain prefix
+* `token` is **required**, follows CAIP-19 but omits chain prefix
```
bumblefudge, line 47: what if the token in question is the native token of that chain? the MetaMask folks pointed out that caip-19 doesn't differentiate gas/native tokens from erc20-equivalents...

CAIPs/caip-xxx.md
```diff
@@ -0,0 +1,126 @@
...
+* `amount` is **optional**, denominated in the smallest unit of the token (e.g., wei, lamports, satoshi)
```
bumblefudge, line 48: as an integer? as a "json number" (i.e. stringtyped)? does this work for ERC-1155s or other special-snowflake tokens?

CAIPs/caip-xxx.md
bumblefudge: See also [issue #354](https://github.com/ChainAgnostic/CAIPs/issues/354#issuecomment-2925335141)

CAIPs/caip-xxx.md
```diff
@@ -0,0 +1,126 @@
...
+* Easy parsing and validation by wallet SDKs, apps, and browsers
+
+---
+
+## Specification
+
+### URI Format
+
+```
+caip:<CHAIN>/pay?recipient=<RECIPIENT>&token=<TOKEN>&amount=<AMOUNT>
```
bumblefudge, line 42: This all makes perfect sense to me except the first component.  why not:

`caip://<caip-2>/pay?...`

(a network/chain is nothing if not an authority!)

or

`urn:caip356:payment:<caip-2>?...`

or

https://<example.com>/pay?...

?

bumblefudge: i'd also be interested how this relates to the broader issue of how CAIPs combine to form a unified URN and/or URL scheme:
https://github.com/ChainAgnostic/CAIPs/pull/322/files

"pay" kinda introduces a "verb" into our so-far-noun-only cinematic universe, so maybe it's time to finally commit to a grammar that have nouns and verbs in an extendible system 😅

PR #358, opened 2025-05-26
CAIP-XXX: Universal Payment Request Method

lukaisailovic:
This CAIP standardizes a `wallet_pay` JSON-RPC method enabling one-click cryptocurrency payments across wallets and dapps. Allows merchants to specify multiple payment options (cross-chain assets) in a single request, letting wallets automatically select the optimal payment method based on user's available assets. Eliminates the current multi-step payment flow (select token → select chain → generate address → manual transfer) by moving payment choice and execution to the wallet, reducing friction from 4-6 clicks to a single interaction.

### DIFF
--- a/CAIPs/caip-xxx.md
+++ b/CAIPs/caip-xxx.md
@@ -1,0 +1,239 @@
+---
+caip: CAIP-X <X will be changed to the PR number if accepted>
+title: Universal Payment Request Method
+author: Luka Isailovic (@lukaisailovic), Derek Rein (@arein)
+discussions-to: []
+status: Draft
+type: Standard
+created: 2025-05-26
+updated: 2025-05-26
+requires: 2, 10, 19
+replaces: 
+---
+
+
+## Simple Summary
+<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the CAIP.-->
+A standard for enabling one-interaction cryptocurrency payment experiences across wallets and dapps, allowing payment information to be transmitted in a single round-trip.
+
+## Abstract
+<!--A short (~200 word) description of the technical issue being addressed.-->
+This CAIP standardizes a wallet <> dapp JSON-RPC method `wallet_pay` for more efficient communication about the purchase intent from the dapp to the wallet.
+The method allows merchants to specify product details and payment requirements enabling wallets to handle payment execution with minimal user interaction.
+
+## Motivation
+<!--The motivation is critical for CAIP. It should clearly explain why the state of the art is inadequate to address the problem that the CAIP solves. CAIP submissions without sufficient motivation may be rejected outright.-->
+Current cryptocurrency payment experiences are either error-prone (manual transfers, address QR codes) or suboptimal, requiring multiple interactions from the user.
+In addition to this, different payment providers implement different payment experiences, creating confusion.
+
+Solutions like EIP-681 or `bitcoin:` url are ecosystem specific and didn't get much support from the wallets. They rely on the QR code scan as well, which means that they can't be batched during connection with protocols like WalletConnect.
+
+By standardizing the payment experience both dapp side and wallet side, we can reduce user errors during payment, provide the payment experience in as few clicks as possible reducing the friction in crypto payments.
+ 
+The method transmits all the possible payment requests so the wallet can pick the most optimal one based on the assets that user has in the account.
+
+## Specification
+<!--The technical specification should describe the standard in detail. The specification should be detailed enough to allow competing, interoperable implementations. -->
+### Method: `wallet_pay`
+#### Request
+
+```typescript
+type Hex = `0x${string}`;
+
+type ProductMetadata = {
+  name: string;
+  description?: string;
+  imageUrl?: string;
+  price?: string; 
+}
+
+
+type PaymentOption = {
+  asset: string; 
+  amount: Hex;
+  recipient: string;
+}
+
+// JSON-RPC Request
+type PayRequest = {
+  orderId: string;
+  acceptedPayments: PaymentOption[]; 
+  products?: ProductMetadata[]; 
+  expiry?: number;
+}
+
+```
+
+The application **MUST** include:
+- An `orderId` that uniquely identifies this payment request. `orderId` **MUST NOT** be longer than 128 characters.
+- At least one entry in the `acceptedPayments` array
+
+
+The `orderId` field **MUST** be a string that uniquely identifies the payment request. Implementations **SHOULD** ensure this ID is unique across their system to prevent collisions.
+
+The `acceptedPayments` field **MUST** be an array of `PaymentOption` objects. Each element in the array represents a payment option that the wallet can choose from to complete the payment.
+
+For `PaymentOption` options:
+- The `recipient` field **MUST** be a valid [CAIP-10][] account ID.
+- The `asset` field **MUST** follow the [CAIP-19][] standard.
+- The `amount` field **MUST** be a hex-encoded string representing the amount of the asset to be transferred.
+- The [CAIP-2][] chainId component in the [CAIP-19][] `asset` field **MUST** match the [CAIP-2][] chainId component of the [CAIP-10][] `recipient` account ID.
+
+The application **MAY** include:
+- `products` array containing information about the products being purchased
+- `expiry` timestamp for the payment request
+
+If the `products` array is provided, each entry **MUST** include at least a `name`. The `description`, `imageUrl`, and `price` fields are **OPTIONAL** but **RECOMMENDED** for providing a better user experience.
+
+If the `expiry` field is provided, it **MUST** be a UNIX timestamp (in seconds) after which the payment request is considered expired. Wallets **SHOULD** check this timestamp before processing the payment.
+
+
+Request example:
+```json
+{
+    "orderId": "order-123456",
+    "acceptedPayments": [
+      {
+        "recipient": "eip155:1:0x71C7656EC7ab88b098defB751B7401B5f6d8976F",
+        "asset": "eip155:1/erc20:0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
+        "amount": "0x5F5E100" 
+      },
+      {
+        "recipient": "solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp:9WzDXwBbmkg8ZTbNMqUxvQRAyrZzDsGYdLVL9zYtAWWM",
+        "asset": "solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp/slip44:501",
+        "amount": "0x6F05B59D3B20000"
+      }
+    ],
+    "products": [
+      {
+        "name": "Donut",
+        "description": "Donut with extra chocolate sprinkles on top",
+        "imageUrl": "https://example.com/images/donut.png",
+        "price": "$100.00"
+      }
+    ],
+    "expiry": 1709593200 
+  }
+```
+
+#### Response
+```typescript
+type PayResult = {
+  orderId: string; 
+  txid: string; 
+  recipient: string;  
+  asset: string;  
+  amount: Hex; 
+}
+```
+
+The wallet's response MUST include:
+- `orderId` that matches the original request
+- `txid` with the transaction identifier on the blockchain
+- `recipient` that received the payment. It **MUST** be a valid [CAIP-10][] account ID.
+- `asset` that was used for payment. It **MUST** follow the [CAIP-19][] standard.
+- `amount` that was paid. It **MUST** be represented in hex string
+
+`txid` **MUST** be a valid transaction identifier on the blockchain network specified in the asset's chain ID.
+
+ `recipient`, `asset`, and `amount` **MUST** match those specified in the selected direct payment option in the `acceptedPayments` array.
+
+
+Example response:
+```json
+{
+    "orderId": "order-123456",
+    "txid": "0x8a8c3e0b1b812182db4cabd81c9d6de78e549fa3bf3d505d6e1a2b25a15789ed",
+    "recipient": "eip155:1:0x71C7656EC7ab88b098defB751B7401B5f6d8976F",
+    "asset": "eip155:1/erc20:0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
+    "amount": "0x5F5E100"
+}
+```
+
+#### Error Handling
+If the payment process fails, the wallet **MUST** return an appropriate error message:
+```typescript
+type PayError = {
+  code: number;
+  message: string;
+  data?: any;
+}
+```
+The wallet **MUST** use one of the following error codes when the pay request fails:
+
+- When user rejects the payment
+    - code = 8001
+    - message = "User rejected payment"
+- When no matching assets are available in user's wallet
+    - code = 8002
+    - message = "No matching assets available"
+- When the payment request has expired
+    - code = 8003
+    - message = "Payment request expired"
+- When there are insufficient funds for the payment
+    - code = 8004
+    - message = "Insufficient funds"
+
+
+If a wallet does not support the `wallet_pay` method, it **MUST** return an appropriate JSON-RPC error with code -32601 (Method not found).
+
+Example error response:
+```json
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "error": {
+    "code": 8001,
+    "message": "User rejected payment"
+  }
+}
+```
+
+## Rationale
+
+This specification evolved through multiple iterations to address fundamental usability issues in cryptocurrency payment flows. Initial exploration began as a CAIP alternative to EIP-681, but analysis of existing payment service provider (PSP) implementations revealed significant friction in current user experiences.
+
+Existing cryptocurrency payment flows typically require users to:
+- Select a token type
+- Choose a blockchain network
+- Wait for address/QR code generation
+- Complete the transfer manually
+
+This multi-step process creates excessive friction, often requiring 4-6 user interactions for a simple payment.
+
+The `wallet_pay` method addresses these limitations by:
+- Moving choice to the wallet rather than forcing merchants to pre-select payment methods, wallets can filter available options based on user account balances and preferences
+- All payment options are transmitted in one request, eliminating the need for multiple user interactions
+- The response includes transaction ID and execution details, providing immediate confirmation
+- Can be batched with connection establishment, enabling "connect + pay" flows in protocols like WalletConnect
+
+### Alternative Approaches Considered
+
+An intermediate solution involved encoding multiple payment addresses in a single QR code, allowing merchants to present all payment options simultaneously. However, this approach proved impractical for dapp implementations because:
+
+- PSPs cannot determine which payment option was selected
+- Monitoring requires polling up to 20+ addresses simultaneously
+- No confirmation mechanism exists for payment completion
+
+## Test Cases
+<!--Please add diverse test cases here if applicable. Any normative definition of an interface requires test cases to be implementable. -->
+TODO
+## Security Considerations
+<!--Please add an explicit list of intra-actor assumptions and known risk factors if applicable. Any normative definition of an interface requires these to be implementable; assumptions and risks should be at both individual interaction/use-case scale and systemically, should the interface specified gain ecosystem-namespace adoption. -->
+TODO
+## Privacy Considerations
+<!--Please add an explicit list of intra-actor assumptions and known risk factors if applicable. Any normative definition of an interface requires these to be implementable; assumptions and risks should be at both individual interaction/use-case scale and systemically, should the interface specified gain ecosystem-namespace adoption. -->
+TODO
+## Backwards Compatibility
+<!--All CAIPs that introduce backwards incompatibilities must include a section describing these incompatibilities and their severity. The CAIP must explain how the author proposes to deal with these incompatibilities. CAIP submissions without a sufficient backwards compatibility treatise may be rejected outright.-->
+All CAIPs that introduce backwards incompatibilities must include a section describing these incompatibilities and their severity. The CAIP must explain how the author proposes to deal with these incompatibilities. CAIP submissions without a sufficient backwards compatibility treatise may be rejected outright.
+
+## References 
+<!--Links to external resources that help understanding the CAIP better. This can e.g. be links to existing implementations. See CONTRIBUTING.md#style-guide . -->
+
+- [CAIP-1][CAIP-1] defines the CAIP document structure
+
+[CAIP-1]: https://ChainAgnostic.org/CAIPs/caip-1
+
+## Copyright
+Copyright and related rights waived via [CC0](../LICENSE).


CAIPs/caip-358.md
```diff
@@ -0,0 +1,239 @@
...
+
+```typescript
+type Hex = `0x${string}`;
+
+type ProductMetadata = {
+  name: string;
+  description?: string;
+  imageUrl?: string;
+  price?: string; 
+}
```
pedrouid, line 48: Personally I would remove this interface as this could be addressed through a different API that references the `orderId` instead

CAIPs/caip-358.md
```diff
@@ -0,0 +1,239 @@
...
+  asset: string; 
+  amount: Hex;
+  recipient: string;
+}
+
+// JSON-RPC Request
+type PayRequest = {
+  orderId: string;
+  acceptedPayments: PaymentOption[]; 
+  products?: ProductMetadata[];
```
pedrouid, line 61: I would remove this following my comment above

CAIPs/caip-358.md
```diff
@@ -0,0 +1,239 @@
...
+  amount: Hex;
+  recipient: string;
+}
+
+// JSON-RPC Request
+type PayRequest = {
+  orderId: string;
+  acceptedPayments: PaymentOption[]; 
+  products?: ProductMetadata[]; 
+  expiry?: number;
```
pedrouid, line 62: most of the time these are mandatory so I would change this to be required

pedrouid (commented):
This looks awesome! Exactly in line with the feedback that I've been receiving 👌 

Using this standard we could easily create a universal standard for "Pay with Ethereum" or "Pay with Solana".... and of course.... WalletConnect would offer this as part of the Network!

kdenhartog:
Also as a side note, would you prefer that I publish my draft document or offer a PR onto yours so that we can account for it? We'll likely need both interactions to seamless make this work across platforms and interactions.

CAIPs/caip-358.md
```diff
@@ -0,0 +1,239 @@
...
+
+Existing cryptocurrency payment flows typically require users to:
+- Select a token type
+- Choose a blockchain network
+- Wait for address/QR code generation
+- Complete the transfer manually
+
+This multi-step process creates excessive friction, often requiring 4-6 user interactions for a simple payment.
+
+The `wallet_pay` method addresses these limitations by:
```
arein, line 216: for consideration: EIP-681 and Solana Pay are similar but chain specific. In order to remove chain selection on the PSP level it has to be chain agnostic by design

CAIPs/caip-358.md
```diff
@@ -0,0 +1,239 @@
...
+  "error": {
+    "code": 8001,
+    "message": "User rejected payment"
+  }
+}
+```
+
+## Rationale
+
+This specification evolved through multiple iterations to address fundamental usability issues in cryptocurrency payment flows. Initial exploration began as a CAIP alternative to EIP-681, but analysis of existing payment service provider (PSP) implementations revealed significant friction in current user experiences.
```
arein:
```suggestion
This specification evolved through multiple iterations to address fundamental usability issues in cryptocurrency payment flows. Initial exploration began as a CAIP alternative to EIP-681/Solana Pay, but analysis of existing payment service provider (PSP) implementations revealed significant friction in current user experiences.
```

CAIPs/caip-358.md
```diff
@@ -0,0 +1,239 @@
...
+  "id": 1,
+  "jsonrpc": "2.0",
+  "error": {
+    "code": 8001,
+    "message": "User rejected payment"
+  }
+}
+```
+
+## Rationale
```
arein, line 203: Should something about supporting Smart Contract payment flavors like Coinbase Commerce be added here?

 Also should the be a discussion somewhere in the CAIP on how developers can reconcile payments? What happens when the wallet can send the transaction but then the internet cuts off and the user can no longer switch back to the dapp?

CAIPs/caip-358.md
jxom: Why? This interface is symmetrical to the [Payment Request API](https://developer.mozilla.org/en-US/docs/Web/API/PaymentRequest/PaymentRequest#displayitems)

jxom:
Would it be a good idea to align the API to [`PaymentRequest.details`](https://developer.mozilla.org/en-US/docs/Web/API/PaymentRequest/PaymentRequest#details)? There are some properties in there which may be nice for this API too.

lukaisailovic:
@kdenhartog I don't think this only focuses on the QR code scan. It defines the RPC but not the method of transfer. It can be via the QR Code (WalletConnect) or via the provider.

I'll create a discussion and we can follow up there

CAIPs/caip-358.md
lukaisailovic: I removed it for now. My reasoning is that while its nice to have, it makes wallet integrations a bit harder. I'd prefer to not have it in v1 and maybe add it as the follow up CAIP with something like `productMetadata` field which contains the full specification or maybe a URL to the resource

CAIPs/caip-358.md
```diff
@@ -0,0 +1,227 @@
...
+#### Error Handling
+If the payment process fails, the wallet **MUST** return an appropriate error message:
+```typescript
+type PayError = {
+  code: number;
+  message: string;
+  data?: any;
+}
+```
+The wallet **MUST** use one of the following error codes when the pay request fails:
```
bumblefudge, line 173: might it be more future-proof to require an error message 800X or 80XX, or is the idea specifically that these are the only 4 valid errors to throw?

CAIPs/caip-358.md
```diff
@@ -0,0 +1,227 @@
...
+
+An intermediate solution involved encoding multiple payment addresses in a single QR code, allowing merchants to present all payment options simultaneously. However, this approach proved impractical for dapp implementations because:
+
+- PSPs cannot determine which payment option was selected
+- Monitoring requires polling up to 20+ addresses simultaneously
+- No confirmation mechanism exists for payment completion
+
+## Test Cases
+<!--Please add diverse test cases here if applicable. Any normative definition of an interface requires test cases to be implementable. -->
+TODO
```
bumblefudge, line 234: I would recommend adding both positive and negative cases here, it never hurts! it might also be a good place to stick end-to-end examples showing how it can be combined with x402, caip-25, wallet-standard, etc (i.e. end-to-end examples showing other layers, not just payment requests).

CAIPs/caip-358.md
```diff
@@ -0,0 +1,227 @@
...
+## Backwards Compatibility
+<!--All CAIPs that introduce backwards incompatibilities must include a section describing these incompatibilities and their severity. The CAIP must explain how the author proposes to deal with these incompatibilities. CAIP submissions without a sufficient backwards compatibility treatise may be rejected outright.-->
+All CAIPs that introduce backwards incompatibilities must include a section describing these incompatibilities and their severity. The CAIP must explain how the author proposes to deal with these incompatibilities. CAIP submissions without a sufficient backwards compatibility treatise may be rejected outright.
+
+## References 
+<!--Links to external resources that help understanding the CAIP better. This can e.g. be links to existing implementations. See CONTRIBUTING.md#style-guide . -->
+
+- [CAIP-1][CAIP-1] defines the CAIP document structure
+
+[CAIP-1]: https://ChainAgnostic.org/CAIPs/caip-1
```
bumblefudge:
```suggestion
- [CAIP-1] defines the CAIP document structure
- [EIP-681] is ethereum-specific prior art that also includes gas information in the URI 

[CAIP-1]: https://ChainAgnostic.org/CAIPs/caip-1
[EIP-681]: https://eips.ethereum.org/EIPS/eip-681
```

bumblefudge:
> Would it be a good idea to align the API to [`PaymentRequest#details`](https://developer.mozilla.org/en-US/docs/Web/API/PaymentRequest/PaymentRequest#details)? There are some properties in there which may be nice for this API too.

What does this mean, concretely? does this mean renaming `orderId` to `details.id`, or just mentioning the latter in the definition of the former so people know they can pass the former in the place of the latter? I'm all for using the w3c precedent but I am curious how far y'all are thinking of "aligning" the format?  it might be good to tag the other yoav weiss or someone else from shopify's web/devrel team, as they might have deep insight into how that API gets used in the wild in "web2 ecommerce" that could inform this minimalist implementation thereof if you take the alignment far enough to be compliant!

jxom:
> What does this mean, concretely?
It means exposing the interface (and possible a lot of the [spec](https://www.w3.org/TR/payment-request/) itself!) to Wallets, and not _just_ Browsers. Deduping and not reinventing too much would be ideal, as I see this spec having nearly identical intention.

kdenhartog:
> It means exposing the interface (and possibly a lot of the [spec](https://www.w3.org/TR/payment-request/) itself!) to Wallets, and not just Browsers.

Am I understanding properly that you want to polyfill on this? I think if we do this we're going to walk back into the 6963 problem where multiple extensions will polyfill ontop of this. This is one of the issues that's plagued passkeys too is that multiple providers want to participate in the flow (extensions/browsers/OS).

While I'm a fan of mapping the logic to what they've already got, I'm a bit concerned about reintroducing these issues we just solved. Also, there is a possibility for a browser to just prevent overrides of the scripts they inject into the page, so we may want to be careful here such that wallets' participation in this flow don't get consumed by the underlying browser implementation. Brave specifically has taken the stance we won't do this and have settings to allow the user to override it. However, hearing what happened during the first go with PaymentRequest API (plus hardly ever seeing it used) I'm cautious about how others may participate since core business revenues are often built on capturing these flows and the data that's passed in it.

jxom:
No, I purely mean just aligning the interface.

CAIPs/caip-358.md
```diff
@@ -0,0 +1,227 @@
...
+  asset: string; 
+  amount: Hex;
+  recipient: string;
+}
+
+// JSON-RPC Request
+type PayRequest = {
+  orderId: string;
+  acceptedPayments: PaymentOption[]; 
+  expiry: number;
```
xzilja, line 55: Would it be useful to also add an optional, human readable identifier like `label` / `description` here? Something similar to [Solana Pay](https://docs.solanapay.com/spec) to help user identify payment request. Perhaps multiple of these can come through when they open up a wallet or come back online etc.

CAIPs/caip-358.md
```diff
@@ -0,0 +1,227 @@
...
+
+- PSPs cannot determine which payment option was selected
+- Monitoring requires polling up to 20+ addresses simultaneously
+- No confirmation mechanism exists for payment completion
+
+## Test Cases
+<!--Please add diverse test cases here if applicable. Any normative definition of an interface requires test cases to be implementable. -->
+TODO
+## Security Considerations
+<!--Please add an explicit list of intra-actor assumptions and known risk factors if applicable. Any normative definition of an interface requires these to be implementable; assumptions and risks should be at both individual interaction/use-case scale and systemically, should the interface specified gain ecosystem-namespace adoption. -->
```
xzilja: If possible, should provide a verifiable source of payment request i.e dapp domain

CAIPs/caip-358.md
```diff
@@ -0,0 +1,227 @@
...
+<!--The technical specification should describe the standard in detail. The specification should be detailed enough to allow competing, interoperable implementations. -->
+### Method: `wallet_pay`
+#### Request
+
+```typescript
+type Hex = `0x${string}`;
+
+type PaymentOption = {
+  asset: string; 
+  amount: Hex;
```
xzilja, line 46: I am rusty on this, but does amount need to be a Hex, or can it be a simpler string like "100.0"? Is it common for non-evm wallets to use hex's here?

CAIPs/caip-358.md
gitteri: Agreed. Solana Pay includes the following which are not specified here but have had proven use:
* Label: UTF-8 string that describes the source of the transfer request (ex: a brand, store, app etc)
* Message: UTF-8 string that is user facing to describe the nature of the request. This can help add context to wallets who display this request.
* Memo: similar to the orderId above but more broad in specification

Additionally, on Solana you can embed arbitrary public keys in the message to allow for easy tracking or reference. This is the "reference" field and is extremely helpful for wallets/apps to be able to track client IDs

CAIPs/caip-358.md
```diff
@@ -0,0 +1,227 @@
...
+{
+    "orderId": "order-123456",
+    "acceptedPayments": [
+      {
+        "recipient": "eip155:1:0x71C7656EC7ab88b098defB751B7401B5f6d8976F",
+        "asset": "eip155:1/erc20:0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
+        "amount": "0x5F5E100" 
+      },
+      {
+        "recipient": "solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp:9WzDXwBbmkg8ZTbNMqUxvQRAyrZzDsGYdLVL9zYtAWWM",
```
gitteri: nit: historically we have recommended using the solana genesis hash as the "chain id". As is this would not allow for disambiguation between mainnet and devnet for example

CAIPs/caip-358.md
```diff
@@ -0,0 +1,227 @@
...
+    "asset": "eip155:1/erc20:0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
+    "amount": "0x5F5E100"
+}
+```
+
+#### Idempotency
+The `wallet_pay` method **MUST** be idempotent for the same `orderId`. This ensures robustness in case of connection failures or timeout scenarios.
+
+Requirements:
+- If a payment with the same `orderId` has already been completed successfully, the wallet **MUST** return the original `PayResult` without executing a new payment
```
gitteri, line 149: Can you add more details as to the unique constraints for the orderId? is this per merchant app? Many chains don't have the ability to embed orderIds onchain so this information is likely held offchain. 

On Solana, it's common to use the memo + reference key to uniquely identify the transaction as this data is onchain

CAIPs/caip-358.md
kdenhartog: Have there been instances of this data being fraudulently submitted by a scam site for the purposes of tricking the user? It seems like due to instant settlements of crypto payments, there's more risk to submitting these payments to untrusted sites that have no intention of shipping goods or services once they receive payment.

jxom:
> To start, I agree this makes the most sense. You're going to hate me here though... but I want to expose this only with the https://github.com/ChainAgnostic/CAIPs/pull/282/

It's good that we agree there! I was only concerned about the interface part, not the transport part. From my understanding, this spec outlines a JSON-RPC format and the behavior of it, but not the underlying transport itself. Is it right to assume that this can be plugged into any JSON-RPC compatible transport? Whether it be EIP-6963, CAIP-282, etc? Seems like the most backwards-compatible approach. 
I am not disagreeing with you, just want to understand the scope of this spec.

lukaisailovic:
> > To start, I agree this makes the most sense. You're going to hate me here though... but I want to expose this only with the https://github.com/ChainAgnostic/CAIPs/pull/282/
> 
> 
> 
> It's good that we agree there! I was only concerned about the interface part, not the transport part. From my understanding, this spec outlines a JSON-RPC format and the behavior of it, but not the underlying transport itself. Is it right to assume that this can be plugged into any JSON-RPC compatible transport? Whether it be EIP-6963, CAIP-282, `window.ethereum`, `window.solana`, `window.mysickwallet`, etc? Seems like the most backwards-compatible approach. 
> 
> I am not disagreeing with you, just want to understand the scope of this spec.

@jxom You're exactly right. The spec intentionally doesn't define the transport layer, and any one of those you mentioned can be used.

kdenhartog:
Ok, cool sounds like we're all on the same page for that then because that was what I was expecting too. I think the other thing that might need to be defined is how the RPC request data can encoded into a URI that can be fired and forgotten. Then the recipient would poll the chain for data round it or potentially make a callback to a webhook or something.

This way we can make payment requests with the same format over QR codes/NFC/deeplinks/etc

lukaisailovic:
@kdenhartog URI encoding + webhook is significantly more complex, primarily for dapps who now need a backend to check payment status, but for the wallet as well. I specifically wanted to avoid that approach as this is much lower lift to implement. I'd prefer to leave it out of this spec, and if someone wants to do that, they can create a new spec and define exactly how that should be handled.

lukasrosario:
@lukaisailovic nice proposal! curious if you've considered a path where this use case could be handled as elegantly using EIP-5792. happy to provide more feedback on this API specifically, but i just wonder if we'd have more impact overall by leveraging momentum the space already has with `wallet_sendCalls`. as we've seen it takes a lot to coordinate all the necessary players 😅 .

for reference, Coinbase is also working on this feature: https://docs.base.org/identity/smart-wallet/guides/profiles#profiles , which uses EIP-5792 and could serve things like express checkout well. so im curious if there's some middle ground to be found here

CAIPs/caip-358.md
```diff
@@ -0,0 +1,227 @@
...
+
+type PaymentOption = {
+  asset: string; 
+  amount: Hex;
+  recipient: string;
+}
+
+// JSON-RPC Request
+type PayRequest = {
+  orderId: string;
```
lukasrosario: wondering if this should by inside of a good old `context` field. curious if you think this could be used for p2p use cases too, where there might not really be a meaningful `orderId`

CAIPs/caip-358.md
gitteri: I'm sure there have been instances.. it's crypto so scams are bound to happen. I think we can fix this by having some sort of tls-like auth mechanism, or perhaps an integrity signature. I'm not sure if this would be defined in this caip or at a protocol level or as an additional capability to be adopted by teams where it makes sense

CAIPs/caip-358.md
kdenhartog: ++ to making this optional. One of the cases I think we may be interested in this for eventually would be for P2P creator tipping in Brave Rewards.

jxom:
> Additionally, EIP-5792 is a low level RPC call that presumes a DApp has requested access to the account address and will format the transaction. This takes a different approach aiming for it to be a high level RPC API that doesn't require a connection for the page to engage with the RPC.

FWIW, ERC-5792 does not require a connection either (`from` is optional).

CAIPs/caip-358.md
arein: My $0.02: there's not just scams in the form of "this data being fraudulently submitted by a scam site for the purposes of tricking the user" but there's also the problem of "fake stores" selling you iPhones that don't exist.

I believe trying to accomplish this with `wallet_pay` in its initial form is too much. But in a future iteration or addendum the problem of merchant fraud insurance must be covered.

FWIW: if `wallet_pay` is transmitted via WalletConnect then the transport provides some scam protection via WalletConnects ["Verify API"](https://docs.reown.com/walletkit/ios/verify)

CAIPs/caip-358.md
```diff
@@ -0,0 +1,227 @@
...
+- PSPs cannot determine which payment option was selected
+- Monitoring requires polling up to 20+ addresses simultaneously
+- No confirmation mechanism exists for payment completion
+
+## Test Cases
+<!--Please add diverse test cases here if applicable. Any normative definition of an interface requires test cases to be implementable. -->
+TODO
+## Security Considerations
+<!--Please add an explicit list of intra-actor assumptions and known risk factors if applicable. Any normative definition of an interface requires these to be implementable; assumptions and risks should be at both individual interaction/use-case scale and systemically, should the interface specified gain ecosystem-namespace adoption. -->
+TODO
```
arein:
```suggestion
## Security Considerations
<!--Please add an explicit list of intra-actor assumptions and known risk factors if applicable. Any normative definition of an interface requires these to be implementable; assumptions and risks should be at both individual interaction/use-case scale and systemically, should the interface specified gain ecosystem-namespace adoption. -->
TODO

`wallet_pay` does not try to address various cases of merchant fraud that end-users are exposed to today. Specifically it does not try to tackle merchant fraud insurance in case the sold good is not delivered. It also does not attempt to provide dispute functionality. These present ideas for future work.
```

CAIPs/caip-358.md
```diff
@@ -0,0 +1,227 @@
...
+- The [CAIP-2][] chainId component in the [CAIP-19][] `asset` field **MUST** match the [CAIP-2][] chainId component of the [CAIP-10][] `recipient` account ID.
+
+The `expiry` field **MUST** be a UNIX timestamp (in seconds) after which the payment request is considered expired. Wallets **SHOULD** check this timestamp before processing the payment.
+
+
+Request example:
+```json
+{
+    "orderId": "order-123456",
+    "acceptedPayments": [
```
argjv: This may be for another CAIP or even controversial, but with stablecoin proliferation, merchants will soon accept dozens of different stablecoins while only caring about the underlying value (e.g., USD). Moreover, users often hold multiple USD stablecoins across different chains and should be able to use any of them when paying in USD, rather than having to choose a specific implementation. This also applies to other asset equivalents like BTC and wrapped BTC variants.
Adding a grouping or equivalency concept to the payment options where assets with the same underlying value can be grouped together would let wallets show "Pay $100 in USD" where users can either pick a specific asset or let the wallet automatically select the best option (lowest fees, available balance) instead of forcing them to pick between USDC on Ethereum vs Polygon vs Solana when they just want to pay in dollars.

CAIPs/caip-358.md
lukaisailovic: I agree with @arein 
This should definitely be left out. By default wallet would be unable to display untrusted information, or at least it should never do that.

CAIPs/caip-358.md
lukaisailovic: I think hex is a pretty good convention. If you put "100.0" then everyone has to do the conversion. Pretty much all downstream RPCs and SDKs accept hex

CAIPs/caip-358.md
lukaisailovic: The orderId doesn't have to be stored onchain. It should be unique for each order. Merchant generated UUID should work just fine

CAIPs/caip-358.md
lukaisailovic: Yeah, we can make it optional. 

@lukasrosario currently its the only property, but if we wanted to future proof it it might make sense

CAIPs/caip-358.md
lukaisailovic: I actually removed smart contract interactions. It adds too much complexity for not a lot of benefits. 

Currently the only reason its used is to identify payments, which can be done differently in a more efficient way. Also some implementations still use contracts with direct deposit into them

CAIPs/caip-358.md
kdenhartog: This seems worth exploring what works better here through implementation experimentation a bit more. When I was thinking about this, I was somewhat expecting that would be the case for the most part anyways, but through implementation specific wallet UX, so that we can optimize on this overtime. What I originally had in mind was that the wallet operates on the following heuristic:

1. Check which currencies the recipient wants and which are available without a swap/bridge
2. If payment can be handled with currently available funds, offer that first. It's almost certainly lowest in terms of fees, but this is an assumption that may be invalidated as networks and swap/bridge protocols compete for lower fees.
3. If swap/bridge is necessary, check which current currencies the user holds can cover funds. Also, determine which of the currencies the recipient wants to have the most liquidity. 
4. Then get a swap/bridge quote from users current funds that can cover for each pair of users funds and the most liquid stablecoin chosen.
5. Display these options to the user from least to most fees, with the default being least.

Additionally, allow the user to override as necessary in optional transaction panel, in case they don't want to use a particular fund. I think Users will demand what sort of UX they want here too, so we can optimize a bit further as the feature becomes available, without needing to change the RPC data model. That should remain our top priority here IMO.

From the merchant's perspective, one of the issues we've found with spending/receiving funds today with BAT and rewards ads is that there's a constant need to bridge/swap funds and in some cases off-ramp back to fiat (hopefully that's not a problem of the future). This can be a bit of a headache for accounting/tax purposes on the merchant side. So, part of the advantage to the recipient specifying network/currency exactly is that it makes merchant onboarding a bit easier in terms of fund management which makes it more likely that [Metcalfe's law](https://en.wikipedia.org/wiki/Metcalfe's_law) kicks in here. E.g. they can always have USDC on their network of choice and easily offramp that as needed to pay their invoices/employees. Additionally, since most wallets universally have some form of swap/bridge functionality I don't think it's going to be overly difficult for us to swap/bridge as needed and handle the complexity for the user. This also means we'll be able to easily handle currency conversion via stablecoins using the same heuristic too.

I may be making some incorrect assumptions here though, so thank you for calling this out. Ultimately, whatever the implementation feedback and early testing with users/merchants tells us here is going to be the decider IMO.

CAIPs/caip-358.md
```diff
@@ -0,0 +1,231 @@
...
+    "amount": "0x5F5E100"
+}
+```
+
+#### Idempotency
+The `wallet_pay` method **MUST** be idempotent for the same `orderId`. This ensures robustness in case of connection failures or timeout scenarios.
+
+Requirements:
+- If a payment with the same `orderId` has already been completed successfully, the wallet **MUST** return the original `PayResult` without executing a new payment
+- If a payment with the same `orderId` is currently pending, the wallet **SHOULD** return the result of the original payment attempt
```
jakubuid, line 150: Just to clarify. What's the result of the original payment attempt? If the payment is currently pending the result is not yet available

lukaisailovic: If the wallet returned `txid` and you submit the payment with the same `orderId`, the same `txid` will be returned. The result doesn't specify the tx state (pending/success)

kdenhartog: With that in mind, it would likely be useful to the site to be able to reliably use this. Otherwise it could potentially lead to a UX divergence where a wallet that doesn't support this re-prompts for a new payment where as a wallet that does respond automatically with the TxID without a user interaction.

Therefore, I think we should upgrade this to a MUST rather than a SHOULD. Is there some reason you went with SHOULD here instead?

CAIPs/caip-358.md
gitteri: btw hex is extremely not standard in Solana

lukaisailovic: We can consider `uint`

jxom: Don't think we need to overthink it. Hex is a universal convention beyond crypto applications (ie. JS `0x69` resolves to `105` as a number, and we are referring to a **JS**ON-RPC interface here) for numbers exceeding `Number.MAX_SAFE_INTEGER`.

kdenhartog: Since this is something that's going to be interpreted by the wallet is there a reason to not just go with JSON number? Seems like then it just becomes the wallets responsibility to format it correctly for the particular chain in the transaction. In this way language specific aspects around types (e.g. some need doubles, floats, uints, etc) also become simpler too.

CAIPs/caip-358.md
bumblefudge: > Adding a grouping or equivalency concept to the payment options where assets with the same underlying value can be grouped together would let wallets show "Pay $100 in USD"

@argjv Agreed that this requires orthogonal/separate specification, whether as another CAIP or outside of CASA. Still going to chime in here because I used to work at Circle and am fascinated personally by the topic-- tag me if this is proposed somewhere else as a spec!

Upfront I would just say that $100 in USDC (on any chain/in any formfactor) is an abstraction that makes sense to me (cf. Layer Zero and Circle's devrel strategy), but $100 in "any stablecoin" is a non-starter IMHO: not just because some people have hard regulatory requirements to use, e.g. Circle but not Tether, or Paxos but not Luna/GUSD/etc etc, but ALSO because their "semantic equivalence" is just not deliverable in practice given the way DEXs and solvers and bridges are incentivized by their contracts, API billing structures, and the terms of their integrations deals to favor one over the other and devoid "de minimis" monthly transaction volumes in any other stables...

What I _could_ see happening is USDC and comparable-regulated stablecoins federating to create one broader meta-coin, e.g. "USDC+", which would abstract over and translate between USDC on any chain, Paxos on any chain, etc etc.

CAIPs/caip-358.md
lukaisailovic: In this version, there are. If there are other errors, or different use cases someone can write a new CAIP

CAIPs/caip-358.md
lukaisailovic: @gitteri Are you referring to https://chainagnostic.org/CAIPs/caip-30?

gitteri: Didn't know this exists but yes

CAIPs/caip-358.md
```diff
@@ -0,0 +1,263 @@
...
+The application **MUST** include:
+
+- At least one entry in the `acceptedPayments` array
+- `expiry` timestamp for the payment request
+
+The application **MAY** include:
+
+- An `orderId` that uniquely identifies this payment request. If provided, `orderId` **MUST NOT** be longer than 128 characters.
+
+ When `orderId` is provided, it **MUST** be a string and implementations **SHOULD** ensure this ID is unique across their system to prevent collisions.
```
bumblefudge, line 69: would it make more sense to require (or at least recommend) a UUID full-stop? human-readable or monotonic orderIds seem like they would wreak havok since it's not just "across their own system" that collisions could create problems (for, e.g., indexers)

CAIPs/caip-358.md
```diff
@@ -0,0 +1,263 @@
...
+
+The `acceptedPayments` field **MUST** be an array of `PaymentOption` objects. Each element in the array represents a payment option that the wallet can choose from to complete the payment.
+
+For `PaymentOption` options:
+
+- The `recipient` field **MUST** be a valid [CAIP-10][] account ID.
+- The `asset` field **MUST** follow the [CAIP-19][] standard.
+- The `amount` field **MUST** be a hex-encoded string representing the amount of the asset to be transferred.
+- The [CAIP-2][] chainId component in the [CAIP-19][] `asset` field **MUST** match the [CAIP-2][] chainId component of the [CAIP-10][] `recipient` account ID.
```
bumblefudge, line 78: what about the [CAIP-107][] `namespace` component? 😏

CAIPs/caip-358.md
```diff
@@ -0,0 +1,263 @@
...
+
+The `expiry` field **MUST** be a UNIX timestamp (in seconds) after which the payment request is considered expired. Wallets **SHOULD** check this timestamp before processing the payment.
```
bumblefudge, line 80:
```suggestion
The `expiry` field **MUST** be a UNIX timestamp (in seconds) after which the payment request is considered expired. Wallets **SHOULD** check this timestamp is not in the past before proposing or submitting a payment.
```

CAIPs/caip-358.md
lukaisailovic: Wdym? chainId is namespace + reference (blockchain Id). Do you mean asset namespace in caip-19? That can't be used for anything, not sure why should we compare it to CAIP-2

CAIPs/caip-358.md
lukaisailovic: I think recommendation is fine, but I don't think it should be a requirement. This is dapp <> wallet RPC, so the orderId is not stored anywhere onchain

CAIPs/caip-358.md
```diff
@@ -0,0 +1,263 @@
...
+- If a payment with the same `orderId` has already been completed successfully, the wallet **MUST** return the original `PayResult` without executing a new payment
+- If a payment with the same `orderId` is currently pending, the wallet **SHOULD** return the result of the original payment attempt
+- If a payment with the same `orderId` has failed previously, the wallet **MAY** attempt the payment again or return the previous error
+- Wallets **SHOULD** maintain payment status for completed transactions for at least 24 hours after completion
+- If connection is lost during payment execution, dapps **MAY** retry the same request to query the payment status
+
+When `orderId` is not provided:
+
+- Each payment request **SHOULD** be treated as a new payment attempt
+- Wallets **MAY** implement their own deduplication logic based on other request parameters (recipient, asset, amount, expiry)
```
bumblefudge, line 158:
```suggestion
- Wallets **MAY**, however, implement their own deduplication logic based on other request parameters (recipient, asset, amount, expiry) to prevent unintended duplicate payments for identical requests
```

kdenhartog (approved):
This is more than ready to merge in a draft state. We can iterate as we go from here

obstropolos (approved):

lukaisailovic:
@bumblefudge I think you guys have to merge it right?

kelvin7891:
> This CAIP standardizes a `wallet_pay` JSON-RPC method enabling one-click cryptocurrency payments across wallets and dapps. Allows merchants to specify multiple payment options (cross-chain assets) in a single request, letting wallets automatically select the optimal payment method based on user's available assets. Eliminates the current multi-step payment flow (select token → select chain → generate address → manual transfer) by moving payment choice and execution to the wallet, reducing friction from 4-6 clicks to a single interaction.

PR #360, opened 2025-06-24
Modify CAIP-25 spec to account for additive vs atomic wallet_createSession requests

ffmcgee725:
Per our CASA discussion, we want to modify UI/UX behavior to include existing permissions in the pre-selected defaults we show to a user in iterative wallet_createSession calls.

So we suggest the following update to CAIP-25 to account for this.

### DIFF
--- a/CAIPs/caip-25.md
+++ b/CAIPs/caip-25.md
@@ -49,8 +49,9 @@ After a session is established between wallet and caller, subsequent `wallet_cre
 - When a `sessionId` is returned in the initial `wallet_createSession` response, subsequent `wallet_createSession` calls either:
   - include a previously used `sessionId` on the root of the request meaning this request is intended to modify that session, or
   - do not include a `sessionId`, in which case a new session is created - the respondent generates a new `sessionId` and sends it with the success response - and the previous session dangles in parallel (until its expiration, if applicable), though maintaining concurrent sessions is discouraged (see Security Considerations).
-- When the wallet does not provide a `sessionId` in its initial response, subsequent `wallet_createSession` calls target the previous singular session between caller and wallet.
-- `wallet_createSession` calls either increase scope, decrease scope, or both.
+- When the wallet does not provide a `sessionId` in its initial response, subsequent `wallet_createSession` calls target the previous singular session between caller and wallet. 
+ - Wallets MAY interpret the intent of subsequent scope requests as adding to or overwriting of existing permissions
+ - Users MAY choose to grant a subset, superset, or entirely different set of permissions
 
 When a user wishes to update the authorizations of an active session from within the wallet, the wallet should notify the caller of the changes with a [`wallet_sessionChanged`][CAIP-311] notification.
 


CAIPs/caip-25.md
```diff
@@ -46,10 +46,14 @@ Note that wallets NOT returning `sessionId`s MUST implement additional methods a
...
-- When a `sessionId` is returned in the initial `wallet_createSession` response, subsequent `wallet_createSession` calls either: 
+- When a `sessionId` is returned in the initial `wallet_createSession` response, subsequent `wallet_createSession` calls either:
   - include a previously used `sessionId` on the root of the request meaning this request is intended to modify that session, or
   - do not include a `sessionId`, in which case a new session is created - the respondent generates a new `sessionId` and sends it with the success response - and the previous session dangles in parallel (until its expiration, if applicable), though maintaining concurrent sessions is discouraged (see Security Considerations).
-- When the wallet does not provide a `sessionId` in its initial response, subsequent `wallet_createSession` calls overwrite the previous singular session between caller and wallet.
+- When the wallet does not provide a `sessionId` in its initial response, subsequent `wallet_createSession` calls either:
+  - overwrite the previous singular session between caller and wallet, or
+  - increment the previous session properties with the subsequent requested session properties.
```
jiexi: what are your thoughts on addressing this from a different direction. Much in the same way that a dapp may get less permissions than it asks for, it's also possible it gets more permissions than it asks for. What permissions the dapp requests is simply a suggestion, but the wallet and user do not have to accept it in its entirety

@adonesky1

adonesky1: Yeah I definitely think it makes sense to highlight that the `wallet_createSession` call is a suggestion and different wallets may handle the UX of how they present that suggestion to users, meaning as @jiexi says, in some cases users be given the option to modify the suggested permissions to add or remove accounts and scopes

CAIPs/caip-25.md
```diff
@@ -46,10 +46,14 @@ Note that wallets NOT returning `sessionId`s MUST implement additional methods a
...
+
+__We advise a dapp using this procedure to always call `wallet_revokeSession` before calling a subsequent `wallet_createSession`, if it intends to guarantee an overwriting of a previously existing set of properties and authorization scopes.__
```
adonesky1:
```suggestion
```

I think we can remove this elaboration, IMO it may cause more confusion than clarity

CAIPs/caip-25.md
ffmcgee725: Added in https://github.com/ChainAgnostic/CAIPs/pull/360/commits/f076279310874c0e735c68d46693c2f1ebceae0f

pedrouid (approved):
👍

CAIPs/caip-25.md
adonesky1: @ffmcgee725  The paragraph you added in https://github.com/ChainAgnostic/CAIPs/commit/f076279310874c0e735c68d46693c2f1ebceae0f appears to be the exact same as [a paragraph already in the spec a bit lower down](https://github.com/ChainAgnostic/CAIPs/blob/692847beda01c73faf91d741d30d59ef5a3b96c2/CAIPs/caip-25.md?plain=1#L76) 🤔

jiexi: does it matter if `When the wallet does not provide a `sessionId` in its initial response` that we make this distinction?

Feels like this is two separate items
1. When the wallet does not provide a `sessionId` in its initial response, subsequent `wallet_createSessions` calls target the previous singular sesssion between caller and wallet.
2. `wallet_createSession` calls either: increase scope, decrease scope, or both

ffmcgee725: @adonesky1 was unaware of this at the time, addressed here https://github.com/ChainAgnostic/CAIPs/pull/360/commits/aacdbbb91f31969f6e1ce7d0cf630736c7b87fc5 as well as @jiexi 's feedback

CAIPs/caip-25.md
```diff
@@ -46,10 +46,11 @@ Note that wallets NOT returning `sessionId`s MUST implement additional methods a
...
-- When a `sessionId` is returned in the initial `wallet_createSession` response, subsequent `wallet_createSession` calls either: 
+- When a `sessionId` is returned in the initial `wallet_createSession` response, subsequent `wallet_createSession` calls either:
   - include a previously used `sessionId` on the root of the request meaning this request is intended to modify that session, or
   - do not include a `sessionId`, in which case a new session is created - the respondent generates a new `sessionId` and sends it with the success response - and the previous session dangles in parallel (until its expiration, if applicable), though maintaining concurrent sessions is discouraged (see Security Considerations).
-- When the wallet does not provide a `sessionId` in its initial response, subsequent `wallet_createSession` calls overwrite the previous singular session between caller and wallet.
+- When the wallet does not provide a `sessionId` in its initial response, subsequent `wallet_createSessions` calls target the previous singular session between caller and wallet.
```
pedrouid:
```suggestion
- When the wallet does not provide a `sessionId` in its initial response, subsequent `wallet_createSession` calls target the previous singular session between caller and wallet.
```

CAIPs/caip-25.md
```diff
@@ -46,10 +46,11 @@ Note that wallets NOT returning `sessionId`s MUST implement additional methods a
...
-- When a `sessionId` is returned in the initial `wallet_createSession` response, subsequent `wallet_createSession` calls either: 
+- When a `sessionId` is returned in the initial `wallet_createSession` response, subsequent `wallet_createSession` calls either:
   - include a previously used `sessionId` on the root of the request meaning this request is intended to modify that session, or
   - do not include a `sessionId`, in which case a new session is created - the respondent generates a new `sessionId` and sends it with the success response - and the previous session dangles in parallel (until its expiration, if applicable), though maintaining concurrent sessions is discouraged (see Security Considerations).
-- When the wallet does not provide a `sessionId` in its initial response, subsequent `wallet_createSession` calls overwrite the previous singular session between caller and wallet.
+- When the wallet does not provide a `sessionId` in its initial response, subsequent `wallet_createSession` calls target the previous singular session between caller and wallet.
+- `wallet_createSession` calls either increase scope, decrease scope, or both.
```
adonesky1:
```suggestion
- When the wallet does not provide a `sessionId` in its initial response, subsequent `wallet_createSession` calls target the previous singular session between caller and wallet. 
 - Wallets MAY interpret the intent of subsequent scope requests as adding to or overwriting of existing permissions
 - Users MAY choose to grant a subset, superset, or entirely different set of permissions
```

adonesky1: I don't know... Something like this. WDYT @jiexi

jiexi: I think the overlapping bit is not totally correct. The additive behavior happens for us even if the request is partially overlapping. Maybe we are trying to describe two separate things as one still

* Wallets MAY interpret the intent of subsequent scope requests as adding to or overwriting of existing permissions
* Users MAY choose to grant a subset, superset, or entirely different set of permissions

adonesky1: Yeah that sounds right to me! I've modified the suggested change to match @jiexi's suggestion

adonesky1 (approved):
LGTM!

obstropolos (approved):

PR #363, opened 2025-07-01
Add CAIP-210

danroc:


CAIPs/caip-210.md
```diff
@@ -0,0 +1,93 @@
+---
+caip: CAIP-210
```
hmalik88:
```suggestion
caip: CAIP-363
```

CAIPs/caip-210.md
```diff
@@ -0,0 +1,93 @@
...
+```text
+<namespace>:_:<account-address>
+```
+
+For example:
+
+- `eip155:_:0x1234...abcd` represents the same address across all `eip155`
+  chains.
+
+- `solana:_:cosmos1abcd...` represents a Solana account across all Solana
```
hmalik88: Do you mind fixing this example to be representative of a Solana address? The cosmos bit is misleading.

danroc:
> Is address derivation assumed to be the same for all chains across the various namespaces?

No. I don't think we should make any assumptions about how the address is derived, as it might be chain or wallet-specific.

hmalik88:
> > Is address derivation assumed to be the same for all chains across the various namespaces?
> 
> No. I don't think we should make any assumptions about how the address is derived, as it might be chain or wallet-specific.

I guess my question was getting at how can this be a reliable identifier for all chains in a namespace if that assumption can't be made? (In terms of using this as a CAIP-10 extension)

danroc:
I think we should only use the address `eip155:_:0x1234...` if we can expect it to exist on any `eip155:_` chain. If not, we shouldn't use the `_` wildcard, just as in CAIP-10 where the address must be valid on the specified chain.

hmalik88:
> I think we should only use the address `eip155:_:0x1234...` if we can expect it to exist on any `eip155:_` chain. If not, we shouldn't use the `_` wildcard, just as in CAIP-10 where the address is expected to be valid on the specified chain.

But that doesn't work for other namespaces is my point, if we're reserving the `_` wildcard across namespaces then that doesn't make sense as an extension for CAIP-10. Wondering what your thoughts are @bumblefudge? Maybe we can add a caveat section that mentions this only applies for namespaces that have the same derivation scheme across its namespace?

CAIPs/caip-363.md
```diff
@@ -0,0 +1,93 @@
...
+
+- `eip155:_:0x59f3...d0c3` represents the same address across all `eip155`
+  chains.
+
+- `solana:_:DAXa...bx77` represents a Solana account across all Solana chains
+  (for example, Mainnet, Testnet, and Devnet).
+
+### Reserved Character
+
+The `_` character is reserved solely for the wildcard chain reference and MUST
```
bumblefudge:
```suggestion
A single `_` character is reserved solely for the wildcard chain reference and MUST
```

CAIPs/caip-363.md
```diff
@@ -0,0 +1,93 @@
...
+NOT be used as a valid reference or account identifier otherwise.
```
bumblefudge:
```suggestion
NOT be designated as a valid reference or account identifier in any future [CAIP-2] profile.
```

CAIPs/caip-363.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Reference Implementation
+
+No code changes are required in CAIP-2 and CAIP-10, but supporting
+implementations (e.g., wallets, dapps) should treat `_` as matching any chain
+reference within the specified namespace.
+
+## Copyright
+
+Copyright and related rights waived via CC0.
```
bumblefudge, line 98:
```suggestion
Copyright and related rights waived via CC0.

## References

- [CAIP-2]
- [CAIP-10]

[CAIP-2]: https://chainagnostic.org/CAIPs/caip-2
[CAIP-10]: https://chainagnostic.org/CAIPs/caip-10

```

bumblefudge (approved):
Looks good to me, modulo my tiny clarity edits.

bumblefudge (approved):
@obstropolos @pedrouid @lukaisailovic any objections? it's a distinct CAIP so whoever doesn't want to can just... no implement it and reject `_` chainIds

bumblefudge:
> No. I don't think we should make any assumptions about how the address is derived, as it might be chain or wallet-specific.

See, @danroc , on second glance I'm actually a little confused on the "semantics" of `_` versus `0`.  

`0` === EOA, i.e., for an address `eip155:0:0x123...` the private key controlled in an EOA derives the EVM "standard" way to that address on mainnet _and any other chain which derives its addresses the standard way_. By default, any chain X that isn't particularly special (maybe some ZKL2s still use special address types or derivations?) can assume `eip155:0:0x123...`'s private key to control `eip155:X:0x123...`, so apps would just have to keep track of which valid chainIds are !X.

What's the control/privkey/deriviation assumption for `_`?  Does the same X/!X logic/allowlist/mapping still apply? or is it just more opaque than `0`, such that you want a notation for where the EOA/privkey/derivation-function assumptions aren't known, and which chain(s) the address is valid for is also unknown?

It might be worth writing out these assumptions here in the CAIP and/or in /namespaces/eip155/caip363.md so that it's maximally clear when to use `_` versus when to use `0` in the eip155 case (and which namespaces can or can't validly have caip363 profiles)

bumblefudge:
(Also, if you think through this and no namespace other than EVM would even _need_ this notation, because its existence is necessitated by `0` and chainId complexity coming late to the party, it's also perfectly valid to just ditch the CAIP and add a paragraph under the chainId 0 paragraph in `namespaces/eip155/caip10.md`.  But if you're confident at least MM would use `_` for at least 1 other namespace, I'm convinced this CAIP is worth formalizing)

danroc:
> But if you're confident at least MM would use _ for at least 1 other namespace, I'm convinced this CAIP is worth formalizing.

We are going to use it for other namespaces too. In Solana, for example, we will use it to cover Mainnet, Testnet, and Devnet. Other chains will follow with similar use cases.

> Or is it just more opaque than 0, such that you want a notation for where the EOA/privkey/derivation-function assumptions aren't known, and which chain(s) the address is valid for is also unknown

It's indeed intended to be more opaque, and I expect that we don't make any assumptions about how the address is obtained.

WDYT of this addition:

> This CAIP makes no assumptions about how the address is derived or whether it is valid across all chains in the namespace. The use of _ simply denotes intent to refer to the same address across chains, without implying compatibility or derivation logic.

PR #366, opened 2025-07-06
CAIP-XXX: Chain-Agnostic Wallet Address Request URI Scheme

timjrobinson:
This CAIP adds a chain agnostic `wallet:address` URI for wallets, which they respond to with their CAIP-10 address. 

I created this for a real life payment terminal to be able to request via NFC the customers wallet address so that it can determine the best chain/token to send a payment request for. Without this URI the merchant is unable to know what assets a customer has to pay with and so cannot send a good payment request. 

Open Question: Should this be `wallet:addresses` and wallets can return more than one address? How should that be formatted, comma separated?

### DIFF
--- a/CAIPs/caip-xxx.md
+++ b/CAIPs/caip-xxx.md
@@ -18,7 +18,7 @@ The wallet URI scheme is defined as follows:
 
 `web3://wallet_getAccounts`
 
-When invoked by a wallet, this intent will return all wallet addresses in the CAIP-10 format:
+When invoked by an application, this intent will return all wallet addresses in the CAIP-10 format:
 
 `[namespace:reference:address]`
 


CAIPs/caip-xxx.md
```diff
@@ -0,0 +1,61 @@
...
+
+Wallets must ensure user consent when responding to wallet: URIs.
+
+Wallets should implement warnings and confirmation dialogs to prevent inadvertent exposure of wallet addresses.
+
+## Implementation
+
+Implementations should:
+
+Register `wallet:` URI handlers.
```
bumblefudge: yowza!

timjrobinson: What do you mean by this? I'd be happy to use any other chain-agnostic URI but I haven't seen one anywhere. Is there a better way?

bumblefudge: there's a couple, but it's the same amount of work distributed across difference governance surfaces, debatably 😅 

i think konrad probably already sent you this, but preserving it here for future martian anthropologists:
https://kkonrad.substack.com/p/stop-inventing-custom-uris-use-urls

let's talk tomorrow on zoom and figure out what version of this is most productive for ya!

timjrobinson:
I like the idea of using `web3://` instead. 

This is more of a call/response protocol. Also thinking it could be `getAccounts` and it returns an array of CAIP-10 formatted accounts for more flexibility and to be more similar to what walletconnect does.

lukaisailovic:
Interesting approach! Have you seen https://github.com/ChainAgnostic/CAIPs/pull/358? It handles the problem in a different way

timjrobinson:
@lukaisailovic this is great! This seems like it would be a great next step after requesting the wallet address, as then the requester could look up the senders tokens and send a CAIP-358 request for any tokens that the sender has that the requester supports. 

Without knowing what tokens the sender has available it seems like a shot in the dark with the exponentially growing number of tokens/chains out there. Which is why I think it helps a lot to gather that information first.

lukaisailovic:
@timjrobinson The goal of 358 was intentionally not to request the address beforehand. It allows the recipient to specify which tokens are accepted and then wallet to complete the request in any one of them. 

It eliminates one more round trip, which is significant for the speed of checkout.

Also we can make some assumptions, like most likely token will be stablecoin, chains etc and growing number of wallets support some aspect of Chain/Token Abstraction (swap before execute) which makes it extremely likely for wallet to complete at least one of the payment requests.

CAIPs/caip-xxx.md
```diff
@@ -0,0 +1,61 @@
...
+
+Currently there is no way for a service to request an address from a wallet in a chain agnostic way. This makes chain agnostic payments hard because the requester does not know what chains/tokens the payer supports before sending out a payment request. By requesting a wallet address the requester can look up the payers available tokens out of band (or via something like [EIP-7811][]) and then send them an appropriate chain-specific payment request. 
+
+## Specification
+
+The wallet URI scheme is defined as follows:
+
+`web3://wallet_getAccounts`
+
+When invoked by a wallet, this intent will return all wallet addresses in the CAIP-10 format:
```
bumblefudge:
```suggestion
When invoked by an application, this intent will return all wallet addresses in the CAIP-10 format:
```

CAIPs/caip-xxx.md
```diff
@@ -0,0 +1,61 @@
...
+
+Example response for Solana mainnet:
+
+`[solana:mainnet:7YhWuwZfYExVt1WkdFdHEbvhb7LgPCz4vNYBrH1N8eYQ]`
+
+If the wallet has a preferred chain for a multi-chain address it could specify that in the response, to guide the requester towards the chain the wallet would prefer. 
+
+## Rationale
+
+This allows a user to use NFC or other simple intent methods to provide their wallet address without manual copying, scanning, or complex integrations. The [CAIP-10][] standard is chosen as it clearly indicates the blockchain namespace and reference.
```
bumblefudge, line 37:
```suggestion
This allows a user to use NFC or other transports as a channel over which an application can query and a wallet can provideaddress without manually copying, scanning, or complex integrations. The [CAIP-10][] standard is chosen as it clearly indicates the blockchain namespace and reference.
```

CAIPs/caip-xxx.md
```diff
@@ -0,0 +1,61 @@
...
+Authors: Tim Robinson
+Created: 2025-07-06
+
+## Abstract
+
+This CAIP introduces a chain-agnostic URI command (`web3://wallet_getAccounts`) designed to request an array of wallet addresses from any wallet application or service. It returns an array of standardised CAIP-10 format wallet addresses, which clearly specifies the blockchain namespace, chain reference, and wallet address.
+
+## Motivation
+
+Currently there is no way for a service to request an address from a wallet in a chain agnostic way. This makes chain agnostic payments hard because the requester does not know what chains/tokens the payer supports before sending out a payment request. By requesting a wallet address the requester can look up the payers available tokens out of band (or via something like [EIP-7811][]) and then send them an appropriate chain-specific payment request.
```
bumblefudge, line 13:
```suggestion
Currently there is no way for a service to request an address from a wallet in a chain-agnostic way, except in the negotiations to open a CAIP-25 connection. This makes chain agnostic payments hard because the requester does not know what chains/tokens the payer supports before sending out a payment request. By requesting a wallet address the requester can look up the payers available tokens out of band (or via something like [EIP-7811][]) and then send them an appropriate chain-specific payment request. 
```

CAIPs/caip-xxx.md
```diff
@@ -0,0 +1,61 @@
+Title: Chain-Agnostic Wallet Address Request URI Scheme
+Type: Standard
+Status: Draft
+Authors: Tim Robinson
+Created: 2025-07-06
+
+## Abstract
+
+This CAIP introduces a chain-agnostic URI command (`web3://wallet_getAccounts`) designed to request an array of wallet addresses from any wallet application or service. It returns an array of standardised CAIP-10 format wallet addresses, which clearly specifies the blockchain namespace, chain reference, and wallet address.
```
bumblefudge, line 9: If this is an NFC-only or transport-agnostic protocol, the abstract should probably mention it!

And either way, a # Test Cases section (with both success cases and failure cases) would help, to make explicit what this query/return actually looks like as an NFC exchange, a JSON-RPC call/response, an HTTP GET/response, etc.

bumblefudge: @timjrobinson did you get any interest from implementers to take this over the line?  I feel like it's 99% of the way there, just a few gaps to make this worth publishing as a draft to leave a hook in the [more public] water[s]...

PR #330, opened 2024-09-17
Provide more guidance on restructuring scopes to disambiguate double-requests

bumblefudge:
as per yesterday's discussion in the RPC WG, I took a stab at adding a section on this corner-case to ## Security Considerations, with examples.

PR #367, opened 2025-07-31
CAIP-25: Simplified and pragmatic spec 

pedrouid:
TLDR'
- Reduced verbosity of most sections to make it easier to read
- Renamed:
    - `optionalScopes` to `scopes`
    - `scopedProperties` to `capabilities`
    - `sessionProperties` to `properties`
- Removed `requiredScopes` and retained only `scopes` (fka `optionalScopes`).
- Merged `capabilities` (fka `scopedProperties`) into `scopeObjects`.
- Removed Namespace-scoped `scopeObjects` and retained only Chain-scoped `scopeObjects`
- Remove chain prefix for accounts inside `scopeObjects`


jiexi:
thoughts on dropping the session prefix? `sessionScopes` -> `scopes`, `sessionCapabilities` -> `capabilities`?

I'm still unsure if i like the rename from properties to capabilities. Will have to think more.

While we're making changes, what are your thoughts on removing the `references` shorthand? Sure, it kind of makes sense convenience wise for humans manually forming these requests, but I don't think that's a scenario we should optimize for. A machine can easily form the correct request object without the `references` shorthand. Removing the `references` shorthand simplifies the validation and handling logic for the implementors of CAIP-25

CAIPs/caip-25.md
```diff
@@ -16,110 +16,90 @@ CAIP-25 defines an authorization procedure for a chain agnostic provider to inte
...
-It defines a lightweight protocol for negotiating and persisting authorizations during a session managed by a provider construct.
+This proposal defines a standard procedure for decentralized applications to interface with chain agnostic cryptocurrency wallets and other user agents that govern identities (including accounts) across multiple cryptographic systems. It specifies a lightweight protocol for negotiating and persisting authorizations during a session managed by a provider construct.
 
 ## Motivation
 
-The motivation comes from the lack of standardization across blockchains to expose accounts and define the expected JSON-RPC methods to be used by an application through a provider connecting to a signer or other user agent.
+The lack of standardization across blockchains for exposing accounts and defining expected JSON-RPC methods has created inconsistencies. CAIP-25 addresses this by enabling a unified, session-based interface between applications and wallets.
```
adonesky1: Maybe?
```suggestion
The absence of standardized interfaces and abstractions for reading from and writing to blockchains—such as consistent account models and JSON-RPC method specifications—has fragmented application and wallet interactions. CAIP-25 resolves this by defining a unified, session-based interface that standardizes communication between applications and wallets.
```

CAIPs/caip-25.md
```diff
@@ -16,110 +16,90 @@ CAIP-25 defines an authorization procedure for a chain agnostic provider to inte
...
-- When a wallet responds with a success response containing a `sessionId` (an entropic [identifier][CAIP-171]), the properties and authorization scopes that make up the session should be persisted and tracked over the life of the session by both parties in a discrete data store.
-- When the wallet does not provide a `sessionId` in its initial response, the wallet MUST persist and track the properties and authorization scopes that make up the session.
-The caller is not expected to persist session data or even a `sessionId`.
-Note that wallets NOT returning `sessionId`s MUST implement additional methods and notifications to handle the full lifecycle of the session:
-    * [`wallet_getSession`][CAIP-312] to enable the caller to query for the current status of the session at any time.
-    * [`wallet_revokeSession`][CAIP-285] to explicitly end the session
-    * [`wallet_sessionChanged`][CAIP-311] to notify caller of updated session authorizations.
+The session is proposed by a caller and accepted by the respondent. The respondent may return a `sessionId` which both parties then persist along with session properties and authorization scopes.
```
adonesky1: should we atleast reference [CAIP-316](https://chainagnostic.org/CAIPs/caip-316)?
```suggestion
The session is proposed by a caller and accepted by the respondent. The respondent may return a `sessionId` which both parties then persist along with session properties and authorization scopes. See [CAIP-316][] for guidance on session lifecycles with and without `sessionId`s.
```

CAIPs/caip-25.md
```diff
@@ -16,110 +16,90 @@ CAIP-25 defines an authorization procedure for a chain agnostic provider to inte
...
 ```jsonc
 {
   "id": 1,
   "jsonrpc": "2.0",
   "method": "wallet_createSession",
   "params": {
-    "requiredScopes": {
+    "sessionScopes": {
       "eip155": {
         "references": ["1", "137"],
```
adonesky1: What do we think of getting rid of this format altogether? Per @bumblefudge's language [here](https://github.com/ChainAgnostic/CAIPs/pull/330/files#diff-0864406dba09fa71f65989c7b3d779c71ee5cfc4cc39dec39c33c3eb96c8d35eR85-R86):

> Folding properties from a [CAIP-2][]-keyed, chain-specific scope object into a [CAIP-104][]-keyed, namespace-wide scope object, for example, may introduce ambiguities or violate the law of Least Privilege;
> it is safer, in such cases, to restructure in the opposite direction, i.e. from compact [CAIP-104] scopes into more verbose single-chain [CAIP-2] requests. (See Security Considerations below).

Why don't we take this opportunity with breaking changes to simplify this formatting as well?

CAIPs/caip-25.md
```diff
@@ -130,54 +110,17 @@ Example:
...
-
-- The `requiredScopes` object MUST contain 1 or more `scopeObjects`, if present.
-- The `optionalScopes` object MUST contain 1 or more `scopeObjects`, if present.
+The `sessionScopes` object MUST contain one or more scopeObjects.
 
-A third object is the `scopedProperties` object, which also MUST contain 1 or more objects if present.
-Each object should be keyed to the scope of a `sessionScopes` member to which it corresponds.
-All properties of each object in `scopedProperties` MUST be interpreted by the respondent as proposals or declarations rather than as requirements.
-In addition to making additional properties of or metadata about the corresponding `sessionScopes` member explicit, they can also annotate, support, or extend the negotiation of scope proposals (e.g., providing connection information about unfamiliar scopes, or which accounts to expose to each).
+The `sessionCapabilities` object MAY be present and provides metadata or annotations tied to specific `sessionScopes` keys.
```
adonesky1: Seems like we're removing some possibly important language around the optionality of these capabilities? And perhaps should be more specific about what kinds of annotations/metadata we would expect to be put here?

CAIPs/caip-25.md
```diff
@@ -188,193 +131,73 @@ An example of a successful response follows:
...
-        "methods": ["eth_sendTransaction", "eth_signTransaction", "get_balance", "eth_sign", "personal_sign"]
-        "notifications": ["accountsChanged", "chainChanged"],
-        "accounts": ["eip155:1:0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb", "eip155:137:0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb"]
-      },
-      "eip155:10": {
-        "methods": ["get_balance"],
-        "notifications": ["accountsChanged", "chainChanged"],
-        "accounts": []
+        "methods": ["eth_sendTransaction", "personal_sign"],
+        "notifications": ["accountsChanged", "chainChanged"]
```
adonesky1: Aren't accounts required for the response?
From one of the deleted lines above:
> Unlike the request, each scope object MUST also contain an `accounts` array,

CAIPs/caip-25.md
bumblefudge: yeah even if it's a lot of bathwater maybe we can condense the baby into a concise little paragraph or two

pedrouid:
Actually I think its a good idea to drop the references shorthand and just make it all flat into single-scope scopes

If that is the case then we can also include capabilities inside the scope and even rename it to chains

pedrouid:
Thanks again @jiexi @adonesky1 for the first round of comments!

1. Updated the "Motivation" was suggested in the comments
2. Included a reference to CAIP-316 for session lifecyscles
3. Remove Namespace-scoped in favor of Chain-scoped scopes
4. Removed `session` prefix from top-level fields
5. Merged `capabilities` into `scopedObjects`
6. Included both @jiexi and @adonesky1 as authors

CAIPs/caip-25.md
```diff
@@ -16,422 +16,247 @@ CAIP-25 defines an authorization procedure for a chain agnostic provider to inte
...
+          }
+        }
       },
       "eip155:0": {
-        "methods": ["wallet_getPermissions", "wallet_creds_store", "wallet_creds_verify", "wallet_creds_issue", "wallet_creds_present"],
-        "notifications": []
+        "methods": ["wallet_grantPermissions"],
+        "notifications": [],
+        "capabilities": {
+          "signInWithEthereum": {
```
jiexi: should this key be `eip4361` instead?

CAIPs/caip-25.md
```diff
@@ -16,422 +16,247 @@ CAIP-25 defines an authorization procedure for a chain agnostic provider to inte
...
-The JSON-RPC method is labeled as `wallet_createSession` and its `params` object contains "requiredScopes" and/or "optionalScopes" objects populated with [CAIP-217][] "scope objects" keyed to [CAIP-217][] scope strings.
-
-- The `requiredScopes` object MUST contain 1 or more `scopeObjects`, if present.
-- The `optionalScopes` object MUST contain 1 or more `scopeObjects`, if present.
-
-A third object is the `scopedProperties` object, which also MUST contain 1 or more objects if present.
-Each object should be keyed to the scope of a `sessionScopes` member to which it corresponds.
-All properties of each object in `scopedProperties` MUST be interpreted by the respondent as proposals or declarations rather than as requirements.
-In addition to making additional properties of or metadata about the corresponding `sessionScopes` member explicit, they can also annotate, support, or extend the negotiation of scope proposals (e.g., providing connection information about unfamiliar scopes, or which accounts to expose to each).
+The `scopes` object MUST contain one or more scopeObjects.
```
jiexi: i think this line highlights some of the awkwardness with the "ScopeObject" naming.

CAIPs/caip-25.md
```diff
@@ -16,422 +16,247 @@ CAIP-25 defines an authorization procedure for a chain agnostic provider to inte
...
 
-The wallet can respond to this method with either a success result or an error message.
-
 #### Success
 
-The successful result MAY contain a string (keyed as `sessionId` with a value conformant to [CAIP-171][]). As described above, if a `sessionId` is returned in the response, the caller should persist and track the properties and authorization scopes associated with this `sessionid`. If the wallet does not return a `sessionId` in the response, the connection will only consist of one session at a time, the contents of which are always retrievable for the caller via [`wallet_getSession`][CAIP-312].
-
-The successful result MUST contain an object called `sessionScopes` which contains 1 or more `scopeObjects`.
-- All required `scopeObjects` and all, none, or some of the optional `scopeObject`s (at the discretion of the provider) MUST be included if successful.
-- Unlike the request, each scope object MUST also contain an `accounts` array,
```
jiexi, line 171: is this being loosened?

CAIPs/caip-25.md
```diff
@@ -16,422 +16,247 @@ CAIP-25 defines an authorization procedure for a chain agnostic provider to inte
...
+The `scopes` object MUST contain one or more scopeObjects.
 
-A fourth object, `sessionProperties`, is optional and its shape undefined.
-It is intended for metadata or additional information not bound to any specific authorization scope, but made "global" to the connection.
-
-The respondent SHOULD ignore and drop from its response any properties not defined in this document or in another CAIP document extending this protocol which the respondent has implemented in its entirety;
-similarly, the `requiredScopes`, `optionalScopes`, and `sessionScopes` objects returned by the respondent SHOULD contain only valid [CAIP-217][] objects, and properties not defined in [CAIP-217][] SHOULD also be dropped from each of those objects.
-The same absolute security posture is not expected for the metadata objects `scopedProperties` and `sessionProperties`, but caution is still recommended for such extensions:
-callers and respondents alike SHOULD allow for their counterparties dropping or ignoring unfamiliar members from either.
```
jiexi, line 154: i feel like some of this guidance is helpful. Is it worth giving up these sections to make the spec appear simpler?

CAIPs/caip-25.md
```diff
@@ -16,422 +16,247 @@ CAIP-25 defines an authorization procedure for a chain agnostic provider to inte
...
-An empty or absent `references` array SHOULD NOT be interpreted as a namespace-wide authorization (i.e. authorization for ANY network therein), but rather as a null authorization of 0 specified `references`s within that namespace.
-(See [CAIP-217][] for more details on the structure of the typed objects included in these `...Scopes` objects.)
+Each `scopeObject` is keyed by a [CAIP-2][] chain ID.
 
-The distinction between `requiredScopes` and `optionalScopes` is ultimately semantic, since a wallet may still choose to establish a connection authorizing a subset of requested networks or requested capabilities from each; the primary function of the distinction is to offer callers a mechanism for signaling which scopes they consider primary and which they consider secondary to their request, in order to better inform the authorization logic of the respondent.
+Wallets MAY authorize a subset of scopes. This enables granular control and flexibility.
```
jiexi: We allow subset, superset, or entirely disjoint set to be granted. The dapp requested scopes are only taken as a suggestion on what defaults to show to the user. The user is free to deviate from the default as they wish.

CAIPs/caip-25.md
```diff
@@ -16,422 +16,247 @@ CAIP-25 defines an authorization procedure for a chain agnostic provider to inte
...
-(i.e., additional properties of an authorization target in `requiredScopes` may be requested in a separate `scopeObject` with the same key in the `optionalScopes` array).
+Authorization requests are expressed as a top-level object `scopes` containing keyed [scopeObjects][CAIP-217].
 
-Each `scopeObject` in these parent `...Scopes` objects can be keyed to a specific [CAIP-2][] network identifier, or to an entire [CAIP-104][] namespace.
-`scopeObjects` keyed to an entire [CAIP-104][] namespace SHOULD contain a non-empty `references` array to be actionable, making them functionally equivalent to a series of identical `scopeObjects`, each keyed to one of the members of `references` expressed as a [CAIP-2][] scope.
-An empty or absent `references` array SHOULD NOT be interpreted as a namespace-wide authorization (i.e. authorization for ANY network therein), but rather as a null authorization of 0 specified `references`s within that namespace.
-(See [CAIP-217][] for more details on the structure of the typed objects included in these `...Scopes` objects.)
+Each `scopeObject` is keyed by a [CAIP-2][] chain ID.
```
jiexi: This touches on how we don't have a good place to park methods that do not belong to any ecosystem/chain. Right now we're using the `wallet` scope for that, but it is sort of a pain as it isn't a CAIP-2 chain ID.

The other issue is that we don't have a good place to park methods/accounts that belong to an ecosystem but to no specific chain. In the example below, I see `eip155:0` used, but AFAICT the `0` reference is only meant to be used when specifying that an CAIP-10 Account ID is an EOA account. I feel that this is implied both by the wording in the [EIP-155 CAIP-10 spec](https://github.com/ChainAgnostic/namespaces/blob/main/eip155/caip10.md), but also that this specifically is specified on the CAIP-10 spec and not the CAIP-2 spec

jiexi:
adding this here before I forget. Should we revisit if `accounts` needs to be full CAIP-10` addresses since they are nested in a CAIP-2 scope? I can see an argument for and against.

CAIPs/caip-25.md
```diff
@@ -16,422 +16,247 @@ CAIP-25 defines an authorization procedure for a chain agnostic provider to inte
 
 ## Abstract
 
-This proposal has the goal to define a standard procedure for decentralized applications to interface with chain agnostic cryptocurrency wallets and other user agents which govern identities (including accounts) in multiple cryptographic systems.
-It defines a lightweight protocol for negotiating and persisting authorizations during a session managed by a provider construct.
+This proposal defines a standard procedure for decentralized applications to interface with chain agnostic cryptocurrency wallets and other user agents that govern identities (including accounts) across multiple cryptographic systems. It specifies a lightweight protocol for negotiating and persisting authorizations during a session managed by a provider construct.
```
bumblefudge:
```suggestion
This proposal defines a standard procedure for decentralized applications to interface with chain agnostic cryptocurrency wallets and other user agents that govern identities (including accounts) across multiple cryptographic systems. It specifies a lightweight protocol for negotiating and persisting authorizations during a session managed either by an in-DOM provider construct, by a securely-addressed browser extensions, or by a distinct user-agent.
```

CAIPs/caip-25.md
```diff
@@ -16,422 +16,247 @@ CAIP-25 defines an authorization procedure for a chain agnostic provider to inte
...
+Subsequent `wallet_createSession` calls may:
 
-If a connection is initially established without a `sessionId` and the wallet later implements `sessionId` support, the wallet can revoke the single session and notify the caller via `wallet_sessionChanged`. When the caller seeks to re-establish the session via `wallet_createSession`, the wallet should return a `sessionId` in the response.
+- Update an existing session by including the same `sessionId`
+- Create a new session if no `sessionId` is provided (discouraged)
 
-When a caller wishes revoke an active session, it can do so by calling [`wallet_revokeSession`][CAIP-285].
-- When a `sessionId` is returned in the initial `wallet_createSession` response, the caller MUST call `wallet_revokeSession` with the supplied `sessionId` to revoke that session.
-- When the wallet does not provide a `sessionId` in its initial response, a call to `wallet_revokeSession` revokes the single active session between caller and wallet.
+Session updates triggered by the wallet must notify the caller using `wallet_sessionChanged`.
```
bumblefudge:
```suggestion
Session updates initiated by the wallet must notify the caller using `wallet_sessionChanged`.
```

CAIPs/caip-25.md
```diff
@@ -16,422 +16,247 @@ CAIP-25 defines an authorization procedure for a chain agnostic provider to inte
...
 
-For more detail on the lifecycle and management of sessions with and without `sessionId`s, see the informational [CAIP-316][].
+Callers may revoke sessions using `wallet_revokeSession`, either by `sessionId` or directly if no sessionId was established.
```
bumblefudge:
```suggestion
Callers may revoke sessions using `wallet_revokeSession`, passing the `sessionId` parameter if it was returned by the initial response.
```

"directly" is a little confusing, implies there are 2 steps one way and 1 the other, rather than just an explanation of why the parameter is optional in the definition of `wallet_revokeSession`

CAIPs/caip-25.md
bumblefudge: Good point.  I deliberately only defined the "CAIP-10 usage" of chainId 0 at the time because it was only discussed in the context of EOAs (and the older Gnosis Safe convention), there was no evidence of 0 being used as a chainId proper in the wild.  

Adding a similar section to the EIP-155 CAIP-2 spec would work if CAIP-2 wanted to be able to use `eip155:0` specifically as a key for `scopeObject`s, for example.  If that's cleaner or more durable long-term than keeping the "wallet:eip155" special-case, let's discuss.

CAIPs/caip-25.md
bumblefudge: "MAY authorize a subset" implies a superset or disjoint set is NOT valid. Perhaps:

```suggestion
Wallets MAY authorize a subset of scopes or scope properties as requested, and MAY also authorize additional scopes or scope properties. This enables granular control and flexibility on the part of the respondent.
```

CAIPs/caip-25.md
```diff
@@ -16,422 +16,247 @@ CAIP-25 defines an authorization procedure for a chain agnostic provider to inte
...
+Wallets MAY authorize a subset of scopes. This enables granular control and flexibility.
 
-If a connection is being rejected, whether on the basis of end-user input or on the basis of evaluating `requiredScopes` against available capabilities, the respondent SHOULD choose its response based on trust:
-e.g., one or more specific failure states MAY be sent (see [#### failure states](#failure-states) below) for trusted counterparties, but an `undefined` response (or no response, depending on implementation) MAY also be sent to prevent incentivizing unwanted requests and to minimize the surface for fingerprinting of public web traffic (See Privacy Considerations below).
+Upon successful negotiation, the response includes a unified `scopes` object containing all granted scopes. Identically-keyed `scopeObjects` from multiple requests MUST be merged. No duplicate keys are allowed.
```
bumblefudge:
```suggestion
Upon successful negotiation, the response includes a unified `scopes` object containing all granted scopes. Identically-keyed `scopeObjects` from multiple requests MUST be merged. No duplicate scopes with identical keys are allowed.
```
I assume we mean syntactically/stringwise identical keys, right? In theory there could be semantically-equivalent keys, like EVM-compatible L2s that are addressable via both eip:XXX and otherthing:YYY, which we WOULD want to allow, since `otherthing:` would have different RPC calls and properties...

CAIPs/caip-25.md
bumblefudge: could go either way, or even allow both if it's written down anywhere obvious that "signInWithEthereum" is an alias for "eip4361".  i see the benefit of aligning with the Coinbase-wallet/5792 capability names and/or aliasing the more educational spec-rooted names to them.

CAIPs/caip-25.md
bumblefudge: harsh but fair

CAIPs/caip-25.md
bumblefudge: Move them to non-normative appendices, maybe? or, worst case, separate non-normative/informational CAIPs?

CAIPs/caip-25.md
```diff
@@ -16,422 +16,247 @@ CAIP-25 defines an authorization procedure for a chain agnostic provider to inte
...
-(i.e., additional properties of an authorization target in `requiredScopes` may be requested in a separate `scopeObject` with the same key in the `optionalScopes` array).
+Authorization requests are expressed as a top-level object `scopes` containing keyed [scopeObjects][CAIP-217].
 
-Each `scopeObject` in these parent `...Scopes` objects can be keyed to a specific [CAIP-2][] network identifier, or to an entire [CAIP-104][] namespace.
-`scopeObjects` keyed to an entire [CAIP-104][] namespace SHOULD contain a non-empty `references` array to be actionable, making them functionally equivalent to a series of identical `scopeObjects`, each keyed to one of the members of `references` expressed as a [CAIP-2][] scope.
-An empty or absent `references` array SHOULD NOT be interpreted as a namespace-wide authorization (i.e. authorization for ANY network therein), but rather as a null authorization of 0 specified `references`s within that namespace.
-(See [CAIP-217][] for more details on the structure of the typed objects included in these `...Scopes` objects.)
+Each `scopeObject` is keyed by a [CAIP-2][] chain ID.
```
pedrouid: Let's add a small note here about null vales like `eip155:0`

CAIPs/caip-25.md
```diff
@@ -16,422 +16,247 @@ CAIP-25 defines an authorization procedure for a chain agnostic provider to inte
...
+          "atomicBatch": "true"
+        }
       },
-      "eip155:10": {
-        "methods": ["get_balance"],
-        "notifications": ["accountsChanged", "chainChanged"]
+      "eip155:42161": {
+        "methods": ["eth_sendTransaction", "personal_sign", "wallet_sendCalls"],
+        "notifications": ["accountsChanged", "chainChanged"],
+        "capabilities": {
```
pedrouid: 1. Remove capabilities on request
2. Adding properties inside scopes for both request and response

CAIPs/caip-25.md
jiexi: i'm also fine with not using specs as the name of the field, but i'd prefer clearer guidance from us on how to extend this in the future? Do we require an information CAIP for each capability?

CAIPs/caip-25.md
jiexi: This wording works for me

CAIPs/caip-25.md
jiexi: I'm fine going down the `eip155:0` route and deciding null identifier on a per namespace basis. At least it seems we were happy with that approach during the CASA call.

This doesn't resolve the plain `wallet` scope for where we should park methods that do not belong to any chain. @pedrouid had suggested these methods don't belong in the scopes object. I'm still unsure about how I feel moving the `wallet` scope out of the scopes object and where it be moved

pedrouid:
> adding this here before I forget. Should we revisit if `accounts` needs to be full CAIP-10` addresses since they are nested in a CAIP-2 scope? I can see an argument for and against.

I've updated the example response to reflect this

CAIPs/caip-25.md
```diff
@@ -108,6 +112,8 @@ If a connection is rejected, the wallet MAY respond with a generic error or sile
 }
 ```
 
+The `wallet` object MUST be included for wallet-level method and notification requests.
```
adonesky1: I'm not totally clear what this means. Must it be included if there are no `wallet-level` methods or notifications?

adonesky1: Maybe?
```suggestion
The `wallet` object MUST be included for wallet-level methods and notifications.
```

CAIPs/caip-25.md
```diff
@@ -16,422 +16,229 @@ CAIP-25 defines an authorization procedure for a chain agnostic provider to inte
...
   "jsonrpc": "2.0",
   "method": "wallet_createSession",
   "params": {
-    "requiredScopes": {
-      "eip155": {
-        "references": ["1", "137"],
-        "methods": ["eth_sendTransaction", "eth_signTransaction", "eth_sign", "get_balance", "personal_sign"],
-        "notifications": ["accountsChanged", "chainChanged"]
+    "scopes": {
+      "wallet:any": {
```
jiexi: `any` is probably a safe identifier since I doubt there would be a ecosystem calling itself `any`. Is there any other ones we can think of that would also be CAIP-2 compliant? I remember @bumblefudge throwing out `wallet:wallet` before. I know it sounds silly, but the namespace `wallet` is reserved already for us, so this is 100% a safe reference value

CAIPs/caip-25.md
```diff
@@ -16,422 +16,229 @@ CAIP-25 defines an authorization procedure for a chain agnostic provider to inte
...
-      "expiry": "2022-11-31T17:07:31+00:00",
-      "globalConfig": {
-          "foo": "bar"
+    "properties": {
+      "expiry": "2022-12-24T17:07:31+00:00",
+      "walletInfo": {
+        "uuid": "350670db-19fa-4704-a166-e52e178b59d2",
+        "name": "Example Wallet",
+        "icon": "data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg'/>",
+        "rdns": "com.example.wallet"
```
jiexi, line 343: i like this over here a lot more than in the `wallet` scope

CAIPs/caip-25.md
```diff
@@ -16,422 +16,229 @@ CAIP-25 defines an authorization procedure for a chain agnostic provider to inte
...
+        "capabilities": {
+          "walletService": "https://wallet-service.example.com/rpc"
+        }
+      },
+      "eip155:1": {
+        "accounts": ["0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb"],
+        "methods": ["eth_sendTransaction", "personal_sign"],
         "notifications": ["accountsChanged", "chainChanged"],
-        "accounts": ["eip155:1:0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb", "eip155:137:0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb"]
+        "capabilities": {}
```
adonesky1: I left this comment somewhere earlier on this PR (seems to have gotten lost?) but `accounts` should be required in each scopeObject here... with the possible exception of the new `wallet:any` scope

adonesky1: oh wait nevermind I see they are here now, just moved... I'm just being dense. Do we need to specify somewhere how/why the `wallet:*` scopes don't need accounts though?

CAIPs/caip-25.md
```diff
@@ -16,422 +16,229 @@ CAIP-25 defines an authorization procedure for a chain agnostic provider to inte
...
 }
 ```
 
-#### Failure States
-
-The response MUST NOT be a JSON-RPC success result in any of the following failure states.
+Each entry within `scopes` object MAY contain `accounts` and `capabilities` as part of its object for success response.
 
-##### Generic Failure Code
+#### Error
```
adonesky1: ?
```suggestion
#### Error Codes
```

CAIPs/caip-25.md
pedrouid: Agreed

CAIPs/caip-25.md
pedrouid: glad to hear that

CAIPs/caip-25.md
pedrouid: Do accounts really need to be required?

bumblefudge: couldn't `accounts` be a required array but one for which an empty array is a valid value? that feels more explicit, like "I AM GIVING YOU ZERO ACCOUNTS FOR THIS" 😅 

(and, sidenote, might a wallet namespace be the right place to stick an ephemeral/implicit/offchain-only address used for completely off-node usecases like presenting verifiable credentials? a girl can dream...)

CAIPs/caip-25.md
bumblefudge: i don't even remember throwing it out, haha. 

but as [currently defined](https://namespaces.chainagnostic.org/wallet/caip2#semantics), `wallet:any` would be a crazy thing to authorize, because it is basically saying "free free to throw at my the offchain RPC methods of ANY namespace in the CASA registry" 😅 .  "wallet:eip155" or "wallet:solana" might be better as an illustrative case

CAIPs/caip-25.md
bumblefudge: (happy to do this in follow-on PRs if it's more practical/political/timely to plant the simplified-CAIP-25 flag ASAP and tack on non-normative stuff later)

CAIPs/caip-25.md
bumblefudge: `eip155:any` would also be good to include in examples

jiexi (approved):

PR #369, opened 2025-08-04
CAIP-369 - Wallet Session Property for Personal Information

pedrouid:
This CAIP introduces a new session property for use within CAIP-25 session negotiations, enabling dApps to request structured personal information from users through wallets:
- firstName
- lastName
- emailAddress
- phoneNumber
- shippingAddress
- billingAddress


bumblefudge:
Some people from fintech/wallet companies at DIF have been collaborating on a "sensible default" set of schemata for KYC info, complete with semantics, might be helpful here:
https://identity.foundation/credential-schemas/#verified-person-schema

Note in particular that the convention in US fintech/compliance is to refer to addresses compressed into single strings/fields as `unstructuredAddress`es, as most backends use the "street address"+"street address part 2(optional)"+"city"+"state/province"+"zip code/mailing code" that's default in US apps...

pedrouid:
Fair enough! Personally do not have strong opinions on this particular approach but I believe that inevitably we will get to a point where this information will be provided by wallets

Travel Rule is one example where this might be useful but there are also several centralized solutions currently exploring a similar approach to allow ecommerce checkout experiences with self-custodial wallets

I would recommend us finding something that we feel comfortable to allow wallets to expose this information without requiring a centralized third-party

kdenhartog:
Yeah, I think it's going to be necessary it's more about when more than anything IMO.

Let's let this sit for a bit, because if we can make it so that wallet addresses either become ephemeral or are blinded by ZK proofs of knowledge in some way, then that is the time where it makes sense IMO.

Having something sitting though at least gives people the idea to look in this direction when you're talking with them. As more companies start thinking about these problems then we can see what the consensus direction to go is because I am a fan of trying to combine it for better UX, I just don't see a way to make the compliance simple quite yet.

What if we just merge this as is in draft form so it's published and you can link to it as needed?

Dustin4444 (approved):

bumblefudge:
Not urgent? Awaiting implementer/evaluator feedback/enthusiasm to devote cycles to fixing the privacy approach

Dustin4444 (approved):

Dustin4444 (approved):

PR #370, opened 2025-08-05
add privacy considerations for CAIP-358

kdenhartog:
some basic privacy considerations I promised to submit awhile ago

### DIFF
--- a/CAIPs/caip-358.md
+++ b/CAIPs/caip-358.md
@@ -241,7 +241,22 @@ It also does not attempt to provide dispute functionality. These present ideas f
 
 ## Privacy Considerations
 
-TODO
+### Wallet Address Sharing
+
+Wallet addresses were intentionally left off here both for a purpose of UX simplicity as well as privacy. By opting to limit the usage of a wallet address,
+we make it so this API can be implemented without first needing to request permission for the user's wallet address. The wallet address acts as a cross-origin identifier
+which can be used to link a user's financial transactions across sites.  Since the wallet address is not needed, we can leave this up to the wallet to address this. 
+Therefore, it's the responsibility of the wallet to request from the user which wallet or token they wish to make a payment from.
+
+### Transaction Privacy
+
+Wallets are encouraged to utilize transaction privacy protocols to prevent payment data from leaking browsing history onchain. A sufficient transaction privacy protocol
+are expected prevent someone from viewing the tranasaction on a block explorer from identifying either the sender, the recipient, or both parties having transacted. A
+protocol which focuses only on the sender's privacy will prevent leaking of purchase data being used to build a behavioral profile through purchase history of an onchain account.
+A protocol which focuses only on recipient privacy will prevent leaking real-time transaction data of businesses potentially not looking to let their competitors or investors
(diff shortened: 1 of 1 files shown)


CAIPs/caip-358.md
```diff
@@ -241,7 +241,22 @@ It also does not attempt to provide dispute functionality. These present ideas f
 
 ## Privacy Considerations
 
-TODO
+### Wallet Address Sharing
+
+Wallet addresses were intentionally left off here both for a purpose of UX simplicity as well as privacy. By opting to limit the usage of a wallet address,
+we make it so this API can be implemented without first needing to request permission for the user's wallet address. The wallet address acts as a cross-origin identifier
+which can be used to link a user's financial transactions across sites.  Since the wallet address is not needed, we can leave this up to the wallet to address this. 
+Therefore, it's the responsibility of the wallet to request from the user which wallet or token they wish to make a payment from.
```
lukaisailovic: Would change the wording here. Wallets are not required to request this info every time they receive the RPC. I would assume that wallets will have some defaults set up

kdenhartog: I can update, but not sure I understand what you have in mind. Could you describe your use case you have in mind further or offer some suggested text?

In my mind, a consent screen will always need to be displayed on the wallet side in order to conduct the spend. I don't think its a good idea for a spend to be done automatically (at least not to start until wallets have heuristics to detect bad spend patterns) otherwise sites will introduce dark patterns to conduct spending under limits so that a consent pop up doesn't display and collect more money.

E.g. if user sets a limit to max spend 10 USDC as long as transaction is less than .25USDC per transaction, then if a semi trusted site wanted to get 2.50 (e.g trusted in the sense they get a $10 limit, but not trusted enough to get more than .25 transaction limit), the site would just request .25 10x rather than $2.50 in a single TXN to avoid the prompt.

I'm not sure I understand the UX you're thinking about to make sure it aligns with what you have in mind and addressing concerns like what I'm thinking about.

lukaisailovic: I don't think the spend should be done automatically, but I think a lot of the choices can be done on behalf of the user. For example, lets say you are given the option to pay in USDC, USDT, both on OP and mainnet. 
The wallet doesn't always have to ask you which token do you want, it will already pre-fill a lot of the stuff based on the knowledge of your assets, gas costs etc.

I think that in 90% of the cases, the user should only click "Approve" and nothing else. 


> Therefore, it's the responsibility of the wallet to request from the user which wallet or token they wish to make a payment from.

This is the part that I would change specifically. Maybe even switching out the word `request` to `determine`

kdenhartog: Ahh, ok yeah makes sense. Your suggestion makes sense to me to align expectations better.

kdenhartog:
```suggestion
Therefore, it's the responsibility of the wallet to determine if possible which wallet or token they wish to make a payment from. This may be done automatically to improve the user experience or allowing the user to select and override assumed defaults.
```

bumblefudge (approved):
Took the liberty of redoing line breaks and wordsmithing a bit, but no substantial changes (non-normative section anyways)

kdenhartog:
Thanks, I wasn’t sure what standard linting was so just yolo’d it.

PR #371, opened 2025-08-08
CAIP-27: Add capabilities when invoking methods

pedrouid:
Related to #367 

This change allows capabilities discovered through a [CAIP-25][] session exposed by the wallet to be invoked with this method

### DIFF
--- a/CAIPs/caip-27.md
+++ b/CAIPs/caip-27.md
@@ -81,7 +81,7 @@ The JSON-RPC method is labeled as `wallet_invokeMethod` and expects three parame
 - **request** (required) - an object containing the fields:
   - **method** (required) - the JSON-RPC method to invoke (previously authorized for the targeted network)
   - **params** (required) - JSON-RPC parameters to invoke (may be empty but must be set)
-- **capabilities** (optional) - an object containing metadata discovered through a [CAIP-25][] session of capabilities exposed by the wallet to be invoked with this method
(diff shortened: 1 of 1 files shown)


CAIPs/caip-27.md
```diff
@@ -72,6 +81,7 @@ The JSON-RPC method is labeled as `wallet_invokeMethod` and expects three parame
 - **request** (required) - an object containing the fields:
   - **method** (required) - the JSON-RPC method to invoke (previously authorized for the targeted network)
   - **params** (required) - JSON-RPC parameters to invoke (may be empty but must be set)
+- **capabilities** (optional) - an object containing metadata discovered through a [CAIP-25][] session of capabilities exposed by the wallet to be invoked with this method
```
adonesky1:
```suggestion
- **capabilities** (optional) - an object containing metadata pertaining to a capability announced by the wallet in the [CAIP-25][] return object, to be invoked with this method
```

bumblefudge: i like Al's suggestion, but even re-reading the whole thing (both example and spec-text), it's not exactly clear to me what the capability adds-- does including, e.g., the paymaster cap mean the app is asking the wallet to use it? what does it mean if the app sends a subset of the caps discovered in CAIP-25, or a cap whose relevance to the method being called is unclear, or a patently irrelevant cap?  i feel like this change introduces a lot of corner-cases and guesswork without some logic (at least in spec-text) explaining what the app can decide, and what the wallet does if it can't understand (or chooses to reject) the decisions of the app!

bumblefudge (approved):

obstropolos (approved):

PR #373, opened 2025-08-18
CAIP-373 EVM Smart Contract Call Reference Specification

davidfurlong:
This CAIP defines a way to reference objects stored in EVM smart contract mappings via function calls, because developers on [Ethereum Comments Protocol](https://ethcomments.xyz) want to be able to [Quote-post](https://github.com/ecp-eth/ECPIP/discussions/2) using an app-agnostic identifier that references the onchain comment, which is stored in a mapping.


CAIPs/caip-XXX.md
```diff
@@ -0,0 +1,174 @@
...
+
+The `function_call_reference` is a case-sensitive string in the form:
+
+```
+function_call_reference:   account_id + ":" + method + ":" + function_data + [":" + block_number]
+account_id:                chain_id + ":" + account_address (See CAIP-10)
+chain_id:                  [-a-z0-9]{3,8}:[-_a-zA-Z0-9]{1,32} (See CAIP-2)
+account_address:           0x[a-fA-F0-9]{40}
+method:                    "call"
+function_data:             0x[a-fA-F0-9]*
```
bumblefudge: @pedrouid might have thoughts on whether a truly open-ended CALLDATA field makes sense for wallet connect's usage of CAIP-19 over the years, although of course compatibility with existing CAIP-10/19 systems is not a requirement of new CAIPs!

CAIPs/caip-XXX.md
```diff
@@ -0,0 +1,174 @@
...
+block_number:              [0-9]+ (optional)
+```
+
+Note: Unlike other CAIP specifications, `function_data` has no fixed maximum length as ABI-encoded function parameters can vary significantly in size.
```
bumblefudge:
```suggestion
Note: Unlike other CAIP specifications, `function_data` has no fixed maximum length as real-world usage of on-chain function calls (e.g. ABI-encoded function parameters, in EVM-based contexts) can vary significantly in size.
```

CAIPs/caip-XXX.md
```diff
@@ -0,0 +1,174 @@
...
+This is a deliberate design trade-off to accommodate the full range of possible function calls.
+
+The `block_number` is optional and when omitted, the function call will be executed against the latest block.
+When specified, the function call will be executed against the specified block number for historical data access.
```
bumblefudge:
```suggestion
When specified, the function call will be executed against the specified block number for historical data access in VMs/namespaces where this is supported. See per-namespace profiles for syntax, validation rules, etc. of this segment.
```

CAIPs/caip-XXX.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Semantics
+
+- `account_id`: Identifies the smart contract using CAIP-10 format (e.g., `eip155:1:0x6B175474E89094C44Da98b954EedeAC495271d0F`)
+- `method`: Specifies the method type - currently only "call" is supported for EVM chains
+- `function_data`: The complete ABI-encoded function call including function selector and parameters
```
bumblefudge:
```suggestion
- `function_data`: The complete parameters passed to the target address, exactly as passed (i.e., in EVM contexts, fully ABI-encoded, including the function's selector and concatenated parameters)
```
this could work for non-EVM L1s as well! we should probably get someone with bitcoin experience to review this, to think through, e.g., Taproot corner cases

CAIPs/caip-XXX.md
```diff
@@ -0,0 +1,174 @@
...
+- `block_number`: Optional block number for historical data access (e.g., `12345678`). When omitted, the function call executes against the latest block.
+
+This extends CAIP-10 by appending `:method:function_data[:block_number]` to a standard CAIP-10 account identifier.
```
bumblefudge:
```suggestion
This extends CAIP-10 by appending `:{method}:{function_data}[:{block_number}]` to a standard CAIP-10 account identifier.
```

CAIPs/caip-XXX.md
```diff
@@ -0,0 +1,174 @@
...
+
+1. 4-byte function selector (first 4 bytes of keccak256 hash of function signature)
+2. ABI-encoded parameters according to function signature
+
+### Reference Resolution
+
+Applications that need to resolve these references can:
+
+1. Parse the reference to extract account_id, method, function_data, and optional block_number (where account_id follows CAIP-10 format)
+2. Validate the chain_id exists and is accessible
```
bumblefudge:
```suggestion
2. Validate the `chain_id` exists and is accessible
```

bumblefudge (changes requested):
you don't have to take my exact suggestions but making changes along similar lines would be g2g AFAIC

davidfurlong:
Great! Thanks for the suggestions @bumblefudge, I've committed them all

bumblefudge (approved):

PR #375, opened 2025-08-25
CAIP-375 - Wallet Sign Message

pedrouid:
Introducing a chain-agnostic RPC method `wallet_signMessage` for signing messages

### DIFF
--- a/CAIPs/caip-375.md
+++ b/CAIPs/caip-375.md
@@ -81,7 +81,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 **Rules:**
 
-- `sessionId` is OPTIONAL but SHOULD follow CAIP-171 if provided.
+- `sessionId` is OPTIONAL but MUST follow CAIP-171 if provided.
 - Each message MUST include `messageType` and `content`.
 - Wallets MAY choose any of the provided `signatureTypes`.
 - Response MUST include `account` and `signatureType` for each signature


CAIPs/caip-375.md
```diff
@@ -0,0 +1,116 @@
...
+
+- Malicious apps can trick users into signing harmful messages; wallets MUST show clear prompts including `content` and `account`.
+- Including nonces or timestamps in `content` is RECOMMENDED to prevent replay attacks.
+- Sessions referenced by `sessionId` SHOULD be validated to ensure scope compliance.
+
+## Privacy Considerations
+
+- Signing may reveal account addresses; wallets SHOULD only return requested data.
+- `capabilities` could expose metadata; apps and wallets SHOULD handle them carefully.
+- Multi-message requests could link identities; wallets MAY warn users.
```
kdenhartog, line 100: ++ this covers the general set of concerns I'd have around sec/privacy.

We may want to include more details in a follow up PR, but don't think that should block this from merging.

CAIPs/caip-375.md
```diff
@@ -0,0 +1,116 @@
...
+  "jsonrpc": "2.0",
+  "method": "wallet_signMessage",
+  "params": {
+    "sessionId": "0xdeadbeef", // optional
+    "messages": [
+      {
+        "account": "eip155:1:0xabc123...", // optional
+        "signatureTypes": ["eip191", "eip1271"], // optional
+        "messageType": "ethPersonalSign",
+        "content": "Hello World"
```
kdenhartog, line 54: For eip-712 would we want this to be a JSON object or stringified JSON object?

CAIPs/caip-375.md
```diff
@@ -0,0 +1,116 @@
...
+      }
+    ],
+    "capabilities": {} // optional
+  }
+}
+```
+
+**Rules:**
+
+- `sessionId` is OPTIONAL but SHOULD follow CAIP-171 if provided.
```
kdenhartog:
```suggestion
- `sessionId` is OPTIONAL but MUST follow CAIP-171 if provided.
```

Unless we know an explicit reason not to, it's better to just force this. It closes down a code path that we probably don't need since people can opt out via the OPTIONAL aspect

CAIPs/caip-375.md
```diff
@@ -0,0 +1,116 @@
...
+- Each message MUST include `messageType` and `content`.
+- Wallets MAY choose any of the provided `signatureTypes`.
```
kdenhartog, line 86: nit: Would it be useful to include a reference table of known signature types just for easier discovery when reading the spec?

kdenhartog (approved):
Overall, this looks like it's in a mergable state right now IMO.

Most of this is documenting a defacto abstraction layer above current wallet sign RPC calls.

Also, I wasn't sure where the best place to add this context, but I'd like to discourage usage of these lower level APIs, if there's higher level APIs available since that will solve for the security consideration around parsing and displaying the contents in a readable way.

CAIPs/caip-375.md
bumblefudge: i have no idea but seems worth specifying either way

bumblefudge:
> Overall, this looks like it's in a mergable state right now IMO.

I agree although the ERC-20 thing is probably worth deciding at least tentively before publishing?

bumblefudge (approved):
> Overall, this looks like it's in a mergable state right now IMO.

I agree although the ~~ERC-20~~ _(I MEAN [ERC-721](https://github.com/ChainAgnostic/CAIPs/pull/375#discussion_r2305447716) !!)_ thing is probably worth deciding at least tentively before publishing?

@pedrouid ?

CAIPs/caip-375.md
bumblefudge: @kdenhartog -- pedro lobbied heavily on today's call to stringify to avoid every implementation having to deal with arbitrary size and shaped messages, so stringify it is!

PR #380, opened 2025-10-01
CAIP-XXX - Portable Proof

cl34l:
A 6-field, EIP-191–signed, canonical-JSON envelope anchored by a SHAKE-256 qHash. Apps validate once off-chain; the same qHash can optionally be surfaced on-chain for indexing/transport (complements ERC-7683 vouchers).

### DIFF
--- a/CAIPs/caip-xxx.md
+++ b/CAIPs/caip-xxx.md
@@ -1,0 +1,137 @@
+---
+caip: CAIP-X
+title: Portable Proof
+author: Chris Leal (@cl34l)
+status: Draft
+type: Standard
+created: 2025-10-01
+updated: 2025-10-01
+requires: [2,10]
+---
+
+## Simple Summary
+
+A canonical, chain-agnostic, wallet-signed envelope that applications validate once off-chain and can reference anywhere via a deterministic `qHash` anchor. Identities follow CAIP-10 (`did:pkh`) and chains follow CAIP-2.
+
+## Abstract
+
+Defines a chain-agnostic, wallet-signed proof object anchored by a SHAKE-256 digest (`qHash`). Applications validate once off-chain; the same `qHash` may be surfaced on-chain for indexing/transport via EIP-7683–compatible vouchers.
+
+## Motivation
+
+Establish a canonical, deterministic envelope that can be validated once and referenced anywhere (off-chain or on-chain) without binding to any vendor, domain, or transport. Align identities with CAIP-10 and chain context with CAIP-2 to avoid namespace collisions.
+
+## Rationale
+
+- Use EIP-191 for broad wallet compatibility; support EIP-1271 and detect EIP-6492 to cover smart accounts and pre-deploy signatures.
+- Keep `qHash` anchored to a canonical subset to ensure stable, cross-environment equivalence and idempotency.
+- Treat vouchers as optional EIP-7683–compatible artifacts keyed by `qHash` to enable cross-chain transport without constraining settlement designs.
+
+## Specification
+
+- **Terminology:**
+  - Envelope: the signed data structure defined in this document
+  - Anchor: `qHash` computed over the canonical subset
+  - Verifier: logic module identified by `verifierId`
+  - Voucher: optional on-chain artifact (EIP-7683–compatible) keyed by `qHash`
+
+- **Identifiers:**
+  - Chains: `eip155:<chainId>` per CAIP-2 (numeric `chainId` on-wire)
+  - Accounts: `did:pkh:eip155:<chainId>:<address>` per CAIP-10
+- **Envelope (top-level):**
+  ```json
+  {
+    "did": "did:pkh:eip155:1:0xabc000000000000000000000000000000000def0",
+    "qHash": "0x<shake256-hex>",
+    "verifierIds": ["ownership-basic"],
+    "data": {},
+    "signature": "0x…",
+    "signedMessage": "Portable Proof Verification Request\nWallet: 0xabc000000000000000000000000000000000def0\nChain: 1\nVerifiers: ownership-basic\nData: {}\nTimestamp: 1730000000000",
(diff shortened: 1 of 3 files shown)


CAIPs/caip-380.md
```diff
@@ -0,0 +1,178 @@
...
+
+- Use [EIP-191] for broad wallet compatibility; support [EIP-1271] and detect [EIP-6492] to cover smart accounts and pre-deploy signatures.
+- Keep `qHash` anchored to a canonical subset to ensure stable, cross-environment equivalence and idempotency.
+- Treat vouchers as optional [EIP-7683]–compatible artifacts keyed by `qHash` to enable cross-chain transport without constraining settlement designs.
+
+## Specification
+
+- **Terminology:**
+  - Envelope: the signed data structure defined in this document
+  - Anchor: `qHash` computed over the canonical subset
```
bumblefudge: i assume that by canonical subset you mean a core set of top-level properties, but spelling this out more explicitly (and also being explicit about exactly how many and which properties/shapes/objects can be added _after_ computing the digest) would probably help clarify this for implementers and evaluators

CAIPs/caip-380.md
```diff
@@ -0,0 +1,178 @@
...
+    "signedTimestamp": 1730000000000,
+    "chainId": 1,
+    "signatureMethod": "eip191",
+    "options": {},
+    "meta": {}
+  }
+```
+
+- **Canonical signing (MUST):** [EIP-191] (also known as `personal_sign`). Exact six-line signer string structure (line 1 is a fixed context label; default label shown):
+  - Freshness: reject if older than 5 minutes or >1 minute in the future
```
bumblefudge: are these sub-requirements also hard MUSTs?

CAIPs/caip-380.md
```diff
@@ -0,0 +1,178 @@
...
+  - Smart-accounts: support [EIP-1271];
+  - detect [EIP-6492] wrappers.
```
bumblefudge: ^ a little more detail here might help!

CAIPs/caip-380.md
```diff
@@ -0,0 +1,178 @@
...
+
+```sh
```
bumblefudge: my markdown linter was requiring a language here, feel free to put a better one than shell formatting if you can think of one

CAIPs/caip-380.md
```diff
@@ -0,0 +1,178 @@
...
+  Portable Proof Verification Request
+  Wallet: 0x<lowercased address>
+  Chain: <numeric chain id>
+  Verifiers: <comma-separated verifier ids>
+  Data: <deterministic JSON of envelope.data>
+  Timestamp: <unix ms>
+```
+
+- **Deterministic JSON (MUST):**
+  - key-sorted objects;
```
bumblefudge: does this mean the contents of any objects (and/or of top-level properties?) need to be lexigraphically sorted by key?  the more explicit you can get on canonicalization the better!

CAIPs/caip-380.md
```diff
@@ -0,0 +1,178 @@
...
+  - omit `undefined`;
```
bumblefudge: i.e. no properties (in or outside the canonical subset) can have a value of `undefined`?

CAIPs/caip-380.md
```diff
@@ -0,0 +1,178 @@
...
+  - preserve `null`;
```
bumblefudge: same, specify if only canonical subset or anywhere in the envelope

bumblefudge (approved):
Great stuff.  I left some nits/pointers for maximum clarity (and, sadly, verbosity), but none of them are really blocking changes to be merged as "Draft".

CAIPs/caip-380.md
chrisjleal: Added ‘Canonical Subset (Normative)’ enumerating the five fields (did, verifierIds, data, signedTimestamp, chainId) and an ‘Extensibility’ note; validators ignore non-canonical props when hashing.

CAIPs/caip-380.md
chrisjleal: Added RFC 2119/8174 note and made ‘Canonical Signing’ explicit: all sub-requirements are MUST unless stated otherwise.

CAIPs/caip-380.md
chrisjleal: Added a normative verification algorithm, chain context (isValidSignature on chainId), magic value 0x1626ba7e, and unwrap/validate for 6492.

CAIPs/caip-380.md
chrisjleal: All fences labeled (abnf, json, text); signer template now abnf.

CAIPs/caip-380.md
chrisjleal: Specified scope (Canonical Subset + data only), sorting ‘lexicographically by Unicode code point, 'undefined' forbidden and omitted, null preserved, UTF-8, no insignificant whitespace.

CAIPs/caip-380.md
chrisjleal: Only the Canonical Subset and data are canonicalized; undefined is prohibited in those. Extensions are outside the anchor.

CAIPs/caip-380.md
chrisjleal: This should be fixed as well for clarity.

bumblefudge:
Non-blocking curiosity question, as a n00b to intents stuff-- is there an overlap with ERC-7683? Could a valid CAIP-380 message also be a valid ERC-7683, or is it worth adding an example (normative or non-normative) about how to wrap one in the other, or reference once from the other?

bumblefudge (approved):
I'd love to see confirmation from Solana people (or any other 2nd VM) before this goes to final, but definitely complete as a draft imho!

PR #385, opened 2025-11-07
Expanding CAIP10, CAIP19 charset to include underscores

tristan-bp:
I would like to expand the charset for account addresses in CAIP 10 and asset references / token ids in CAIP 19 to include underscores (`_`).

Underscores are used in blockchain addresses (example: telegram's [ton](https://ton.org/)) and typically don't require URL encoding for security purposes.


curtisdow1973-sys (approved):

tristan-bp:
Ok, that makes sense, thank you!

PR #390, opened 2025-12-22
CAIP-390: Minimal Cross-Chain Asset Metadata Standard

zensh:


bumblefudge:
Hey!  This is promising, and I like the basic idea.  I have some questions, although none of them have a right answer or even one I prefer:

- is an array of strings enough for "locations"?  I wonder if people might prefer some kind of ordering or ranking (JSON array order is famously unstable across languages and type systems), or some other kind of metadata (are some locations "preferred" or "canonical"? are some specifically non-canonical/secondary?)
- can non-https urls go in `external_url`? for example, an [IPFS native URL](https://docs.ipfs.tech/how-to/address-ipfs-on-web/#native-urls), or an ENS-based mutable pointer?  are there any security considerations about bare-IPv4 addresses or non-ICANN addresses? 

@pelle and @ajunge from Nota Bene have a similar use case and it might be looking at a system they [proposed a year or two ago](https://github.com/ChainAgnostic/namespaces/pull/113) for having a canonical registry for metadata based on the more free-form, human-readable namespaces documentation. Maybe there's some overlap there? At least worth reviewing as adjacent prior art.

obstropolos (approved):
It's a great way to consolidate this sort of thing if adopted / make it nice and clean. 

I'm fine approving as a draft for now.

bumblefudge (approved):
well, no word from the NB crew, let's just merge as draft and they can open PRs later if they want to align this with existing work in prod or otherwise lobby for changes!

PR #392, opened 2025-12-28
feat: added examples, added IANA requirement, fixed spelling issues

PatrickAlphaC:
I don't think 20, 21, 22, are required for this CAIP to go through. But I could be wrong. Let me know what you all think. I have also applied to register caip with IANA with this PR.

### DIFF
--- a/CAIPs/caip-322.md
+++ b/CAIPs/caip-322.md
@@ -6,17 +6,17 @@ discussions-to: https://github.com/ChainAgnostic/CAIPs/issues/67
 status: Draft
 type: Standard
 created: 2024-08-08
-requires: 2, 10, 19, 20, 21, 22, 104
+requires: 2, 10, 19, 104
 ---
 
 ## Simple Summary
 
 Various CAIPs to date have defined generic chain-agnostic identifier schemes for certain informational primitives common to blockchain and other decentralized peer-to-peer systems.
-This URI scheme combines them into a structure that can be parsed heirarchically.
+This URI scheme combines them into a structure that can be parsed hierarchically.
 
 ## Abstract
 
-A general-purpose heirarchical URI scheme for identifiers that dereference to "on-chain" or otherwise cryptographically self-certifying records from peer-to-peer cryptographic systems.
+A general-purpose hierarchical URI scheme for identifiers that dereference to "on-chain" or otherwise cryptographically self-certifying records from peer-to-peer cryptographic systems.
 The model is not equally amenable to all native identifier schemes but is engineered to facilitate translation between, and access to, those systems as broadly as possible.
 
 ## Motivation
@@ -36,7 +36,7 @@ caip:<[CAIP-104] namespace>:<[CAIP-2] chain identifier>:<{[CAIP-10]|[CAIP-19]} o
 
 The top-level segment refers to short ASCII strings identifying entries in the Chain-Agnostic Namespace registry.
 The applicability of other CAIPs, as well as any namespace-specific constraints, validation syntax, and caveats, are defined in entries there.
-All segments after the namespace are optional but heirarchical, i.e., a [CAIP-2][] chain identifier without a preceding namespace segment is invalid, and a [CAIP-10][] account identifier without a preceding chain identifier is invalid.
+All segments after the namespace are optional but hierarchical, i.e., a [CAIP-2][] chain identifier without a preceding namespace segment is invalid, and a [CAIP-10][] account identifier without a preceding chain identifier is invalid.
 
 ## Rationale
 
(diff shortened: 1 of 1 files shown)


bumblefudge:
Hey!  This is awesome, thanks for iterating the old draft (and implicitly championing it).  If you have some thoughts on use-cases that might help.

For me, the one open question preventing me from taking this to IANA is **which** current and future CAIPs MUST or CAN a parser of these URIs know how to parse.  All CAIPs that define an identifier? Only heirarchical CAIPs? Do they have to be "final" status? 

For example, dropping CAIP-21 and -22 makes perfect sense, as they've been superseded. CAIP-20 there are many arguments for dropping (no one, AFAIK, uses CAIP-20 in prod! the MetaMask Snaps team has even expressed interest in superceding it with some other way of canonicalizing native/gas tokens).  But, devil's advocate, it's also an heirarchical identifier syntactically equivalent to CAIP-19, and the CAIP defining it is at the same status... so it gets a little squishy. 

I wouldn't say these questions are blocking this being accepted as a Draft and soliciting further feedback, but I _DO_ think looking at use-cases and how this could get adopted in prod could be informed by these options (leaving it open-ended to future CAIPs, or never finalizing this CAIP and adding new forms of identifier one by one, or writing new CAIPs over time if additional URI shapes get added).

bumblefudge (approved):

CAIPs/caip-322.md
```diff
@@ -45,7 +45,62 @@ Detailed rationales for each component of this scheme can be found following the
 
 ## Test Cases
 
-// TODO
+### Chain Identifiers (CAIP-2)
+
+```
+# Ethereum Mainnet
+caip:eip155:1
```
bumblefudge, line 52: Aren't these URNs, not URIs? URIs would be caip://eip155:1, non?

CAIPs/caip-322.md
```diff
@@ -45,7 +45,62 @@ Detailed rationales for each component of this scheme can be found following the
...
+```
+
+### Asset Identifiers (CAIP-19)
+
+```
+# Ether (native token via SLIP-44)
+caip:eip155:1/slip44:60
+
+# Bitcoin (native token via SLIP-44)
+caip:bip122:000000000019d6689c085ae165831e93/slip44:0
```
bumblefudge, line 90: this one is CAIP-20 😄

CAIPs/caip-322.md
```diff
@@ -60,8 +115,17 @@ As such, this should be the baseline assumption unless contradicted or caveated
 
 ## Backwards Compatibility
 
-It is important to note that most usage to date of [CAIP-2][], [CAIP-10][], and [CAIP-19][] identifiers has used these without a `caip:` prefix or a `caip:///` prefix, in contexts where these identifiers are unlikely to be encountered outside of their meaningful context.
-Care should be taken to add the `caip:` prefix when merging such lists or datasets into more 
+It is important to note that most usage to date of [CAIP-2][], [CAIP-10][], and [CAIP-19][] identifiers has used these without a `caip:` prefix, in contexts where these identifiers are unlikely to be encountered outside of their meaningful context.
+Care should be taken to add the `caip:` prefix when merging such lists or datasets into more general-purpose systems or URI-aware contexts.
+
+When migrating to URI-based systems:
+- Add the `caip:` prefix when exporting identifiers to external systems
```
bumblefudge, line 122: `caip://`?

PatrickAlphaC: Just `caip:`, the `://` is if we wanted to make it a URL. 

Let me give you another example, in the [ERC-721](https://eips.ethereum.org/EIPS/eip-721), they specify a `tokenURI`, which is a valid URI for the metadata of a token.

That can 100% be "normal" URL, like "https://some-domain.extension/nftmetadata".

But it can also be a base64 encoded raw data URI, like so:

```
data:image/svg+xml;base64,PHN2ZyB2aWV3Qm94PSIwIDAgMjAwIDIwMCIgd2lkdGg9IjQwMCIgIGhlaWdodD0iNDAwIiB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciPgogIDxjaXJjbGUgY3g9IjEwMCIgY3k9IjEwMCIgZmlsbD0ieWVsbG93IiByPSI3OCIgc3Ryb2tlPSJibGFjayIgc3Ryb2tlLXdpZHRoPSIzIi8+CiAgPGcgY2xhc3M9ImV5ZXMiPgogICAgPGNpcmNsZSBjeD0iNzAiIGN5PSI4MiIgcj0iMTIiLz4KICAgIDxjaXJjbGUgY3g9IjEyNyIgY3k9IjgyIiByPSIxMiIvPgogIDwvZz4KICA8cGF0aCBkPSJtMTM2LjgxIDExNi41M2MuNjkgMjYuMTctNjQuMTEgNDItODEuNTItLjczIiBzdHlsZT0iZmlsbDpub25lOyBzdHJva2U6IGJsYWNrOyBzdHJva2Utd2lkdGg6IDM7Ii8+Cjwvc3ZnPgo=
```

If you paste this into your browser, you'll just the raw data that's embedded in this URI. It's the imageURI of an NFT!

For reference, we teach this in our [Cyfrin Updraft NFT curriculum.](https://github.com/Cyfrin/foundry-nft-cu)

Does this make sense?

PatrickAlphaC:
I responded to your messages, I've let the SEAL team know to comment on this as well! Let me know what you think.

Also, you can see the `caip` as a provisional URI on the IANA site here: https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml

Woot woot!

bumblefudge:
> I responded to your messages,

@PatrickAlphaC did you maybe forget to push the new commit or review? i'm not seeing any responses or updates to the PR!

PatrickAlphaC:
@bumblefudge I responded to your comments. Let me know if you disagree with my rationale for the changes as such, or how I should update given my responses.

PR #394, opened 2026-01-03
CAIP: Generic Credential Interaction Protocol (GCIP)

anchupin:
CAIP for Generic Credential Interaction Protocol (GCIP)
Preview - https://github.com/Open-Store-Foundation/app/blob/main/lib/gcip/CAIP.md
Discussion - http://t.me/firebox_official

P.S. Will update image resources, title, commit name etc. after CAIP number will be defined.

### DIFF
--- a/CAIPs/caip-gcip.md
+++ b/CAIPs/caip-gcip.md
@@ -1,0 +1,1223 @@
+---
+caip: UNDEFINED
+title: Generic Credential Interaction Protocol (GCIP)
+discussions-to: https://github.com/ChainAgnostic/CAIPs/pull/
+author: Andrei Chupin
+status: Draft
+type: Standard
+category: Interface
+created: 2025-11-30
+---
+
+## Abstract
+
+A protocol based on FIDO2 WebAuthn principles for delegating private key storage and transaction
+signing to a separate secure application (Signer App). It allows Client Apps (Wallets, DApps) running
+on various platforms (Mobile, Desktop, Web) to request Public Keys and sign content using a
+credential ID, without direct access to the user's mnemonic phrases or private keys. The protocol
+utilizes a fully binary block structure to ensure efficiency and compatibility across various
+transports.
+
+## Motivation
+
+### 1. Problems with keys living inside “hot” wallet apps
+
+The current model of storing keys within wallet applications has several significant security and
+operational drawbacks:
+
+1. **Audit Impossibility:** Frequent updates for new features make it infeasible to conduct professional security audits for every release.
+2. **Unverifiable Integrity:** Centralized, closed-source release processes mean users cannot verify that the app code matches the published source, forcing them to trust that no backdoors were added.
+3. **Bloated Attack Surface:** Modern wallets have browsers, dapp connectors, push messaging, analytics, and many other SDKs. Each feature introduces potential vulnerabilities.
+4. **Network Exposure:** Wallets inherently require internet access, providing a direct channel for malware to access private keys.
+5. **Seed Exposure:** Users repeatedly import seeds into multiple apps/devices, multiplying the points of failure and increasing the chance of falling for phishing attacks.
+
+### 2. Benefits of a dedicated Signer
+
+This CAIP proposes a **Separation of Concerns**: offload sensitive operations (seed storage and
+signing) to an isolated Signer App, while Client Apps request public keys and signatures via a
+narrow, auditable protocol. Key benefits:
+
+- **Isolation:** Private keys are stored in a minimal, isolated application, protecting them from
+  vulnerabilities in complex, frequently updated Client Apps.
+- **Reduced Remote Surface:** The Signer App can operate without internet permissions and avoid
+  shared/public storage, effectively acting as software-based cold storage.
+- **Auditability:** The Signer's small codebase and minimal dependencies make it significantly
+  easier to audit and secure than a monolithic wallet.
+- **One Seed, Many Apps:** Users import their mnemonic phrase once into the Signer, then use it
+  across unlimited Wallets and DApps without re-exposing the seed.
+- **Consistent Trust:** Users always review and confirm transactions in the same trusted Signer
+  interface, regardless of which app initiates the request.
+- **Cross-Device Ready:** Designed to support cross-device signing and future credential
+  synchronization between devices.
+- **Binary Protocol:** Uses a compact binary structure (CBOR/Bytes), reducing memory usage and
+  processing overhead. The compressed format allows passing large datasets through constrained
+  OS channels (e.g., Android Intents, BLE, or QR codes) where standard JSON would fail.
+- **Reduced Liability:** Wallet developers can build rich interfaces without the high risk and
+  complexity of managing long-term secrets directly.
+- **Innovations Velocity:** Solves the "cold start" problem for new security technologies. New
+  key storage solutions (e.g., MPC, Web3Auth, Identity Solutions etc.) can launch as GCIP-compliant apps and
+  immediately work with all existing GCIP-enabled wallets, without needing every wallet developer to
+  integrate a custom SDK.
+
+### Signer and Wallet
+| | |
+|---|---|
+| <img src="https://github.com/Open-Store-Foundation/brandbook/blob/main/Firebox/screens/1Signer.png?raw=true" alt="1Signer" width="250"> | <img src="https://github.com/Open-Store-Foundation/brandbook/blob/main/Firebox/screens/2Wallet.png?raw=true" alt="2Wallet" width="250"> |
+
+### Connecting Flow
+| 1.                                                                                                                                | 2.                                                                                                                                        | 3.                                                                                                                                        | 4.                                                                                                                                  |
+|-----------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------|
+| <img src="https://github.com/Open-Store-Foundation/brandbook/blob/main/Firebox/screens/3Der.png?raw=true" alt="3Der" width="250"> | <img src="https://github.com/Open-Store-Foundation/brandbook/blob/main/Firebox/screens/4Wallets.png?raw=true" alt="4Wallets" width="250"> | <img src="https://github.com/Open-Store-Foundation/brandbook/blob/main/Firebox/screens/5Connect.png?raw=true" alt="5Connect" width="250"> | <img src="https://github.com/Open-Store-Foundation/brandbook/blob/main/Firebox/screens/6Pass.png?raw=true" alt="6Pass" width="250"> |
+
+### Signing Flow
+| 1.                                                                                                                                       | 2.                                                                                                                                  | 3.                                                                                                                                 | 4.                                                                                                                                       |
+|------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| <img src="https://github.com/Open-Store-Foundation/brandbook/blob/main/Firebox/screens/7Connects.png?raw=true" alt="7Connects" width="250"> | <img src="https://github.com/Open-Store-Foundation/brandbook/blob/main/Firebox/screens/8Sign.png?raw=true" alt="8Sign" width="250"> | <img src="https://github.com/Open-Store-Foundation/brandbook/blob/main/Firebox/screens/9Pass.png?raw=true" alt="9Pass" width="250"> | <img src="https://github.com/Open-Store-Foundation/brandbook/blob/main/Firebox/screens/10Send.png?raw=true" alt="10Send" width="250"> |
+
+## Specification
+
+### 1. Definitions
+
+#### 1.1. Actors and State
+
+- **GCIP:** Generic Credential Interaction Protocol.
+- **Signer App (Authenticator):** The secure application that stores private keys and performs
+  signing. Acts similarly to a FIDO2 Authenticator.
+- **Client App (Relying Party / Client):** The interface application for interacting with
+  blockchains. It requests credentials (Public Keys) and signing operations.
+- **Credential:** A unique handle representing a specific key pair (Public/Private) derived from the
+  Signer App's master seed using a specific derivation path and algorithm. It is identified by a
+  `credId`.
+- **Connection** is session concept represented by `connectionId`, created during `gcip.connect` and
+  referenced by later calls (for example, `gcip.extend` and `gcip.sign`).
+
+### 2. Core Protocol
+
+GCIP is a request/response protocol. The Client sends a binary `GcipBlock`, and the Signer returns a
+binary `GcipBlock` with a status code and an optional payload.
+
+At the wire level, a packet is:
+
+`GcipBlockHeader || EncryptionMessage(PayloadBytes)`
+
+For all operations `PayloadBytes` is the CBOR-encoded `EncryptionMessage` wrapper (see **2.2.0.
+EncryptionMessage**). The operation request/response payload is carried inside
+`EncryptionMessage.data`.
+
+#### 2.1. Block Info
+
+##### 2.1.1. `GcipBlock` header fields (big-endian)
+
+| Offset | Size | Field            | Type     | Description                                            |
+|:-------|:-----|:-----------------|:---------|:-------------------------------------------------------|
+| 0      | 1    | **Version**      | `uint8`  | Protocol version (e.g., `1`).                          |
+| 1      | 2    | **Status**       | `uint16` | Status code. See **3.9.0**.                            |
+| 3      | 2    | **Nonce**        | `int16`  | Rolling number to prevent replay / correlate requests. |
+| 5      | 1    | **Method**       | `int8`   | Operation method code. See **2.1.2**.                  |
+| 6      | 4    | **Length**       | `int32`  | Payload length in bytes.                               |
+| 10     | N    | **PayloadBytes** | `bytes`  | Payload bytes (typically CBOR `EncryptionMessage`).    |
+
+If `Status` is a common error (status values `1..63`), the block MAY be sent as a short header
+without `Method`, `Length` and `PayloadBytes`.
+
+##### 2.1.2. Method code registry
+
+Method codes are distinguishable by range: `0-127` are reserved for requests and `128-255` are
+reserved for responses. The `Method` field is `int8`, so response codes may be represented as
+negative values in some implementations.
+
+| Method                     | Value (Byte) | Description                                                |
+|:---------------------------|:-------------|:-----------------------------------------------------------|
+| `gcip.exchange.request`    | 0            | Request to establish a secure session or exchange keys.    |
+| `gcip.exchange.response`   | 128 (-128)   | Response to `gcip.exchange.request`.                       |
+| `gcip.connect.request`     | 1            | Request credentials from the Signer App.                   |
+| `gcip.connect.response`    | 129 (-127)   | Response to `gcip.connect.request`.                        |
+| `gcip.extend.request`      | 2            | Request additional credentials for an existing connection. |
+| `gcip.extend.response`     | 130 (-126)   | Response to `gcip.extend.request`.                         |
+| `gcip.sign.request`        | 3            | Request a signature for a specific payload.                |
+| `gcip.sign.response`       | 131 (-125)   | Response to `gcip.sign.request`.                           |
+| `gcip.disconnect.request`  | 4            | Request to close an existing connection.                   |
+| `gcip.disconnect.response` | 132 (-124)   | Response to `gcip.disconnect.request`.                     |
+
+#### 2.2. Common Information about Encryption & Wrapping
+
+To ensure security across various transport environments (both trusted and untrusted), GCIP supports
+an encryption wrapping mechanism for session-based communications using `EncryptionMessage` (see *
+*2.2.0**).
+
+##### 2.2.0. EncryptionMessage (Wrapper)
+
+Used as the payload of `GcipBlock` to carry either plaintext method CBOR bytes (handshake) or
+ciphertext bytes (session).
+
+| Key    | Field           | Type                 | Description                                                                               |
+|:-------|:----------------|:---------------------|:------------------------------------------------------------------------------------------|
+| `0x01` | **eid**         | ID (16 bytes) / null | Exchange/session identifier. Optional (nil) for initial handshake request.                |
+| `0x02` | **iv**          | Bytes (12) / null    | AES-GCM nonce for encrypted sessions.                                                     |
+| `0x03` | **exchangeKey** | **COSE_Key** / null  | Optional ephemeral ECDH public key (COSE Map) during handshake / exchange. See **3.9.2**. |
+| `0x04` | **data**        | Bytes                | Plaintext CBOR method payload (when `iv` absent) or ciphertext bytes (when `iv` present). |
+
+##### 2.2.1. Wrapping rules
+
+- **Handshake (Connect / Exchange)**: Requests may be sent with an unencrypted `EncryptionMessage`
+  wrapper where `iv` is omitted. The `eid` MUST be absent (null) or all-zero to indicate no active
+  session.
+- **Session (Connect / Sign / Extend / Disconnect)**: When a secure session is established, requests
+  and responses are wrapped in an `EncryptionMessage` where `iv` is present and `data` contains
+  AES-GCM ciphertext. The `eid` identifies the session used to derive the key.
+
+##### 2.2.2. Supported primitives
+
+- **Key agreement**: ECDH over P-256, ECDH over X25519
+- **Key derivation**: HKDF-SHA256
+- **Encryption**: AES-256-GCM
+
+GCIP uses two operational modes, depending on transport trust assumptions.
+
+##### 2.2.3. Same-device trusted transports (Intent, ActionExtension, etc.)
+
+In same-device and OS-mediated transports where the channel has strong integrity and identity
+guarantees, the connection SHOULD be established with one RTT by performing `Connect` with an inline
+`exchangeKey` carried by `EncryptionMessage`.
+
+##### 2.2.4. Cross environments (QR, BLE, NFC etc.)
+
+In cross-device or otherwise interceptable transports, the client SHOULD establish a session first
+using `Exchange`, and then run `Connect` inside the established encrypted session.
+
+##### 2.2.5. Session key derivation
+
+Each party generates an ephemeral ECDH key pair and exchanges public keys (during `Exchange` or
+`Connect` handshake).
+
+| Name               | Size | Calculation                                       | Description                                 |
+|:-------------------|:-----|:--------------------------------------------------|:--------------------------------------------|
+| `ephemeralPrivKey` | 32   | `random()`                                        | Sender's ephemeral ECDH private key         |
+| `peerPubKey`       | 65   | `bytes`                                           | Peer's ephemeral ECDH public key            |
+| `sharedSecret`     | 32   | `ECDH(ephemeralPrivKey, peerPubKey)`              | Raw shared secret bytes                     |
+| `eid`              | 16   | `ID`                                              | Exchange session identifier, **See 3.9.1.** |
+| `salt`             | 16   | `eid`                                             | HKDF salt                                   |
+| `info`             | 30   | `"GCIP/1/Session/AES-256-GCM"`                    | HKDF info                                   |
+| `sessionKey`       | 32   | `HKDF(secret=sharedSecret, salt=salt, info=info)` | AES-GCM session key                         |
+
+##### 2.2.6. Message formats: handshake vs session
+
+##### 2.2.6.1 Handshake (no established session yet)
+
+<p align="left">
+<img
+ src="https://github.com/anchupin/ticket-5313/blob/main/Screenshot%202025-12-29%20at%2017.26.12.png?raw=true"
+ alt="GcipBlock"
+ width="450"
+ >
+</p>
+
+Request:
+
+| Field         | Size | Value          | Description                                    |
+|:--------------|:-----|:---------------|:-----------------------------------------------|
+| `eid`         | 0    | `null`         | Absent (null)                                  |
+| `iv`          | 0    | `null`         | Absent for handshake request                   |
+| `exchangeKey` | N    | `clientPubKey` | Client's ephemeral P-256 public key (COSE_Key) |
+| `data`        | N    | `plaintext`    | CBOR-encoded method payload                    |
+
+Response:
+
+| Field         | Size | Value          | Description                                    |
+|:--------------|:-----|:---------------|:-----------------------------------------------|
+| `eid`         | 16   | `ID`           | New session ID                                 |
+| `iv`          | 12   | `random()`     | AES-GCM nonce                                  |
+| `exchangeKey` | N    | `signerPubKey` | Signer's ephemeral P-256 public key (COSE_Key) |
+| `data`        | N    | `ciphertext`   | AES-GCM ciphertext                             |
+
+##### 2.2.6.2 Session (existing session)
+
+<p align="left">
+<img
+ src="https://github.com/anchupin/ticket-5313/blob/main/Screenshot%202025-12-24%20at%2012.09.07.png?raw=true"
+ alt="GcipBlock"
+ width="450"
+ >
+</p>
+
+Request / Response:
+
+| Field         | Size | Value        | Description         |
+|:--------------|:-----|:-------------|:--------------------|
+| `eid`         | 16   | `ID`         | Existing session ID |
+| `iv`          | 12   | `random()`   | AES-GCM nonce       |
+| `exchangeKey` | 0    | `null`       | MUST be absent      |
+| `data`        | N    | `ciphertext` | AES-GCM ciphertext  |
+
+##### 2.2.7. AEAD (AES-GCM) and AAD binding
+
+AES-GCM encryption is performed as:
+
+| Name         | Size | Calculation                               | Description              |
+|:-------------|:-----|:------------------------------------------|:-------------------------|
+| `ciphertext` | N    | `AES-GCM(sessionKey, iv, aad, plaintext)` | AES-GCM ciphertext bytes |
+
+The AAD binds the encrypted payload to the `GcipBlock` header and the session context.
+
+| Name        | Size | Calculation                              | Description                        |
+|:------------|:-----|:-----------------------------------------|:-----------------------------------|
+| `headerSig` | 8    | `ver \|\| status \|\| nonce \|\| method` | Header binding (excludes `Length`) |
+| `aad`       | 24   | `headerSig \|\| eid`                     | AAD for session messages           |
+
+##### 2.2.8. `GcipBlock.Length` and payload bytes
+
+`GcipBlock.Length` MUST equal the length in bytes of `GcipBlock.PayloadBytes`.
+
+In encrypted operations, `PayloadBytes` is the CBOR-encoded `EncryptionMessage` map. This means
+`Length` is the size of the encoded wrapper, not the raw ciphertext size.
+
+
+
+#### 2.3. Failure Handling
+
+If an operation fails, the response block MUST return an appropriate non-zero status code. 
+**See 3.9.0.**
+
+#### 2.4. Operations
+
+This section describes operations as a single combined unit: overall block structure, main
+request/response structure, and flow description. Nested CBOR types are defined in **3**.
+
+##### 2.4.1. Connect
+
+**Overview**: Create a new connection session and return one or more derived public credentials (
+`credId` plus public key material).
+
+**Communication flow graph**: placeholder (image will be attached)
+
+**Main request structure** (`gcip.connect.request`):
+
+```
+{
+  clientData(0x01): ClientData,
+  credentials(0x02): [ CredentialRequest ],
+  transport(0x03): Transport,
+  meta(0x20): b'data'
+}
+```
+
+**Main response structure** (`gcip.connect.response`):
+
+```
+{
+  connectionId(0x01): ID,
+  connectionType(0x02): 0,
+  signerData(0x03): SignerData,
+  credentials(0x04): [ CredentialResponse ],
+  meta(0x20): b'data'
+}
+```
+
+**Flow description**:
+
+1. Client sends `gcip.connect.request` wrapped in `EncryptionMessage`.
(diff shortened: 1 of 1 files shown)


bumblefudge:
Wow this is a really promising CAIP, although I have to recuse myself from deep review because I have dangerously little grasp on OS-level security and even the basics of WebAuthN qua protocol. Maybe @kdenhartog could review? @arein whom do we know that's implemented WebAuthN in a wallet already?

anchupin:
@bumblefudge 
Just in case, I'm actually one of the architects of Swift Wallet(based on Passkey) at Trust Wallet .

anchupin:
@bumblefudge do you have any updates?

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+- **Key agreement**: ECDH over P-256, ECDH over X25519
+- **Key derivation**: HKDF-SHA256
+- **Encryption**: AES-256-GCM
+
+GCIP uses two operational modes, depending on transport trust assumptions.
+
+##### 2.2.3. Same-device trusted transports (Intent, ActionExtension, etc.)
+
+In same-device and OS-mediated transports where the channel has strong integrity and identity
+guarantees, the connection SHOULD be established with one RTT by performing `Connect` with an inline
```
kdenhartog: Why SHOULD instead of MUST here? It seems like under a formal analysis this would make it harder to reason about the TOFU problem to assess security guarantees.

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+`exchangeKey` carried by `EncryptionMessage`.
+
+##### 2.2.4. Cross environments (QR, BLE, NFC etc.)
+
+In cross-device or otherwise interceptable transports, the client SHOULD establish a session first
```
kdenhartog: Why SHOULD instead of MUST here? It seems MUST would provide better security guarantees about the protocol security under a formal verification analysis

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+  and responses are wrapped in an `EncryptionMessage` where `iv` is present and `data` contains
+  AES-GCM ciphertext. The `eid` identifies the session used to derive the key.
+
+##### 2.2.2. Supported primitives
+
+- **Key agreement**: ECDH over P-256, ECDH over X25519
+- **Key derivation**: HKDF-SHA256
+- **Encryption**: AES-256-GCM
+
+GCIP uses two operational modes, depending on transport trust assumptions.
```
kdenhartog, line 177: What's the reason for these cryptographic parameters? Typically on non-CPU accelerated instruction sets chacha20poly1305 (orxchacha20poly1305 may be used for better guarantees, but less widely used) operates at a faster rate than AES-256-GCM.

Do you plan to have any form of cryptographic agility via versioning for this?

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+`exchangeKey` carried by `EncryptionMessage`.
+
+##### 2.2.4. Cross environments (QR, BLE, NFC etc.)
+
+In cross-device or otherwise interceptable transports, the client SHOULD establish a session first
+using `Exchange`, and then run `Connect` inside the established encrypted session.
+
+##### 2.2.5. Session key derivation
+
+Each party generates an ephemeral ECDH key pair and exchanges public keys (during `Exchange` or
```
kdenhartog, line 192: If both parties operate under an ephemeral ECDH generation you'll introduce a  [TOFU problem](https://en.wikipedia.org/wiki/Trust_on_first_use) for the session. Is that an acceptable tradeoff that a person could MITM the session and tamper with the transaction requests made to the authenticator?

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+</p>
+
+Request:
+
+| Field         | Size | Value          | Description                                    |
+|:--------------|:-----|:---------------|:-----------------------------------------------|
+| `eid`         | 0    | `null`         | Absent (null)                                  |
+| `iv`          | 0    | `null`         | Absent for handshake request                   |
+| `exchangeKey` | N    | `clientPubKey` | Client's ephemeral P-256 public key (COSE_Key) |
+| `data`        | N    | `plaintext`    | CBOR-encoded method payload                    |
```
kdenhartog, line 250: What's the reason for sending plaintext directly in the initial handshake request? That seems like it leaves it susceptible to tampering/interception issues

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+<img
+ src="https://github.com/anchupin/ticket-5313/blob/main/Screenshot%202025-12-29%20at%2017.26.12.png?raw=true"
+ alt="GcipBlock"
+ width="450"
+ >
+</p>
+
+Request:
+
+| Field         | Size | Value          | Description                                    |
```
kdenhartog, line 245: You'll probably want to negotiate a cryptographic suite version during this request/response pair too

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+|:--------------|:-----|:---------------|:-----------------------------------------------|
+| `eid`         | 0    | `null`         | Absent (null)                                  |
+| `iv`          | 0    | `null`         | Absent for handshake request                   |
+| `exchangeKey` | N    | `clientPubKey` | Client's ephemeral P-256 public key (COSE_Key) |
+| `data`        | N    | `plaintext`    | CBOR-encoded method payload                    |
+
+Response:
```
kdenhartog, line 252: Is there an expected error response flow here too?

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+
+| Field         | Size | Value          | Description                                    |
+|:--------------|:-----|:---------------|:-----------------------------------------------|
+| `eid`         | 16   | `ID`           | New session ID                                 |
+| `iv`          | 12   | `random()`     | AES-GCM nonce                                  |
+| `exchangeKey` | N    | `signerPubKey` | Signer's ephemeral P-256 public key (COSE_Key) |
+| `data`        | N    | `ciphertext`   | AES-GCM ciphertext                             |
+
+##### 2.2.6.2 Session (existing session)
```
kdenhartog, line 261: How long is a session persisted for and how is it agreed upon by the parties? An attacker could utilize implementation details here to spam requests for new sessions to force the responder to overload their storage state trying to keep sessions alive as a DoS attack. A recommendation should be included about how to handle session management or an expiry parameter should be added during session setup.

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+ alt="GcipBlock"
+ width="450"
+ >
+</p>
+
+Request / Response:
+
+| Field         | Size | Value        | Description         |
+|:--------------|:-----|:-------------|:--------------------|
+| `eid`         | 16   | `ID`         | Existing session ID |
```
kdenhartog, line 299: Is the `eid` cryptographically tied in any way to the session or is it meant to act like an object capability identifier which needs to be cryptographically randomly generated?

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+
+| Name         | Size | Calculation                               | Description              |
+|:-------------|:-----|:------------------------------------------|:-------------------------|
+| `ciphertext` | N    | `AES-GCM(sessionKey, iv, aad, plaintext)` | AES-GCM ciphertext bytes |
+
+The AAD binds the encrypted payload to the `GcipBlock` header and the session context.
+
+| Name        | Size | Calculation                              | Description                        |
+|:------------|:-----|:-----------------------------------------|:-----------------------------------|
+| `headerSig` | 8    | `ver \|\| status \|\| nonce \|\| method` | Header binding (excludes `Length`) |
```
kdenhartog, line 383: why is length binding excluded here? It should be deterministically calculat-able such that it can be integrity protected too. Otherwise length value tampering could lead to weird handling and potential DoS by a MITM.

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+    - To establish a new secure session in one RTT, the client includes `exchangeKey` in
+      `EncryptionMessage`, with `eid`, `iv` omitted. In this mode, the `gcip.connect.request` CBOR
+      payload is sent in plaintext inside `EncryptionMessage.data`.
+2. Signer validates request format, supported algorithms, and origin (when possible).
+3. Signer prompts the user to approve the connection and the disclosure of derived public keys.
+4. Signer derives requested credentials, assigns a `credId` to each derivation.
+5. Signer generates its own ephemeral key pair, computes the shared secret (`sessionKey`), and
+   includes its `exchangeKey` in the response `EncryptionMessage` alongside a newly generated `eid`
+   and ecrypted `payloadBody` with `sessionKey`.
+6. Signer returns `gcip.connect.response`.
```
kdenhartog: It seems like this should be using more rigorous normative language here with explanations for why they're done. Same for all the flow descriptions below

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+
+**Flow description**:
+
+1. Client sends `gcip.exchange.request` wrapped in `EncryptionMessage` with `exchangeKey` present,
+   `eid` and `iv` omitted.
+2. Signer generates a response key pair and a new `eid`.
+3. Signer returns `gcip.exchange.response` wrapped in `EncryptionMessage` with `eid` set and
+   `exchangeKey` present.
+4. Both parties compute the shared secret (`sessionKey`) using ECDH. Future messages using this`eid`
+   are encrypted with this key.
```
kdenhartog: Same here about normative language

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+and policy input, and verify it only when the transport can attest it.
+
+If a client cares about **traffic sniffing** or wants to keep the full `connect` request body
+private on interceptable channels:
+
+- Preferred: run `gcip.exchange` first, then send `gcip.connect` **inside the encrypted session** (2
+  RTT). This ensures `clientData`, credential requests, and any metadata are not observable on the
+  wire.
+- Optional (performance): send `gcip.connect` directly in handshake mode (1 RTT). This is acceptable
+  for trusted same-device transports or when plaintext request body exposure is not a concern.
```
kdenhartog, line 501: What's the point of doing both 2-RTT and 1-RTT? Seems like it would be better to just do 1-RTT for session setup then do message encryption instead after the session is setup. This is what TLS 1.3 moved to is to only support 1-RTT and 0-RTT optionally

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+        }
+    }
+    
+    func handleGcipRequest(data: Data) {
+        // Parse GCIP Block, verify, and present Signer UI...
+    }
+}
+```
+
+##### 4.2.2. Wallet iOS Implementation (Sender)
```
kdenhartog: A standard shouldn't be defining implementation details like this. It's purpose is to define the rules of the protocol and things like this should be moved into a separate non-normative note. Otherwise, people implicitly build assumptions in that are not normatively defined and it produces gaps in the implementation.

This is also why I was asking if multiple implementers are working on this is because there's not much value in standardizing something if we don't need to agree on behavior. It would be a strong indicator of pre-mature standardization.

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+|------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| <img src="https://github.com/Open-Store-Foundation/brandbook/blob/main/Firebox/screens/7Connects.png?raw=true" alt="7Connects" width="250"> | <img src="https://github.com/Open-Store-Foundation/brandbook/blob/main/Firebox/screens/8Sign.png?raw=true" alt="8Sign" width="250"> | <img src="https://github.com/Open-Store-Foundation/brandbook/blob/main/Firebox/screens/9Pass.png?raw=true" alt="9Pass" width="250"> | <img src="https://github.com/Open-Store-Foundation/brandbook/blob/main/Firebox/screens/10Send.png?raw=true" alt="10Send" width="250"> |
+
+## Specification
```
kdenhartog, line 77: It would be useful to include a protocol diagram in here to quickly understand how it works with the data being passed around.

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+
+**Overview**: Request additional credentials under an existing connection, returning new or
+already-known derivations.
+
+**Main request structure** (`gcip.extend.request`):
+
+```
+{
+  connectionId(0x01): ID,
+  credentials(0x02): [ CredentialRequest ],
```
kdenhartog, line 513: Are credential request structures being defined anywhere? It's not really clear how credentials are meant to be sanitized, parsed, validated, and processed based on the type of data passed to them.

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+This section defines the nested CBOR structures referenced by the operation payloads in **2.4**.
+
+#### 3.2. Party (ClientData / SignerData)
+
+**ClientData**
+
+| Key    | Field      | Type  | Req/Opt/Cond | Description                                                                                 |
+|:-------|:-----------|:------|:-------------|:--------------------------------------------------------------------------------------------|
+| `0x01` | **name**   | UTF-8 | required     | Display name shown to the user. Max 50 chars.                                               |
+| `0x02` | **origin** | UTF-8 | required     | Web Origin (RFC 6454), e.g. `https://app.uniswap.org`. Required for web-based interactions. |
```
kdenhartog: The defacto standard for browsers is WHATWG URL spec not RFC 6454. While the two are highly related, I'm not certain if there's been an deviations between the two

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+
+**SignerData**
+
+| Key    | Field      | Type  | Req/Opt/Cond | Description                                         |
+|:-------|:-----------|:------|:-------------|:----------------------------------------------------|
+| `0x01` | **name**   | UTF-8 | required     | Display name shown to the user. Max 50 chars.       |
+| `0x02` | **scheme** | UTF-8 | required     | Platform Scheme (e.g., `android`, `ios`, `chrome`). |
```
kdenhartog: scheme here seems like an odd semantic naming. It makes me think of scheme in a URI not what platform is used. Also, is `chrome` being listed as a platform here because of the browser or chromeOS?

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+| `0x03` | **id**     | UTF-8 | required     | **Signer ID**. See **3.9.12**.                      |
+
+#### 3.3. Challenge
+
+The content to be signed.
+
+| Key    | Field         | Type  | Req/Opt/Cond | Description                                                                        |
+|:-------|:--------------|:------|:-------------|:-----------------------------------------------------------------------------------|
+| `0x01` | **payload**   | Bytes | required     | Raw bytes to be signed or challenged.                                              |
```
kdenhartog, line 668: How is this meant to parsed and displayed to the user on the device to prevent blind signing?

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+
+| Type | Size | Standard                                          | Description                    |
+|:-----|:-----|:--------------------------------------------------|:-------------------------------|
+| `ID` | 16   | UUID IETF variant (variant 2) V4 IETF, BLOB(CBOR) | Universally Unique Identifier. |
+
+##### 3.9.2. COSE Definitions
+
+##### 3.9.2.1. COSE Key Structure
+
+A CBOR map conforming to COSE (RFC 8152) standards. Used for `exchangeKey` and credential `payload`.
```
kdenhartog: So is this protocol defining everything within the CWE data model or is the goal here to define a new separate data model on CBOR, but occasionally use CWEs? This wasn't very clear to me.

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+| 0x02 (`2`) | base64 | The payload is a Base64 encoded string.        |
+
+##### 3.9.4. Supported Connection Types
+
+| ID         | Type             | Description                                                                                       |
+|:-----------|:-----------------|:--------------------------------------------------------------------------------------------------|
+| 0x00 (`0`) | `platform`       | Connection established within the same device (e.g., via Android Intent or iOS Action Extension). |
+| 0x01 (`1`) | `cross-platform` | Connection established between two different devices (e.g., via QR code bridge).                  |
+
+##### 3.9.5. Supported Transport Types
```
kdenhartog, line 825: No support for Wifi aware or airdrop now that it's been [reverse engineered](https://blog.google/products-and-platforms/platforms/android/quick-share-airdrop/) by Google?

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+| 0x00 (`0`) | `internal` | Internal or OS-based IPC (Intent, XPC). |
+| 0x01 (`1`) | `usb`      | USB connection.                         |
+| 0x02 (`2`) | `nfc`      | NFC connection.                         |
+| 0x03 (`3`) | `ble`      | Bluetooth Low Energy connection.        |
+
+##### 3.9.6. Supported Key Exchange Algorithms
+
+| ID    | Algorithm        | Description                                                |
+|:------|:-----------------|:-----------------------------------------------------------|
+| `-25` | ECDH-ES-HKDF-256 | Elliptic Curve Diffie-Hellman Ephemeral Static (COSE -25). |
```
kdenhartog, line 838: Don't think ES is the correct key exchange algorithm here if both parties are generating ephemeral key pairs for the session setup as mentioned on line 190

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+| `-104`  | BLAKE2b-224             |
+| `-105`  | BLAKE2b-256             |
+
+##### 3.9.10. Supported Origin Schemes
+
+For `origin` field in Party data (Web contexts).
+
+| Scheme  | Description     |
+|:--------|:----------------|
+| `https` | Web Application |
```
kdenhartog, line 876: no support of wss://? Also, this did make me wonder why define all of this when you could probably build a much lighter protocol in the application layer using WebRTC instead?

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+For `id` field in SignerData. Simple string identifier that depends on the scheme.
+
+- **Android**: `com.example.app`
+- **iOS**: `teamID.com.example.app` or `bundleId`
+- **Web**: `https://app.uniswap.org`
+- **Chrome Extension**: `ExtensionID`
+
+### 4. Transport Layer
+
+The protocol is transport-agnostic, but specific implementations are defined for supported
```
kdenhartog: "transport-agnostic" is a can of worms. I wouldn't suggest aspiring to that goal since you always just end up having to make various profiles for each transport because of the guarantees of the transport layer. I'd suggest being opinionated here and then if you need an additional transport layer people have to show why it's needed and it can be extended then.

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+platforms:
+
+#### 4.1. Android Implementation Guidelines
```
kdenhartog: implementation details aren't needed here either. See the comment below about iOS implementation details.

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+within CTAP, which is outside of the scope of this proposal.
+
+### 5. Rationale
+
+#### 5.1. Why not WebAuthn / CTAP?
+
+While Passkeys are a well-designed standard for authentication, they have several limitations for
+cryptocurrency wallets:
+
+* 5.1.1. **Incompatibility with BIP-84 and Derivation Paths:** Passkey = 1 pubKey, for more you
```
kdenhartog: Can't an authenticator choose to support this if it wants too? I don't see anything within how authenticators are defined that preclude this from happening as an implementation detail of that authenticator.

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+  should create another one, so to cover several chains and accounts you should create hundreds of
+  passkeys. They typically do not support Hierarchical Deterministic (HD) Wallets.
+* 5.1.2. **Public Key Retrieval:** It is impossible to retrieve the public key after creation. You must store it somewhere to be able to restore the wallet address
+  later.
+* 5.1.3. **No Private Key Visibility:** Users cannot view their Private Key (PK) to create a paper
```
kdenhartog: This isn't necessarily true, also this was a beneficial design choice. It's highly specific to the authenticator that's used, but if the user really wanted to extract it they'll likely be able to via: https://fidoalliance.org/specs/cx/cxf-v1.0-ps-20250814.html

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+  this client JSON data on-chain, which is sub-optimal (gas expensive).
+* 5.1.6. **Reliability:** Passkeys are often an additional auth level. Providers do not guarantee
+  credential safety, it can be erased any time.
+* 5.1.7. **Limited Signing Transparency:** Users often cannot see the actual content they are
+  signing, as passkeys typically sign the data without displaying it.
+* 5.1.8. **Service-Specific Binding:** Passkeys bind specific keys to specific services (
+  `service1-cred1`, `service2-cred2`). GCIP employs a different model where a single credential
+  source is reused across multiple services (`service1-cred1`, `service2-cred1`), which is more
+  suitable for wallet interoperability.
+* 5.1.9. **Protocol Rigidity:** The protocol is difficult to extend or modify, limiting the ability
```
kdenhartog: rapid extensibility creates tradeoffs of webcompat fragmentation that will harm adoption. This is the reason PRF extension is so un-reliable is that it's highly specific to what's supported by the authenticator and many don't support it. You'll likely encounter the same tradeoffs here with just simple things like cryptographic algorithm support

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+GCIP is not a replacement for WalletConnect; rather, they serve complementary roles in the
+transaction flow.
+
+* **Complementary Flow:** While DApps can connect directly to a Signer via GCIP (e.g.,
+  over Deep Links or BLE), a more robust and secure architecture should involve a Wallet
+  application as an intermediary layer.
+* **Recommended Architecture:**
+  `DApp <-> Wallet (via WalletConnect / Deep Link / JS Interface) <-> Signer (via GCIP)`.
+* **Safety Layer:** Wallets play a crucial role in interpreting raw transactions, simulating
+  outcomes, and performing initial safety checks.
```
kdenhartog, line 1003: This is the same architecture that's used by many extension wallets that support hardware wallets. However, there's been tradeoffs with blind signing that's occurred due to this because the signer needs to be able to display what's being signed because MITM attacks can occur from the initial setup via a TOFU attack.

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+          signature request without the initial private exchange material.
+    * **3. Untrusted / Interceptable Environments**: For physical or wireless transports like BLE
+      and NFC, data can be sniffed or intercepted. The `exchange` mechanism allows establishing a
+      shared secret (ECDH) so that subsequent communication can be encrypted. If the client wants
+      `clientData` and other `connect` request fields to be encrypted as well, it should run
+      `exchange` first and then run `connect` inside the encrypted session.
+
+* **5.4.8. Request Correlation Hash**
+    * For correlation and deduplication, a signer should compute `blockId` for each `gcip.request.sign` and check uniqueness across all requests:
+    * `blockId` (16 bytes) = `SHA-256(GcipBlock)` (SHA-256 digest of the block).
```
kdenhartog, line 1076: This feels very randomly placed here. It probably makes sense to just make this a normative statement and include it in the definition for the sign flow description.

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+- RFC 8949 (CBOR): https://www.rfc-editor.org/rfc/rfc8949.html
+- RFC 8152 (COSE): https://www.rfc-editor.org/rfc/rfc8152.html
+- RFC 6454 (Web Origin): https://www.rfc-editor.org/rfc/rfc6454.html
+- RFC 5869 (HKDF): https://www.rfc-editor.org/rfc/rfc5869.html
+- BIP-32 (HD Wallets): https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki
+- BIP-84 (Derivation Scheme): https://github.com/bitcoin/bips/blob/master/bip-0084.mediawiki
+- RFC 8032 (EdDSA): https://www.rfc-editor.org/rfc/rfc8032.html
+- RFC 7748 (Curve25519): https://www.rfc-editor.org/rfc/rfc7748.html
+- NIST SP 800-38D (
+  AES-GCM): https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38d.pdf
```
kdenhartog: Nit: in Markdown you can reuse this section if you format it properly so that when you use the key here you'll be able to reference them later.

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+**Proposal:** Use the `gcip.exchange` method to securely rotate the `sessionKey` of an active connection.
+
+* **Context**: Long-lived connections may want to rotate encryption keys forward secrecy or valid session maintenance without forcing the user to re-approve the connection in the UI.
+* **Mechanism**: The Client sends a `gcip.exchange.request` inside the existing encrypted session. A fresh ECDH handshake occurs, resulting in a new `eid` and `sessionKey`. The logical `connectionId` remains valid, but subsequent transport messages use the new encryption context.
+
+
+## Known implementation
+- [GCIP KMP](https://github.com/Open-Store-Foundation/app/tree/main/lib/gcip)
+- [Signer](https://github.com/Open-Store-Foundation/app/tree/main/apps/signer)
+- [Wallet](https://github.com/Open-Store-Foundation/app/tree/main/apps/sample/wallet)
```
kdenhartog, line 1238: hmm it seems all of these are paid for services to charge for storage of the DApp so it can be integrity validated by the wallet. Am I interpreting the docs of this right and how is this planning to get DApps to do that so that the wallet can validate it? This seems like a metcalfe's law bootstrapping problem.

kdenhartog (commented):
Overall this looks like a pretty cool design, but I'm a bit skeptical now is the right time to standardize this. Do we have a commitment from multiple parties that they're interested in working on this? If not, defining this as a CAIP is useful in so far as it becomes a public docs repo like what EIPs are which isn't very useful here.

I think we should wait to see if multiple people are committed to implementing this so that we can work through interoperability/webcompat/security tradeoffs together rather than move this thing along further and then miss something because only one or two people were looking at it as it was being defined.

anchupin:
@kdenhartog I'm solo researcher right now I see our flow like:
1. Collect feedback from different parties
2. Apply feedback to protocol implementation 
3. When I sure that first version of protocol are pretty robust ->
3.1. Working on marketing / standartization

CAIPs/caip-gcip.md
anchupin: I've been thinking about it. 
We can change to MUST to reduce complexity, the thing is Signer actually are able to work in 2 flows, so I decided to keep it in case someone want extra security for `clientData`.

CAIPs/caip-gcip.md
anchupin: Yep, here we should change to MUST

CAIPs/caip-gcip.md
anchupin: GCIP has version in block, all supported cryptography algos are connected to it.
I thought about chacha20poly1305, I have an approach to stick to WebAuthn algorithms where it's possible. 
I'm still investigating if there is a convenient way to keep session-key in Android/iOS Keystores, which doesn't support chacha20 for now.

But in overall for now we don't have any blockers to support chacha20poly1305.

CAIPs/caip-gcip.md
anchupin: For BLE, USB it's pretty hard to establish this attack.
For Android/iOS you can read - 5.5.4.
Also I have plan to introduce KeyRotation, but it won't help with 5.5.4. case.

CAIPs/caip-gcip.md
anchupin: First of all it's plaintext only for trusted env(Intent, ActionExtension) request. 

1. It's almost impossible(only by using complicated UI/UX) to make safe encryption for offline only applications for initial request. If I intercepted request by user's mistake and it can read GCIP blocks it highly likely can decrypt the request.
2. Data in `connect.request` can be considered like open by default:
2.1. Anyone can decompile app and see it there
2.2. For Android any app can see all installed applications with their domains, names, logos etc.
3. Also which is why I allow to start session in trusted devices with `exchange.request`, so the `connect.request` will be encrypted.

CAIPs/caip-gcip.md
anchupin: Sorry didn't get your point here.

CAIPs/caip-gcip.md
anchupin: Actually I'm thinking to get rid of P-256, after I'll implement EdDH and use it exclusively(the only doubt is that so many leak devices support P-256 only, will investigate if we need to support it). 

In other case in that protocol version all DH algos considered are equally safe and client request any(we can't afford 2-RTT).

CAIPs/caip-gcip.md
anchupin: All errors in GCIP block status

CAIPs/caip-gcip.md
anchupin: Session live the same like in WebAuthn all the time, unless app deleted or user will delete session manually.
It's pretty hard to conduct DoS attack, because user should manually approve each request in UI.

CAIPs/caip-gcip.md
anchupin: Could you give me any example how would you like it to be?

CAIPs/caip-gcip.md
anchupin: 3. https://github.com/ChainAgnostic/CAIPs/pull/394#discussion_r2681112337

CAIPs/caip-gcip.md
anchupin: For now it's here to easier understand protocol and implementation as well as screenshots at the beginning, I think I'll move it away later.

CAIPs/caip-gcip.md
anchupin: Yep, nice idea

CAIPs/caip-gcip.md
anchupin: You can check 3.9.11., will update e.g.

CAIPs/caip-gcip.md
anchupin: By default it's better than in Passkey, where you don't see what you signing.
Where are two optional(for now) paramtetser - fmt and transform.
fmt - in witch format we shoudl show it to the user
transform - if payload is a raw transaction, how should be transform before signing

It's a big topic, will leave several comments later, I've been thinking on it a lot.

Some points:
If payload is a raw transaction - user can copy content and check with third party service.
If payload is a hash user can compare it with what shown in wallet.(trust problem)

CAIPs/caip-gcip.md
anchupin: Other communication channels are possible to support, didn't investigated it yet in terms of UI/UX implementation, right now I use minimal implementation(what we actually have in Passkey).

CAIPs/caip-gcip.md
anchupin: .well-known/assetlinks.json
.well-known/apple-app-site-association

It's up to OS system, right now only https. Because I can check it on OS level.

CAIPs/caip-gcip.md
anchupin: The point of this phrase that I don't need to create several protocols like CTAP/WebAuthn. Sure I need some utility to establish connection for different transport type, but data transfer will use only that protocol.

CAIPs/caip-gcip.md
anchupin: Same like iOS

CAIPs/caip-gcip.md
anchupin: You can't. You CAN create your own authenticator protocol, but as I understand you can't use it for example within OS(Android, iOS etc) devices, because WebAuthn scheme if defined on protocol(even for specific OS) and OS level(e.g. you 1password, but all request/response going through Apple/Google).
You also can use webauthn.extesnsion(which is ugly), but the main consern in 5.1.9. and it will complicate UX too much.

CAIPs/caip-gcip.md
anchupin: It;s partly true, but most of Passkey providers doesn't support PK revealing and it will be inconvenient for users do distinguish which support which doesn't

CAIPs/caip-gcip.md
anchupin: Yeah, the point is WebAuthn is about different use case and it hard to adopt it to crypto/web3(companies who designed it doesn't have any intention to do that), actually I had a talk with some engs from Microsoft, who was responsible for Passkey there, they told me, it's insane to use it for crypto wallets.

CAIPs/caip-gcip.md
anchupin: There isn't blind signing here.

CAIPs/caip-gcip.md
anchupin: Yep, true

anchupin:
@kdenhartog If you want we can have a call, do discuss it faster and I'll add summary after that.

CAIPs/caip-gcip.md
anchupin: here is the length of encrypted data, we don't know it until encryption(I can calculate it actually), if MitM somehow will change data it won't be decrypted successfully, right?

CAIPs/caip-gcip.md
anchupin: eid randomly generated by Signer.

anchupin: All ID types is randomly generated with UUIDv4

CAIPs/caip-gcip.md
anchupin: Didn't get it? Do you mean COSE?
Don't see a problem, getting that we use COSE all across the protocol, what is the problem?

CAIPs/caip-gcip.md
anchupin: 3.5. Credential Request
The main question in response, it there were optional challenge in connect request(it's was more for wallet like a proof that I have PK for provided PubKey), I deleted it for a while, because it increasing connect.response body much, but it's still discussible.

CAIPs/caip-gcip.md
anchupin: Yep, we should change it to WHATWG

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+**See 3.9.0.**
+
+#### 2.4. Operations
+
+This section describes operations as a single combined unit: overall block structure, main
+request/response structure, and flow description. Nested CBOR types are defined in **3**.
+
+##### 2.4.1. Connect
+
+**Overview**: Create a new connection session and return one or more derived public credentials (
```
vcoolish-tw, line 407: Can this cred ID be revoked?

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1223 @@
...
+| `0x03` | **id**     | UTF-8 | required     | **Signer ID**. See **3.9.12**.                      |
+
+#### 3.3. Challenge
+
+The content to be signed.
+
+| Key    | Field         | Type  | Req/Opt/Cond | Description                                                                        |
+|:-------|:--------------|:------|:-------------|:-----------------------------------------------------------------------------------|
+| `0x01` | **payload**   | Bytes | required     | Raw bytes to be signed or challenged.                                              |
+| `0x02` | **fmt**       | Int   | optional     | Representation hint for `payload`. See **3.9.3**.                                  |
```
vcoolish-tw, line 669: Is there any max size of the payload?

CAIPs/caip-gcip.md
anchupin: Connection can be revoked by 2 entries, same like in Passkey.
1. Be deleted by Wallet App with 'disconnect.request'
2. Inside Signer app with 'disconnect' button

CAIPs/caip-gcip.md
anchupin: I thought about it, Signer can decide how many credential per request can be created and return dedicated error.
By default max size is only restricted by transport. itself(e.g. Android IPC - 1-3 MB)

CAIPs/caip-gcip.md
anchupin: For WebRTC I need INTERNET permission, but we Signer shouldn't have it

anchupin:
- Added BLE support
- Added request/response to sync PK/Mnemonic between signers(uni/cross device)
- Added session key rotation support

Fill changelog - https://github.com/Open-Store-Foundation/app/blob/main/lib/gcip/changelogs/30_01_2026.md

CAIPs/caip-gcip.md
```diff
@@ -0,0 +1,1277 @@
+---
+caip: UNDEFINED
```
bumblefudge:
```suggestion
caip: 394
```

obstropolos:
> ooks like a pretty cool design, but I'm a bit skeptical now is the right time to standardize this. Do we have a commitment from multiple parties that they're interested in working on this? If not, defining this as a CAIP is useful in so far as it becomes a public docs repo like what EIPs are which isn't

Hey @anchupin - definitely agreed here that this is very interesting! Wanted to follow up here as I would be curious to know if there are other folks interested in standardizing this more broadly as a CAIP to have a neutral reference somewhere.

anchupin:
@obstropolos 
I'll continue working on the protocol in couple weeks, I'm search for wallets who can support this.
But for now it's pretty hard, because it's not gonna make revenue to wallets, but takes couple weeks to review+integrate.

bumblefudge:
OK, no stress, there's no time limit on how long it stays open until there's a second implementer-champion!  In the meantime, I'll keep sending the URL of this PR to people who ask about WebAuthN profiles they can use for these kinds of use-cases.

achupin-top:
Probably we'll implement it in several months within https://github.com/tonkeeper/signer-ios

bumblefudge:
Any update here?

PR #395, opened 2026-01-13
Update docs CAIP-350

0xMonoAx:
This update includes two main changes:
- Removed Interoperable Name references: The "Interoperable Name" definition has been replaced with a more general "text representation" terminology. This change aligns with updates to ERC-7930 and has the added benefit of reducing CAIP-350's coupling to ERC-7930.
- Added Chain Identifier Text Representation section: A new section was added to explicitly define the standard format `<namespace>:<chainReference>`. This improves clarity and ensures consistency across all namespace profiles, where each profile now defines only the `<chainReference>` portion.

### DIFF
--- a/CAIPs/caip-350.md
+++ b/CAIPs/caip-350.md
@@ -32,15 +32,30 @@ Standards like CAIP-2 and CAIP-10 are simple text representations of addresses a
 
 The purpose of each namespace's profile is to specify deterministic and unambiguous conversions between the following format-pairs:
 
-- Customary address text formats of the ecosystem (which may also be described in its [CAIP-10] profile) and the Interoperable Names text representation defined in [ERC-7930]
-- Customary chain reference text formats of the ecosystem (which may also be described in [CAIP-2]) and the text representation defined in [ERC-7930]
-- [ERC-7930]'s Interoperable Address and [ERC-7930]'s Interoperable Name of the ecosystem's chain references
-- [ERC-7930]'s Interoperable Address and [ERC-7930]'s Interoperable Name of the ecosystem's addresses *for all address formats of said ecosystem*
+- **Address formats**: Between customary address text formats of the ecosystem (which may also be described in its [CAIP-10] profile) and [ERC-7930]'s Interoperable Address binary format
+- **Chain reference formats**: Between customary chain reference text formats of the ecosystem (which may also be described in [CAIP-2]) and [ERC-7930]'s Interoperable Address binary format
+- **Address text ↔ binary**: Between the text representation of addresses and [ERC-7930]'s Interoperable Address, for all address formats of said ecosystem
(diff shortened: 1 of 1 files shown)


CAIPs/caip-350.md
```diff
@@ -30,18 +30,38 @@ Standards like CAIP-2 and CAIP-10 are simple text representations of addresses a
...
 
-The purpose of each namespace's profile is to specify deterministic and unambiguous conversions between the following format-pairs:
+The purpose of each namespace's profile is to specify deterministic and unambiguous conversions between format pairs for the Addresses and Chain References of the ecosystem.
 
-- Customary address text formats of the ecosystem (which may also be described in its [CAIP-10] profile) and the Interoperable Names text representation defined in [ERC-7930]
-- Customary chain reference text formats of the ecosystem (which may also be described in [CAIP-2]) and the text representation defined in [ERC-7930]
-- [ERC-7930]'s Interoperable Address and [ERC-7930]'s Interoperable Name of the ecosystem's chain references
-- [ERC-7930]'s Interoperable Address and [ERC-7930]'s Interoperable Name of the ecosystem's addresses *for all address formats of said ecosystem*
+Chain References:
+- **customary text ↔ standard text**: Between customary chain reference text formats of the ecosystem (which may also be described in [CAIP-2]) and a standard text representation
```
bumblefudge, line 41: This is kind of confusing because [CAIP-2] profiles per namespace already unambiguously convert between customary text and [a slightly different, in some cases] standard text... I think it would help to better explain that the "standard text" here is a slight variant of CAIP-2 that is identical for many namespaces but for others less lossy.

bumblefudge (changes requested):
Other than making more explicit the relationship between CAIP-2 and CAIP-350 in the ## Specification section, I think the changes are an improvement on the existing text and seem R2G

CAIPs/caip-350.md
0xMonoAx: I added a new section titled “CAIP-2/CAIP-10 Compatibility.”

CAIPs/caip-350.md
```diff
@@ -32,11 +32,16 @@ Standards like CAIP-2 and CAIP-10 are simple text representations of addresses a
 
 The purpose of each namespace's profile is to specify deterministic and unambiguous conversions between format pairs for the Addresses and Chain References of the ecosystem.
 
-Chain References:
+#### CAIP-2/CAIP-10 Compatibility
+The "standard text" in CAIP-350 profiles is a variant of [CAIP-2]/[CAIP-10] that is identical for most namespaces, but less lossy for those where length restrictions cause truncation (e.g., Solana's genesis blockhash).
+
+Each CAIP-350 profile MUST state whether its standard text differs from [CAIP-2]/[CAIP-10], and how.
```
bumblefudge, line 38: nice good thinking

bumblefudge (approved):
💪

obstropolos (approved):

Issue #396, opened 2026-02-23
CAIP Discussion: CLI Wallet Protocol (CWP)


arein:
## CLI Wallet Protocol (CWP)

A standard protocol for CLI applications to discover and interact with wallet providers through executable plugins on PATH.

### Problem

CLI-based blockchain tools hardcode support for specific wallet providers, creating tight coupling. There's no CLI equivalent of EIP-6963 (browser wallet discovery).

### Proposed Solution

- **Binary naming convention**: `wallet-<name>` executables on PATH
- **JSON stdin/stdout communication**: Avoids shell escaping issues
- **Standard operations**: `info`, `accounts`, `sign-message`, `sign-typed-data`, `sign-transaction`, `send-transaction`
- **Semantic exit codes**: Enable orchestration and error handling
- **Orchestrator CLI**: `wallet` binary discovers providers and delegates

### Inspiration

- [EIP-6963](https://eips.ethereum.org/EIPS/eip-6963) — Browser wallet discovery
- [git credential helpers](https://git-scm.com/docs/gitcredentials) — CLI plugin discovery pattern

### Reference Implementation

- Orchestrator: [`@anthropic-ai/wallet-cli`](https://github.com/anthropics/agent-sdk)
- WalletConnect adapter: `wallet-walletconnect` in [`@walletconnect/cli-sdk`](https://github.com/anthropics/agent-sdk)
- Consumer: [`nansen-cli`](https://github.com/nansen-ai/nansen-cli) x402 auto-payment

The full draft spec will be submitted as a PR.

abrahemmrw-a11y:
gbjuTLKKmnbQSfkZ7H6Ipewp7J2VqH7r6/hBY0wG4Fd6dOJFtuQ2/Q

bumblefudge:
Initial draft merged but I am leaving the issue open and assigned to you for the [minor suggestions](https://github.com/ChainAgnostic/CAIPs/pull/397#issuecomment-3986750537) i made in that initial PR!

PR #397, opened 2026-02-23
Add CAIP: CLI Wallet Protocol (CWP)

arein:
## Summary

This CAIP proposes a standard protocol for CLI applications to discover and interact with wallet providers through executable plugins on PATH.

**Key concepts:**

- **Binary naming**: `wallet-<name>` executables (e.g., `wallet-walletconnect`, `wallet-ledger`, `wallet-cast`)
- **Communication**: JSON on stdin/stdout, status on stderr, semantic exit codes
- **Operations**: `info`, `accounts`, `sign-message`, `sign-typed-data`, `sign-transaction`, `send-transaction`
- **Discovery**: PATH scanning + optional `~/.config/wallet/config.json` for preferences
- **Orchestrator**: `wallet` CLI discovers providers and delegates

**Inspired by:**
- [EIP-6963](https://eips.ethereum.org/EIPS/eip-6963) — Browser wallet discovery
- [git credential helpers](https://git-scm.com/docs/gitcredentials) — CLI plugin pattern

## Motivation

CLI blockchain tools (API clients, deployment scripts, AI agents) hardcode specific wallet providers. There's no CLI equivalent of EIP-6963 for pluggable wallet discovery. This creates tight coupling and limits user choice.

## Reference Implementations

- **Orchestrator**: `@anthropic-ai/wallet-cli` — discovers `wallet-*` on PATH, delegates operations
- **Provider**: `wallet-walletconnect` — CWP adapter for the `walletconnect` CLI
- **Consumer**: `nansen-cli` — uses `wallet` CLI for x402 auto-payment signing

## Discussion

https://github.com/ChainAgnostic/CAIPs/issues/396

### DIFF
--- a/CAIPs/caip-draft_cli_wallet_protocol.md
+++ b/CAIPs/caip-draft_cli_wallet_protocol.md
@@ -1,0 +1,395 @@
+---
+caip: CAIP-X
+title: CLI Wallet Protocol
+author: Derek Rein (@arein)
+discussions-to: https://github.com/ChainAgnostic/CAIPs/issues/396
+status: Draft
+type: Standard
+created: 2026-02-23
+updated: 2026-02-23
+requires: [2, 10]
+---
+
+## Simple Summary
+
+A standard protocol for CLI applications to discover and interact with wallet providers through executable plugins on PATH, inspired by git credential helpers and [EIP-6963][] browser wallet discovery.
+
+## Abstract
+
+CLI Wallet Protocol (CWP) defines a convention for wallet providers to expose signing and account capabilities to command-line tools. Wallet providers ship executables named `wallet-<name>` that implement a small set of JSON-based operations (info, accounts, sign-message, sign-typed-data, sign-transaction, send-transaction). A central `wallet` orchestrator discovers providers on PATH and delegates operations to them. This decouples CLI tools that need wallet functionality from specific wallet implementations, enabling hardware wallets, browser extensions, cloud signers, and local keystores to participate equally.
+
+## Motivation
+
+CLI-based blockchain tools (API clients, deployment scripts, AI agents) increasingly need wallet interaction — signing transactions, authorizing payments, proving identity. Today, each tool hardcodes support for a specific wallet provider (e.g., WalletConnect, Ledger, local keystore), creating tight coupling that limits user choice and increases integration burden.
+
+Browser-based ecosystems solved this with [EIP-6963][] (Multi Injected Provider Discovery), allowing dApps to discover all available wallets without hardcoding. No equivalent exists for the CLI environment.
+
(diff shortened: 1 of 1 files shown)


CAIPs/caip-draft_cli_wallet_protocol.md
```diff
@@ -0,0 +1,628 @@
...
+**Output**:
+```json
+{
+  "transactionHash": "0x..."
+}
+```
+
+#### `grant-session`
+
+Creates a scoped permission session for autonomous operation. MUST complete within 120 seconds (requires human approval to authorize the session).
```
bumblefudge, line 244: is human approval assumed to be interactive? Would it be more precise/explicit to say "MUST timeout if interactive user approval has not authorized the session within 120 seconds"?

CAIPs/caip-draft_cli_wallet_protocol.md
```diff
@@ -0,0 +1,628 @@
...
+- **Status/Progress**: stderr only (MUST NOT write non-JSON to stdout)
+- **Exit codes**: Semantic (see [Exit Codes](#exit-codes))
+
+Providers MUST be stateless between invocations. Session state (if needed) MUST be persisted to the filesystem.
+
+### Operations
+
+#### `info`
+
+Returns provider metadata and capabilities. MUST complete within 3 seconds.
```
bumblefudge, line 66:
```suggestion
Returns provider metadata and capabilities, and doubles as [discovery mechanism](#Discovery). MUST complete within 3 seconds.
```

bumblefudge:
Hey this great and well-thought out, appreciate all the Considerations.  Minor things:

1. Is there already a reference implementation, or known implementations in the wild? Linking to them in the spec would be great if so; it looks like the "Reference Implementations" section of the PR summary is more of a "prior art" or "potential users" list.
2. Attenuation-only makes perfect sense but I wonder if being more precise or normref-ing to something more precise because people don't always interpret that constraint predictably[^1]. I also notice that EIP-7715 doesn't actually include this constraint (it uses the word `attenuate` to mean either escalating or de-escalating permissions!), so anchoring this constraint in another specification or policy semantic seems even more important to do to prevent weird divergences in policy creeping in.
3. Is a wallet-cli conformant if it doesn't do one of the operations? Should the supported functions be returned by the discovery function, if so?

[^1]: For instance, if I grant a total of up to 3 payments, totalling 10Eth or less, be made within the next 3 days, everyone would agree that lowering the number of payments to 1 is an attenuation of scope/privilege, but I suspect changing only the maximum total would not universally be seen as an attenuation, just an update to a value.  (OK bad example because in payments that value is pretty obviously correlated to privilege, but you get what I mean, semantics slips in sometimes).

bumblefudge (approved):

pedrouid (approved):
👍

PR #399, opened 2026-03-25
docs: update broken license link to correct URL

Ocheretovich:


bumblefudge:
Actually, that template is meant to be moved into the /CAIPs/ repo, so the `../` is needed!  Thanks, tho.

PR #402, opened 2026-04-18
Fix CAIP-122 Chain IDs in examples

luca992:
The current Chain IDs in the examples do not align with the described spec to use caip2 chain ids

https://namespaces.chainagnostic.org/solana/caip2
https://standards.chainagnostic.org/CAIPs/caip-2

### DIFF
--- a/CAIPs/caip-122.md
+++ b/CAIPs/caip-122.md
@@ -110,7 +110,7 @@ URI: https://service.org/login
 Version: 1
 Nonce: 32891756
 Issued At: 2021-09-30T16:25:24Z
-Chain ID: 1
+Chain ID: eip155:1
 Resources:
 - ipfs://bafybeiemxf5abjwjbikoz4mc3a3dla6ual3jsgpdr4cjr3oz3evfyavhwq/
 - https://example.com/my-web2-claim.json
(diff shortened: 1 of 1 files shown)


obstropolos:
@haardikk21 @ukstv

haardikk21:
> @haardikk21 @ukstv 

What a blast from the past. LGTM!

haardikk21 (approved):

bumblefudge:
See most recent PR on that spec:
https://github.com/ChainAgnostic/CAIPs/pull/272/files

I really wish someone would take over primary editorship of this CAIP, at least once every 6 months we get another PR trying to fix its many internal contradictions :/

luca992:
@bumblefudge if you are closing this, what is actually correct? ... I actually need to implement this correctly irl 😭

bumblefudge:
Hey sorry to be gruff about this, it's been a never-ending headache and no one's ever really "owned" it. The crucial bit is:

> thereby making EIP-4361 a specific implementation of this specification

As I understand it, the proper usage here is not to use full CAIP-10s and CAIP-2s, but JUST the relevant segment thereof.  As in EIP-4361, you use JUST the address as in `account_address`, and you put JUST the chain-id segment (`1`, not `eip155:1`) in `chain_id`.  In namespaces where native address format or native chainId/network identifier differ, I would recommend using the CAIP version.  A PR making the whole CAIP-122 more consistent and explicit on all of the above would be warmly welcome, I've just been to busy to make it myself (and I'd rather someone relying on CAIP-122 in prod took the author credit and ownership of the edits anyways, I'm just the janitor around here!)

